PR #37, opened 2022-03-14
EIP-1153 testing

moodysalem:
Remaining overhead for accurate tests:
- ~currently we are using delegatecall from a library into a yul proxy to use the opcodes (min `2600 + 100*(N-1)` gas over
head for using transient storage where N = number of loads and stores)~ fixed
- ~had to manually write functions into storage slots which are likely producing suboptimal bytecode, and a jump, so probably another `100 * N`~ fixed
- ~the ethereumjs vm is not correctly clearing transient storage after each transaction, so that results in min N additional writes for each lock (so another `200*N` gas) https://github.com/ethereumjs/ethereumjs-monorepo/pull/1768#issuecomment-1069282799~ fixed

~The overhead compared with Uniswap/v3-core looks like about 20k for mints and 40k for swaps. I'm not sure why swaps are so much more expensive. I think fixing these issues will close the gap significantly, and there is overhead associated with core here that isn't present in v3-core but is present in v3-periphery, so it's not entirely apples to apples (e.g. callback validation gets cheaper with singleton, pool key includes the tokens so longer calldata)~

### DIFF
--- a/contracts/PoolManager.sol
+++ b/contracts/PoolManager.sol
@@ -123,25 +123,6 @@ contract PoolManager is IPoolManager, Owned, NoDelegateCall, ERC1155, IERC1155Re
         }
     }
 
-    /// @dev Push the latest locked by address from the stack
-    /// @return index The index of the locker, i.e. the locker's unique identifier
-    function pushLockedBy(address addr) internal returns (uint256 index) {
-        assembly {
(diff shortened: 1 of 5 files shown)


snreynolds:
Closing :) We have custom binaries merged in

PR #71, opened 2022-05-03
tickSpacing validation in initialize

moodysalem:
fixes https://github.com/Uniswap/core-next/issues/68

### DIFF
--- a/test/PoolManager.spec.ts
+++ b/test/PoolManager.spec.ts
@@ -11,6 +11,7 @@ import {
   PoolManagerReentrancyTest,
   PoolDonateTest,
 } from '../typechain'
+import { MAX_TICK_SPACING } from './shared/constants'
 import { expect } from './shared/expect'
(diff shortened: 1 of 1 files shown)


contracts/PoolManager.sol
```diff
@@ -24,15 +24,13 @@ contract PoolManager is IPoolManager, NoDelegateCall, ERC1155, IERC1155Receiver
     /// @inheritdoc IPoolManager
     int24 public constant override MAX_TICK_SPACING = type(int16).max;
 
+    /// @inheritdoc IPoolManager
+    int24 public constant override MIN_TICK_SPACING = 1;
```
wjmelements, line 28: I don't think this needs to be public. Removing methods like this can save gas on all operations because solidity often structures large switches as sequential if-statements.

wjmelements: https://github.com/Uniswap/v4-core/pull/810

PR #181, opened 2023-05-30
lock by address

### LINKED ISSUE #144: allow hooks to flash-across-before/after calls
this would enable e.g. hook to mint/burn to sandwich a swap without providing the full amount of capital required for the liquidity in the beforeSwap call

this is useful for the yield-boosted v2 hook, where the tokens are deposited in a lending protocol when not being used for swaps. currently, there are a few implementations of this:
- one is withdrawing/depositing the full amount of capital from the lending protocol before/after swap, and depositing it as full range liquidity. 
- another is to look at the size of the trade and mint liquidity only within the tick range touched by the swap, but that requires computing the price after the swap which means the hook has to either ban other liquidity being in the same pool or do swap calc by crossing all the ticks
- you can also force the user through your router in the hook which obtains the lock for the hook before executing the swap, but that makes the pools using the hook less usable

withdrawing all the capital from the lending protocol to sandwich a swap is a big problem because some of the capital may be currently borrowed, preventing the withdrawal and the swap even though only a fraction of the liquidity is needed to support the swap (for example, none of the input tokens need to be withdrawn from the lending protocol)

two possible ways to enable this:
- allow the hook to make swap/modifyPosition calls on behalf of the locker that is calling swap/modifyPosition/etc.. not sure about the security issues with this, but at the surface it seems reasonable
- allow the hook to specify via its address that it needs a lock before/after swap; _maybe_ also do locks by address instead of allowing one address to hold multiple locks, although that might cause an issue with reentrance or when swapping against multiple pools that use the same hook

NoahZinsmeister:
closes #144


contracts/PoolManager.sol
```diff
@@ -132,55 +132,69 @@ contract PoolManager is IPoolManager, Owned, NoDelegateCall, ERC1155, IERC1155Re
     /// @inheritdoc IPoolManager
     mapping(Currency currency => uint256) public override reservesOf;
 
+    IPoolManager.Lock[] private locks;
+
     /// @inheritdoc IPoolManager
-    address[] public override lockedBy;
+    function locksGetter(uint256 i) external view override returns (IPoolManager.Lock memory) {
```
ewilz, line 138: so much for`extsload` 😉

test/foundry-tests/PoolManagerReentrancyTest.sol
```diff
@@ -0,0 +1,167 @@
...
+
+    function testTokenRevert() public {
+        TokenLocker locker = new TokenLocker();
+        MockERC20(Currency.unwrap(currency0)).mint(address(locker), 1);
+        MockERC20(Currency.unwrap(currency0)).approve(address(locker), 1);
+        vm.expectRevert(abi.encodeWithSelector(IPoolManager.CurrencyNotSettled.selector, address(locker)));
+        locker.main(manager, currency0, false);
+    }
+
+    function checker(uint256 depth) external {
```
ewilz, line 149: I know your tests are cute with short little names, but I find it takes some mental fortitude to wrap your head around them. Maybe something more descriptive like `assertionChecker` would help with readability

contracts/PoolManager.sol
```diff
@@ -132,55 +132,69 @@ contract PoolManager is IPoolManager, Owned, NoDelegateCall, ERC1155, IERC1155Re
...
+            // only enforce that deltas are 0 when the outermost lock frame is expiring
+            if (lockIndex == 0) {
+                uint256 length = locksLength();
+                for (uint256 i; i < length; i++) {
+                    address locker = locks[i].locker;
+                    if (lockStates[locker].nonzeroDeltaCount != 0) revert CurrencyNotSettled(locker);
+                }
+                delete locks;
+            } else {
+                lockIndex = locks[lockIndex].parentLockIndex;
```
ewilz, line 189: what's the point of `parentLockIndex`?? is it ever not the previous Lock in the array?  🤔

.forge-snapshots/gas overhead of no-op lock.snap
```diff
@@ -1 +1 @@
-61078
\ No newline at end of file
+63747
```
ewilz, line 1: that's +2669 gas for every lock?

.forge-snapshots/donate gas with 2 tokens.snap
```diff
@@ -1 +1 @@
-185897
\ No newline at end of file
+193417
```
ewilz, line 1: duuude, sup with the donate gas diffs of 7500

contracts/PoolManager.sol
hensha256: Why expose a getter and a length getter, and not just make it public?

contracts/PoolManager.sol
```diff
@@ -129,69 +144,54 @@ contract PoolManager is IPoolManager, Owned, NoDelegateCall, ERC1155, IERC1155Re
...
+                nonzeroDeltaCount--;
             } else if (current == 0) {
-                lockState.nonzeroDeltaCount++;
+                nonzeroDeltaCount++;
             }
         }
 
-        lockState.currencyDelta[currency] = next;
+        currencyDeltas[locker][currency] = next;
     }
```
0xb10ckdev, line 196: `lockIndex == 0` means `locks.length` is 1.
Then what's the purpose of below `for` loop since `length` is 1?

contracts/PoolManager.sol
```diff
@@ -201,7 +201,7 @@ contract PoolManager is IPoolManager, Owned, NoDelegateCall, ERC1155, IERC1155Re
...
 
     modifier onlyByLocker() {
-        address locker = lockedBy[lockedBy.length - 1];
+        address locker = locks[lockIndex].locker;
```
0xb10ckdev, line 204: nit: `locks[lockIndex].locker` can be used directly in the comparison. `locker` will be used twice(to save gas) only when it reverts. So actually, it will cost more gas for succeed cases.

contracts/PoolManager.sol
NoahZinsmeister: 😛 yeah until storage is more or less frozen it's not really worth the trouble imo....

and yeah @hensha256 good point i made it public and removed the indexed getter, though you still need a manual getter for the length

test/foundry-tests/PoolManagerReentrancyTest.sol
NoahZinsmeister: changed the name :)

to be fair, writing reentrancy tests is pretty hard/wonky!

contracts/PoolManager.sol
NoahZinsmeister: yeah so this was the problem with the previous version of this - the parentLockIndex is actually (non-intuitively) not always the previous lock. see `assertionChecker1` for an example (the parentLockIndex is 0, then 2 for the subsequent lock; this is because there was an opened + closed lock before it which bumps the index)

contracts/PoolManager.sol
NoahZinsmeister: nice catch! fixed

NoahZinsmeister:
i have a pretty cursed idea to make this even cheaper, ideally ~as cheap as `main` right now

`locks` is storing length in a full storage slot, which is unnecessary. instead i think we could pack all of `locks.length`, `lockIndex`, and `lockIndex` in a single slot, and then use adjoining slots for the actual array elements

.forge-snapshots/gas overhead of no-op lock.snap
NoahZinsmeister: resolving as gas is now way better

contracts/libraries/LockDataLibrary.sol
```diff
@@ -0,0 +1,44 @@
...
+            }
+            self.length = length + 1;
+            self.index = length;
+        }
+    }
+
+    function getLock(uint256 i) internal view returns (address locker, uint96 parentLockIndex) {
+        unchecked {
+            uint256 position = OFFSET + i;
+            /// @solidity memory-safe-assembly
```
marktoda, line 28: cant you do
```
assembly ('memory-safe') {
}
```

https://docs.soliditylang.org/en/v0.8.20/assembly.html#memory-safety

contracts/libraries/LockDataLibrary.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.19;
+
+import {IPoolManager} from "../interfaces/IPoolManager.sol";
+
+library LockDataLibrary {
```
marktoda, line 6: imo think this could use a comment explaining the custom storage layout handling

marktoda: semi scary as you get into potential collision territory ie if using this lib on two diff instances of LockData or something dumb haha

contracts/PoolManager.sol
```diff
@@ -146,53 +146,38 @@ contract PoolManager is IPoolManager, Owned, NoDelegateCall, ERC1155, IERC1155Re
...
 
-        unchecked {
-            LockState storage lockState = lockStates[id];
-            if (lockState.nonzeroDeltaCount != 0) revert CurrencyNotSettled();
+        if (lockData.index == 0) {
+            if (lockData.nonzeroDeltaCount != 0) revert CurrencyNotSettled();
+            delete lockData;
+        } else {
+            (, uint96 parentLockIndex) = lockData.getActiveLock();
+            lockData.index = parentLockIndex;
```
marktoda, line 161: nit maybe `lockData.pop()` would be lil more readable?

contracts/libraries/LockDataLibrary.sol
```diff
@@ -0,0 +1,44 @@
...
+
+        unchecked {
+            uint256 indexToWrite = OFFSET + length;
+            /// @solidity memory-safe-assembly
+            assembly {
+                // note that the current index becomes the parent lock index
+                sstore(indexToWrite, or(shl(96, locker), index))
+            }
+            self.length = length + 1;
+            self.index = length;
```
marktoda, line 21: ok just thinking through this.. if I lock twice with the same address, I get two different entries here, but only one CurrencyDelta shared between both locks. I guess that's OK since we need to unlock twice but only need to ensure zero deltas once?

contracts/interfaces/IPoolManager.sol
```diff
@@ -135,24 +135,29 @@ interface IPoolManager is IERC1155 {
...
+        /// @notice The index of the active locker
+        uint96 index;
+        /// @notice The total number of nonzero deltas over all active + completed lockers
+        uint64 nonzeroDeltaCount;
+    }
 
-    /// @notice Represents the stack of addresses that have locked the pool. Each call to #lock pushes the address onto the stack
-    /// @param index The index of the locker, also known as the id of the locker
-    function lockedBy(uint256 index) external view returns (address);
+    /// @notice Returns the ith element of the locker array
```
marktoda, line 148: nit spec

marktoda (commented):
i like it, some thoughts

contracts/libraries/LockDataLibrary.sol
marktoda: btw how much is the gas advantage over just having 
```
struct LockData {
  ...
  Locker[] lockers;
}
```

contracts/libraries/LockDataLibrary.sol
ewilz: 😆  https://github.com/Uniswap/v4-core/pull/173#discussion_r1221865465

NoahZinsmeister: oh the compiler recognizes the comments format too! this way is just friendly for linting tools that don't have support for the latest compiler version, but honestly it's been long enough that they now might and this could be a non-issue

contracts/libraries/LockDataLibrary.sol
NoahZinsmeister: yeah i don't love that this lib can be used on two different LockDatas either but i couldn't really think of a clean way to limit this behavior. one idea i had is to write a "custom storage offsets" lib to consolidate all of the offsets and then use values from this lib either directly in consumer libs or pass them through directly as an argument but this also felt needlessly verbose. i think i'm in favor of just being careful 😬

as far as gas, we save a fair amount, as we're able to pack the active locker and the non-0 delta count with the length of the array, so there are a lot of saved sstores/sloads. happy to do comps but it's definitely going to be a fair amount

contracts/libraries/LockDataLibrary.sol
NoahZinsmeister: yes that's correct - we need unique locks to maintain the ability to have nested locks etc - the fact that this allows the same locker to get 2 locks in a row is weird, but ultimately as long as we're a) tracking deltas globally by locker + currency and b) only update (global) non-0 delta count when going from 0->non-0 and vice versa, it should be fine, provided that we check that non-0 delta count is 0 after all locks have expired. and the benefit of this is also that you can punt on deltas until your last lock if you know you'll be doing 2 (the primary motivation behind the pr in the first place). so i think this checks out?

NoahZinsmeister:
i was thinking for a while about whether there could ever be a difference in the order of the queue traversal and the order that solidity call frames expire.

right now we're relying on them being the same, because we a) push an element onto the queue, b) call `lockAcquired`, and then c) either pop from queue (and let call frame expire, going back to c)), or if the current index is 0, meaning there are no more locks, enforce that the non-0 delta count is 0.

so i'm pretty sure this always works, but would love more eyes :)

ewilz:
So I like this, and the gas savings, and I want to get this merged lol instead of leave it hanging. Even though this change allows third party lockers to leave hanging deltas past their lock, I don't see how this decreases security (with or without the change, the tx will revert with 3rd party's hanging deltas, just at different points in the operation). But we discussed en persona, I'd really like to see the `parentIndex` stuff get simplified!! That'd be cool to check out.....

contracts/libraries/LockDataLibrary.sol
snreynolds: Core doesn't have two different LockDatas though right? So not an issue persay but just not desireable?

contracts/libraries/LockDataLibrary.sol
```diff
@@ -0,0 +1,50 @@
...
+            assembly {
+                // in the next storage slot, write the locker
+                sstore(indexToWrite, locker)
+            }
+            // update the sentinel storage slot
+            self.length = length + 1;
+        }
+    }
+
+    /// @dev Pops a locker off the end of the queue. Note that no storage gets cleared.
```
snreynolds, line 30: Why don't we clear storage?

NoahZinsmeister: oh yeah i meant to write more of an explanation here - as it stands we're not clearing it to not use the caller's storage refund up - in the final impl this is going to be transient so it doesn't matter

snreynolds: cool and tstore 🔜

ewilz (approved):
🚀

contracts/interfaces/callback/ILockCallback.sol
```diff
@@ -3,8 +3,7 @@ pragma solidity ^0.8.20;
 
 interface ILockCallback {
     /// @notice Called by the pool manager on `msg.sender` when a lock is acquired
-    /// @param id The id of the lock that was acquired
     /// @param data The data that was passed to the call to lock
     /// @return Any data that you want to be returned from the lock call
-    function lockAcquired(uint256 id, bytes calldata data) external returns (bytes memory);
```
snreynolds, line 9: oo nice that we can remove this 😄

contracts/libraries/LockDataLibrary.sol
snreynolds: I guess one odd behavior of not clearing though is that a getLock() may return you a stale locker if you input an index that has not been used in the current txn

snreynolds: Don't see any issues with that per say, and won't be an issue w transient storage but idk worth noting?

snreynolds (approved):
Only note.. we are just not testing extsload until contract storage settles then?

PR #242, opened 2023-06-14
added setup for running tests

mugrebot:
## Related Issue

You cannot run tests right out of the box since the files (node modules, typings) necessary aren't installed/compiled. 

## Description of changes
Updated readme to reflect steps necessary for both yarn test and forge test to work and run all the tests in the v4-core repo

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Additionally, a pool may be initialized with a hook contract, that can implement
 
 Hooks may also elect to specify fees on swaps, or liquidity withdrawal. Much like the actions above, fees are implemented using callback functions.
 
-The fee values, or callback logic, may be updated by the hooks dependent on their implementation. However _which_ callbacks are executed on a pool, including the type of fee or lack of fee, cannot change after  pool initialization.
+The fee values, or callback logic, may be updated by the hooks dependent on their implementation. However _which_ callbacks are executed on a pool, including the type of fee or lack of fee, cannot change after pool initialization.
 
 ## Repository Structure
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,12 +48,12 @@ Note that helper contracts used by tests are held in the `v4-core/contracts/test
...
 ----interfaces/
-    | IPoolManager.sol
-    | ...
+| IPoolManager.sol
+| ...
 ----libraries/
-    | Position.sol
-    | Pool.sol
-    | ...
+| Position.sol
```
hensha256, line 54: They were indented purposefully because they are inside the folder named! Please can you revert the file structure changes?

README.md
```diff
@@ -69,6 +69,20 @@ To utilize the contracts and deploy to a local testnet, you can install the code
 forge install https://github.com/Uniswap/v4-core
 ```
 
+To run Foundry and Yarn tests:
```
hensha256, line 72: Please can you separate these into instructions for foundry and hardhat tests separately?
Hardhat needs `yarn install`,  `yarn compile`, `yarn test`
Forge needs `forge build`, `forge test`

So would be nice to list them separately :)

hensha256: Also - we have some of these instructions already in `CONTRIBUTING.md`. If you would like to update the instructions maybe it makes sense to do so in that file :)

hensha256 (changes requested):

README.md
mugrebot: > Also - we have some of these instructions already in `CONTRIBUTING.md`. If you would like to update the instructions maybe it makes sense to do so in that file :)


just reverted changes from readme, and added the non overlapping instructions in a section called Setup in contributing.md  🖖

hensha256 (approved):

PR #247, opened 2023-06-15
Separate Fee logic

hensha256:
## Description of changes

The PoolManager was quite bloated with logic around handling the different types of fee. This PR tries to move some of that logic into a `Fees` contract. Given the different types of fee, we also have a `FeeLibrary` to handle the fee flag bits.

Not sure this is the best separation yet


.forge-snapshots/poolExtsloadSlot0.snap
```diff
@@ -1 +1 @@
-1129
\ No newline at end of file
+3129
```
hensha256, line 1: This is just because the extsload tests are failing - need to fix them

hensha256:
Thanks for your comments @mugrebot ! The issue was just that the `pools` mapping had moved to a new slot. You can see the storage layout of a contract using `forge inspect PoolManager.sol:PoolManager storageLayout` - and you'll see that `pools` is at slot 10, whereas it was at slot 8 before the PR :) 

Commit [19a8c72](https://github.com/Uniswap/v4-core/pull/247/commits/19a8c7220767ae145774161c565761440afcbf7a) shows the fix!

snreynolds (dismissed):
I like it, just a few nit suggestions!

snreynolds (approved):

snreynolds (dismissed):

snreynolds (approved):

PR #248, opened 2023-06-15
Move step struct creation out of loop

vladyan18:
## Description of changes

The creation of the `StepComputations` structure for the swap is moved outside the main swap loop. As a result, the gas cost of initializing this structure for each iteration of the loop is reduced (plays a role when crossing several ticks).

It might be stylistically more correct to use "immutable" structures, but I don't see any pitfalls in moving this initialization outside the loop.

### DIFF
--- a/contracts/libraries/Pool.sol
+++ b/contracts/libraries/Pool.sol
@@ -411,10 +411,9 @@ library Pool {
             liquidity: cache.liquidityStart
         });
 
+        StepComputations memory step;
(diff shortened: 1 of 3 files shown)


hensha256:
Thanks for making this PR! Please can you merge in main?

vladyan18:
> Thanks for making this PR! Please can you merge in main?

done

hensha256 (approved):

PR #249, opened 2023-06-15
change to using solmate v7

aadams:
## Related Issue

## Description of changes
bump the version of solmate we are using from v6 to v7


contracts/libraries/CurrencyLibrary.sol
```diff
@@ -32,25 +32,27 @@ library CurrencyLibrary {
...
-                        // Counterintuitively, this call must be positioned second to the or() call in the
-                        // surrounding and() call or else returndatasize() will be zero during the computation.
-                        call(gas(), currency, 0, freeMemoryPointer, 68, 0, 32)
+                        // We use 68 because that's the total length of our calldata (4 + 32 * 2)
+                        // Counterintuitively, this call() must be positioned after the or() in the
+                        // surrounding and() because and() evaluates its arguments from right to left.
+                        call(gas(), currency, 0, 0, 68, 0, 32)
                     )
+
+                mstore(0x60, 0) // Restore the zero slot to zero.
```
shuhuiluo, line 54: I don't think memory slot `0x60` is tampered with so it isn't necessary to clean. Besides IMHO Solmate is lagging behind Solady in terms of gas efficiency. Solady's [implementation](https://github.com/Vectorized/solady/blob/37b4fd04dd82977eb0161839389187abb7afdb58/src/utils/SafeTransferLib.sol#L247-L252) removes the unnecessary `mstore(0x60, 0)` and `gt(returndatasize(), 31)`, which hasn't been merged to Solmate since a year ago. If I were to insert the code instead of importing a library to use directly, I would opt for the better one.

ianjdarrow (approved):

hensha256 (approved):

contracts/libraries/CurrencyLibrary.sol
hensha256: thanks for your comment @shuhuiluo! For now we're keeping in these lines even if they arent needed, as this code is an exact copy of a version of solmate. However, if you think that we should be changing this library to use solady instead, please feel free to put in a new PR that does so :)

PR #253, opened 2023-06-15
feat: optimize mulDiv

vladyan18:
## Related Issue

The mulDiv function can be made slightly cheaper in terms of gas

## Description of changes

The changes should remove a couple of stack operations and one require


contracts/libraries/FullMath.sol
```diff
@@ -18,27 +18,25 @@ library FullMath {
             // then use the Chinese Remainder Theorem to reconstruct
             // the 512 bit result. The result is stored in two 256
             // variables such that product = prod1 * 2**256 + prod0
-            uint256 prod0; // Least significant 256 bits of the product
+            uint256 prod0 = a * b; // Least significant 256 bits of the product
```
hensha256, line 21: For what reason is the `mul` taken out of the assembly block below? Is it more efficient to have it here..?

vladyan18: > For what reason is the `mul` taken out of the assembly block below? Is it more efficient to have it here..?

As I understand, code before change (old) should be compiled to smth like this (0.8.17, 800 runs):

```
PUSH 0			uint256 result
DUP1 			uint256 prod0
DUP1 			uint256 prod1
PUSH 0			
NOT 			
DUP6 			b
DUP8 			a
MULMOD 			mulmod(a, b, not(0))
DUP6 			b
DUP8 			a
MUL 			mul(a, b)
SWAP3 			prod0 := mul(a, b)
POP 		
...	
```

and after this change (new):

```
PUSH 0			uint256 result
DUP4 			a
DUP4 			b
MUL 			a * b
DUP2 			uint256 prod1
PUSH 0			
NOT 			
DUP6 			b
DUP8 			a
MULMOD 			mulmod(a, b, not(0))
...
```

As a result, it turns out several operations less. You can check and compare snapshots before and after changes.
Perhaps this should be optimized more efficiently by the compiler (and may be in the future), but now there is an effect.

vladyan18: tldr: In the old implementation, the compiler honestly initializes the variable to zero, and then rotates the stack to change it

hensha256 (dismissed):
lgtm. Please merge in main and update snapshots

vladyan18:
Done

NoahZinsmeister (approved):

PR #228, opened 2023-06-13
refactor(Hooks): Update hook description and add test for new hook combination

penandlim:
## Related Issue
Fixes #227 

## Description of changes
Fix incorrect NatSpec comments regarding example hook address in `Hooks.sol` library contract file.
The description claims

> /// For example, a hooks contract deployed to address: 0x9000000000000000000000000000000000000000
> /// has leading bits '1001' which would cause the 'before initialize' and 'after swap' hooks to be used.

However this seems to be a typo as this would require shifting 1 by 159 and 156, which corresponds to `BEFORE_INITIALIZE_FLAG` and `AFTER_MODIFY_POSITION_FLAG`.

Confirmed the change is correct by running the new test with `forge test --mt testValidateHookAddressBeforeInitializeAfterModify -vvvvv` with console.log() added to view the hook address.

```
  [4626] HooksTest::testValidateHookAddressBeforeInitializeAfterModify(9034) 
    ├─ [0] console::log(0x900000000000000000000000000000000000234a) [staticcall]
    │   └─ ← ()
    └─ ← ()
```

console.log was then removed before adding to the PR for cleanliness.

PR #254, opened 2023-06-16
fix: comment

devtooligan:
## Related Issue

In this comment but one of the hooks mentioned with that prefix was incorrect.

## Description of changes

Correct comment


0xb10ckdev:
It's already fixed in this PR https://github.com/Uniswap/v4-core/pull/228

PR #255, opened 2023-06-16
hoist loop invariant code, strength reduction (mul -> add)

0xalpharush:
## Related Issue
NA
## Description of changes
- The calculation value + 32 is loop invariant i.e. it will be the same on each iteration of the loop so it can be safely moved from the body to the header. 
- Use add instead of mul since it's cheaper.

I figure if Yul is used why not squeeze out the obvious improvements but lmk if this is neglible.


contracts/PoolManager.sol
```diff
@@ -500,8 +500,10 @@ contract PoolManager is IPoolManager, Owned, NoDelegateCall, ERC1155, IERC1155Re
 
         /// @solidity memory-safe-assembly
         assembly {
+            let valuePtr := add(value, 32)
             for { let i := 0 } lt(i, nSlots) { i := add(i, 1) } {
-                mstore(add(value, mul(add(i, 1), 32)), sload(add(startSlot, i)))
+                mstore(valuePtr, sload(add(startSlot, i)))
+                valuePtr := add(valuePtr, 32)
             }
```
shuhuiluo, line 507: Consider this
```Solidity
            let valuePtr := add(value, 32)
            let endSlot := add(startSlot, nSlots)
            for {} 1 {} {
                mstore(valuePtr, sload(startSlot))
                valuePtr := add(valuePtr, 32)
                startSlot := add(startSlot, 1)
                if eq(startSlot, endSlot) { break }
            }
```
after which `poolExtsloadTickInfoStruct.snap` becomes 2704. It saves one `add` per loop iteration and seemingly involves less stack shuffling.

Furthermore, after replacing
```Solidity
bytes memory value = new bytes(32 * nSlots);
```
with
```Solidity
bytes memory value = new bytes(nSlots << 5);
```
the gas comes down to 2625.

0xalpharush: Thanks, I wasn't 100% convinced I couldn't get rid of `i` but didn't figure it out! The first recommendation makes sense, but I imagined the overflow protection is desired for `bytes memory value = new bytes(32 * nSlots)`. I'll see what someone from Uniswap says

shuhuiluo: Someone has to pass `nSlots >= 1 << 251` for it to overflow though, in which case gas would run out in the loop.

hensha256:
hey @0xalpharush ! sorry for the delay here - we stopped using extsload in our tests while v4 was still undergoing larger changes. However I think in the next week or 2 we will start using extsload again so will be looking to merge this!

Please could you:
- merge main
- update `new bytes(32 * nSlots)` to `new bytes(nSlots << 5)` as @shuhuiluo suggested

And we will try to get this merged in the next 2 weeks!

0xalpharush:
@hensha256 I opened a [new PR](https://github.com/Uniswap/v4-core/pull/559) instead of force pushing to this one. It was simpler than merging the upstream history given the project structure has changed.

snreynolds:
Closing this as its being tracked here https://github.com/Uniswap/v4-core/pull/559

PR #257, opened 2023-06-16
Import and modify `BitMath` implementation from `Solady`

shuhuiluo:
## Related Issue

## Description of changes

`BitMath` is optimized heavily by importing and modifying `mostSignificantBit` and `leastSignificantBit` implementations from [Solady::LibBit](https://github.com/Vectorized/solady/blob/37b4fd04dd82977eb0161839389187abb7afdb58/src/utils/LibBit.sol) which uses the De Bruijn lookup.


hensha256 (changes requested):
Thanks for the PR! Please could you insert some comments to explain what each chunk of the assembly does? With assembly its even more important to explain code clearly :)

shuhuiluo:
> Thanks for the PR! Please could you insert some comments to explain what each chunk of the assembly does? With assembly its even more important to explain code clearly :)

Added. Comments were kept at a minimal since the original code wasn't annotated.

snreynolds:
Hi - could you also update this one to main as well? Want to get updated gas estimates! Thanks!

shuhuiluo:
Closing because of #785.

PR #270, opened 2023-06-17
fix: add proper hook for example in comment

gorgos:
## Description of changes

Fix example in comment. Forth hook is called in this example, not after swap.

### DIFF
--- a/contracts/libraries/Hooks.sol
+++ b/contracts/libraries/Hooks.sol
@@ -7,7 +7,7 @@ import {Fees} from "../libraries/Fees.sol";
 /// @notice V4 decides whether to invoke specific hooks by inspecting the leading bits of the address that
 /// the hooks contract is deployed to.
 /// For example, a hooks contract deployed to address: 0x9000000000000000000000000000000000000000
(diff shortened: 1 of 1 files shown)


contracts/libraries/Hooks.sol
```diff
@@ -7,7 +7,7 @@ import {Fees} from "../libraries/Fees.sol";
 /// @notice V4 decides whether to invoke specific hooks by inspecting the leading bits of the address that
 /// the hooks contract is deployed to.
 /// For example, a hooks contract deployed to address: 0x9000000000000000000000000000000000000000
-/// has leading bits '1001' which would cause the 'before initialize' and 'after swap' hooks to be used.
+/// has leading bits '1001' which would cause the 'before initialize' and 'after modify position' hooks to be used.
```
antoncoding, line 10: i would also suggest changing `1001`  to `10010000` to be help clarify there are 8 bits in total that decide which hooks should be invoked

hensha256:
Thanks for your contribution! This was already fixed in PR#228

Issue #197, opened 2023-06-08
Test 1155 burning outside lock


hensha256:
### Components

Lock and Call,  1155 Balances

### Describe the suggested feature and problem it solves.

To burn an 1155 in exchange for pool balance, the onERC1155Received hook burns and accounts the tokens. However this function doesnt have a lock on it. I _think_ the accountDelta function should revert if you try to transfer 1155s into the manager outside of a lock (as the lockedBy.length should be 0). However we should test this to make sure theres nothing weird you can do by calling onERC1155Received outside of a lock.

Secondly what if an _unsafe_ transfer is done, so the onERC1155Received hook isnt called? Does the pool get a balance that then cannot be burnt?

### Describe the desired implementation.

Tests for the above edgecases, and any others we can think of, to check we've covered any unplanned behaviour.

abhithory:
Hi @hensha256 


So as i am able to understand:

- in "onERC1155Received" function there is no lock which means everyone can call this function.
- "_accountDelta" function is called inside this function. so in this function, we should add a revert when "delta == 0" that is currently it is just a return.

So Now:

we should test that:
- nothing happens wrong when we call  "onERC1155Received" directly outside of a lock.
- also check conditions when an unsafe transfer is done. (conditions like:- the balance gets updated etc..)



please tell me...
Am i thinking right?
also if there is anything wrong or missing something then also tell me......

----------------

Also please tell me these things. i am not able to understand this:---

-  what do you mean by unsafe transfer?? when it happen and how? is it when the "onERC1155Received" hook is not get called?

- Which revert should I add when " delta == 0 "?

ZehnSh:
Hi I want to work on this issue
Thanks

zhongeric:
Closed by #379

PR #272, opened 2023-06-17
add burn test #197

bizzyvinci:
## Related Issue
#197 

## Description of changes
Add test to confirm that burn outside of lock would revert. This is because lockedBy.length is 0 and therefore an underflow error in [_accountDelta](https://github.com/Uniswap/v4-core/blob/d8c1ceed2a029cef8a9249f799c2a996bfa27107/contracts/PoolManager.sol#L183)

> Secondly what if an unsafe transfer is done

unsafe transfer is [impossible](https://forum.openzeppelin.com/t/unsafe-transfer-in-erc1155/18758) cause OZ implementation strictly complies with the [ERC1155 spec](https://eips.ethereum.org/EIPS/eip-1155).

### DIFF
--- a/contracts/test/PoolBurnTest.sol
+++ b/contracts/test/PoolBurnTest.sol
@@ -1,0 +1,99 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+import {Currency, CurrencyLibrary} from "../libraries/CurrencyLibrary.sol";
+import {IERC20Minimal} from "../interfaces/external/IERC20Minimal.sol";
+
+import {ILockCallback} from "../interfaces/callback/ILockCallback.sol";
+import {IPoolManager} from "../interfaces/IPoolManager.sol";
+
+contract PoolBurnTest is ILockCallback {
+    using CurrencyLibrary for Currency;
+
+    IPoolManager public immutable manager;
+
+    constructor(IPoolManager _manager) {
+        manager = _manager;
+    }
+
+    // struct CallbackData
+    struct CallbackData {
+        address sender;
+        bool isMint;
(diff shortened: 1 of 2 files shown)


contracts/test/PoolBurnTest.sol
```diff
@@ -0,0 +1,99 @@
...
+
+    function lockAcquired(uint256, bytes calldata rawData) external returns (bytes memory) {
+        require(msg.sender == address(manager));
+
+        CallbackData memory data = abi.decode(rawData, (CallbackData));
+
+        if (data.isMint) {
+            if (data.amount0 > 0) {
+                uint256 currencyBalBefore = IERC20Minimal(Currency.unwrap(data.key.currency0)).balanceOf(data.sender);
+                uint256 managerBalBefore = manager.balanceOf(data.sender, data.key.currency0.toId());
```
emmaguo13, line 44: Might be more clear + consistent (with our other tests) to check the balance of the 1155 in the forge test file like this [line](https://github.com/Uniswap/v4-core/blob/571c7941914b1fb05521e31300a0068a4612358b/test/foundry-tests/PoolManager.t.sol#L895)

emmaguo13 (commented):
Would be a good idea to look into native eth support as well

bizzyvinci:
Eth support has been added and ERC1155 balance check has been moved to PoolManager.t.sol
[Balance check](https://github.com/bizzyvinci/uniswap-v4-core/blob/67d29d40c66cd272ec85805350b219e949702e59/contracts/test/PoolBurnTest.sol#L64) for ETH is different because before and after balance are the same due to `msg.value`

snreynolds:
Closing as burning outside of a lock is no longer an issue/not allowed.

PR #277, opened 2023-06-20
add getPosition

emmaguo13:
Position should be exposed to users

- A bit complex to access Position.Info through extsload
- Using extsload to get multiple words may not be more efficient than just having a getter function that directly accesses the mapping ?


ewilz (commented):
approving for now, but must remove eventually in favor of this getters issue https://github.com/Uniswap/v4-core/issues/278

ewilz (approved):

PR #279, opened 2023-06-21
do `abi.encodePacked` with inline assembly in `Position.get`

shuhuiluo:
## Related Issue

Inefficient `abi.encodePacked` in `Position.get`.

## Description of changes

Hash the position key with efficient encoding in memory, saving 86 gas and some bytecode size.


contracts/libraries/Position.sol
```diff
@@ -31,7 +31,16 @@ library Position {
...
         returns (Position.Info storage position)
     {
-        position = self[keccak256(abi.encodePacked(owner, tickLower, tickUpper))];
+        // positionKey = keccak256(abi.encodePacked(owner, tickLower, tickUpper))
+        bytes32 positionKey;
+        /// @solidity memory-safe-assembly
+        assembly {
+            mstore(0x06, tickUpper) // [0x23, 0x26)
+            mstore(0x03, tickLower) // [0x20, 0x23)
+            mstore(0, owner) // [0x0c, 0x20)
```
ewilz, line 40: is it OK to just start writing to memory address 0 though? this is an internal function and will take on the context of wherever it's called from

shuhuiluo: The scratch space is specifically used for abi encoding and `keccak256`, like computing the storage slot for a key in a mapping, according to the [docs](https://docs.soliditylang.org/en/latest/internals/layout_in_memory.html). Any correctly written EVM code shouldn't have assumptions about values in the scratch space.

snreynolds:
Could you update this PR with main? I'd like to assess the gas differences. From the old gas snapshots it looks like its only saving <100gas so not sure I'm that compelled but I'd like to see it against main. Thanks!

snreynolds:
Hi there, a suggestion from @hensha256 : Could you add a test similar to [this](https://github.com/Uniswap/v4-core/pull/536/files#diff-2ed971e703db7db3ae187d4fc4220c7616958d114c32686891e7aaadec2f9858R14) one that checks that the in memory hash is the same as the abi.encodePacked?

snreynolds (changes requested):
Would love to have a test for this - then feel comfortable merging. Thanks!

hensha256:
the test failures youre seeing are due to a fuzzing issue fixed in another PR. if you merge in main, we should see forge passing

snreynolds (approved):

PR #286, opened 2023-06-25
Increase fee bits precision

### LINKED ISSUE #190: Increase fee bits and pack
Must come after #183 

The `Slot0` struct stores a number of fee-related variables in `uint8`. Namely: `protocolSwapFee`, `protocolWithdrawFee`, `hookSwapFee`, and `hookWithdrawFee`. Each of these uint8s stores 2 separate fee denominators: 4 bits for 0 to 1, and 4 bits for 1 to 0. Using only 4 bits for the denominator does not give us much granularity, and the smallest possible fee is 6.67% (1/15).

If all of these `uint8`s was increased to be a `uint16`, Slot0 would need more storage slots and increase gas drastically. However these is enough space for each fee to be given 6 bits instead of 4.

Proposal: Replace the 4 fee variables, with 2:
- `uint24 withdrawFees`
- `uint24 swapFees`

Where each uint24 is broken down as:
- 6 bits for protocol fee 0->1
- 6 bits for protocol fee 1->0
- 6 bits for hook fee 0->1
- 6 bits for hook fee 1->0

HrikB:
## Related Issue
Increasing fee denominators from 4 bits per direction to 6 bits. 
Closes #190  

## Description of changes
Mostly followed the proposal outlined in the issue, but also opted to follow @snreynolds's suggestion and to separate the two fee variables as `protocolFees` and `hookFees`.


ewilz:
I think you can request to be assigned to the issue, can you plz do that so I can assign it to you?

HrikB:
@ewilz I don't see the ability to request being assigned to an issue. I believe you can just assign me using my tag, no?

ewilz:
> @ewilz I don't see the ability to request being assigned to an issue. I believe you can just assign me using my tag, no?

if you comment on it, I think maybe it becomes possible. maybe try that

contracts/libraries/Pool.sol
```diff
@@ -58,22 +64,21 @@ library Pool {
...
 
-    /// The uint8 fees variables are represented as integer denominators (1/x)
-    /// For swap fees, the upper 4 bits are the fee for trading 1 for 0, and the lower 4 are for 0 for 1 and are taken as a percentage of the lp swap fee.
-    /// For withdraw fees the upper 4 bits are the fee on amount1, and the lower 4 are for amount0 and are taken as a percentage of the principle amount of the underlying position.
-    /// swapFee:     1->0  | 0->1
-    /// withdrawFee: fee1  | fee0
+    /// Each uint24 variable packs both the swap fees and the withdraw fees. The upper 12 bits are the swap fees, and the lower 12 bits are the withdraw fees.
+    /// For swap fees, the upper 6 bits are the fee for trading 1 for 0, and the lower 6 are for 0 for 1 and are taken as a percentage of the lp swap fee.
+    /// For withdraw fees the upper 6 bits are the fee on amount1, and the lower 6 are for amount0 and are taken as a percentage of the principle amount of the underlying position.
+    ///                   withdrawFees    |    swapFees
```
ewilz, line 70:
```suggestion
    /// bits          0  2  4  6  8  10 12 14 16 18 20 22 24
    ///               |  withdrawFees   |    swapFees     |
    ///               ┌────────┬────────┬────────┬────────┐
    /// protocolFees: | fee1   |  fee0  |  1->0  |  0->1  |
    /// hookFees:     | fee1   |  fee0  |  1->0  |  0->1  |
    ///               └────────┴────────┴────────┴────────┘
```

HrikB:
> > @ewilz I don't see the ability to request being assigned to an issue. I believe you can just assign me using my tag, no?
> 
> if you comment on it, I think maybe it becomes possible. maybe try that

Ok made a comment

contracts/libraries/Pool.sol
ewilz: 1-> 0 | 0->1 should be for all swap fees, not protocol fees

HrikB: Ok I see. I think all that needed correction was the comment since this was how the code was already working. However, I think the comment should be updated to
```
    /// bits          0  2  4  6  8  10 12 14 16 18 20 22 24
    ///               |  withdrawFees   |    swapFees     |
    ///               ┌────────┬────────┬────────┬────────┐
    /// protocolFees: | fee1   |  fee0  |  0->1  |  1->0  |
    /// hookFees:     | fee1   |  fee0  |  0->1  |  1->0  |
    ///               └────────┴────────┴────────┴────────┘
```

The update here is from from `1-> 0 | 0->1` to 0-> 1 | 1->0

Pool.sol L418
```
        SwapCache memory cache = SwapCache({
            liquidityStart: self.liquidity,
            protocolFee: params.zeroForOne
                ? (getSwapFee(slot0Start.protocolFees) % 64)
                : (getSwapFee(slot0Start.protocolFees) >> 6),
            hookFee: params.zeroForOne ? (getSwapFee(slot0Start.hookFees) % 64) : (getSwapFee(slot0Start.hookFees) >> 6)
        });
```

From this code, (if `zeroForOne` represents `0 -> 1`), then from the way the code works, it seems the upper bits are used for `1 -> 0`.

ewilz: Yah feel free to make sure the comment matches the code obviously. so the right bits of the swap are `0->1` and the left bits of swap are `1-> 0`. It also looks like swapFees are on the left and withdraw fees are on the right, so that'd be a good change to make too!!!

ewilz:
reallyl sorry will try and get this merged soon so it doesn't happen again, but do you mind dealing with the merge conflicts?

contracts/libraries/Pool.sol
HrikB: So it depends on how 'left' and 'right' are defined. I prefer to use the terms 'upper' and 'lower'. In the way the picture is drawn, upper bits are on the right and upper bits are indeed used for swap fees, so that is already correct.

Pool.sol L136
```
function getSwapFee(uint24 feesStorage) internal pure returns (uint16) {
    return uint16(feesStorage >> 12);
}
```

However, `fee1` and `fee0` do need to be flipped. So this will be the final correct comment:

HrikB:
> reallyl sorry will try and get this merged soon so it doesn't happen again, but do you mind dealing with the merge conflicts?

Yes, will take care of this

contracts/libraries/Pool.sol
ewilz: I see what you're saying. We use "shift right" a lot in the fee logic, so I think the concept of right and left is pretty baked in. What if we switched the bit index's so lsb is on the right `24 <-- 0` ??

HrikB: Yeah so this would also work.

```
/// bits          24 22 20 18 16 14 12 10 8  6  4  2  0
///               |    swapFees     |   withdrawFees  |
///               ┌────────┬────────┬────────┬────────┐
/// protocolFees: | 1->0   |  0->1  |  fee1  |  fee0  |
/// hookFees:     | 1->0   |  0->1  |  fee1  |  fee0  |
///               └────────┴────────┴────────┴────────┘
```

Would you prefer it this way?

ewilz: yes --^ this looks good to me

contracts/libraries/Pool.sol
```diff
@@ -17,6 +17,12 @@ library Pool {
     using Position for mapping(bytes32 => Position.Info);
     using Position for Position.Info;
 
+    uint16 private constant FEE_MAX = 2 ** 12 - 1;
```
ewilz, line 20: total nit (non-blocker): wondering if hex format is more intuitive

```
uint16 private constant FEE_MAX = 0xfff;  // max 12 bits
```

contracts/libraries/Pool.sol
```diff
@@ -120,25 +128,36 @@ library Pool {
...
-    function setHookFees(State storage self, uint8 newHookSwapFee, uint8 newHookWithdrawFee) internal {
+    function setHookFees(State storage self, uint16 newHookSwapFee, uint16 newHookWithdrawFee) internal {
         if (self.slot0.sqrtPriceX96 == 0) revert PoolNotInitialized();
 
-        self.slot0.hookSwapFee = newHookSwapFee;
-        self.slot0.hookWithdrawFee = newHookWithdrawFee;
+        self.slot0.hookFees = _calculateFeesFromSwapAndWithdraw(newHookSwapFee, newHookWithdrawFee);
+    }
+
+    function _calculateFeesFromSwapAndWithdraw(uint16 swapFee, uint16 withdrawFee) internal pure returns (uint24) {
```
ewilz, line 156: `_formatSwapAndWithdrawFee` sounds more descriptive than calculate?

test/foundry-tests/Fees.t.sol
ewilz, line 1: can we add test cases for revert when `setHookFees` and `setProtocolFees` are called with values greater than MAX_FEE? same for initialize, but maybe that goes in Pool.t.sol?

contracts/libraries/Pool.sol
ewilz: oh and technically this is a` FEE_DENOMINATOR_MAX` I do think the denominator aspect is important to clarify.

contracts/libraries/Pool.sol
snreynolds: This version looks good to me too^ !

contracts/libraries/Pool.sol
```diff
@@ -108,10 +116,10 @@ library Pool {
...
         State storage self,
         uint160 sqrtPriceX96,
-        uint8 protocolSwapFee,
-        uint8 hookSwapFee,
-        uint8 protocolWithdrawFee,
-        uint8 hookWithdrawFee
+        uint16 protocolSwapFee,
+        uint16 hookSwapFee,
+        uint16 protocolWithdrawFee,
+        uint16 hookWithdrawFee
```
snreynolds, line 122: I think we can make the interface for this function

`initialize(State storage self, uint160 sqrtPriceX96, uint24 protocolFee, uint24 hookFee)`

contracts/PoolManager.sol
```diff
@@ -380,14 +366,14 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
     }
 
     function setProtocolFees(PoolKey memory key) external {
-        (uint8 newProtocolSwapFee, uint8 newProtocolWithdrawFee) = _fetchProtocolFees(key);
+        (uint16 newProtocolSwapFee, uint16 newProtocolWithdrawFee) = _fetchProtocolFees(key);
```
snreynolds, line 369: _fetchProtocolFees can just return a uint24 `newProtocolFee`

contracts/PoolManager.sol
```diff
@@ -380,14 +366,14 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
...
         PoolId id = key.toId();
         pools[id].setProtocolFees(newProtocolSwapFee, newProtocolWithdrawFee);
         emit ProtocolFeeUpdated(id, newProtocolSwapFee, newProtocolWithdrawFee);
     }
 
     function setHookFees(PoolKey memory key) external {
-        (uint8 newHookSwapFee, uint8 newHookWithdrawFee) = _fetchHookFees(key);
+        (uint16 newHookSwapFee, uint16 newHookWithdrawFee) = _fetchHookFees(key);
```
snreynolds, line 376: same here.. can just return uint24 `newHookFee`

snreynolds: and then you can also change the interface for setHookFees, setProtocolFees, and also initialize

contracts/libraries/Pool.sol
```diff
@@ -120,25 +128,36 @@ library Pool {
...
+    function setHookFees(State storage self, uint16 newHookSwapFee, uint16 newHookWithdrawFee) internal {
         if (self.slot0.sqrtPriceX96 == 0) revert PoolNotInitialized();
 
-        self.slot0.hookSwapFee = newHookSwapFee;
-        self.slot0.hookWithdrawFee = newHookWithdrawFee;
+        self.slot0.hookFees = _calculateFeesFromSwapAndWithdraw(newHookSwapFee, newHookWithdrawFee);
+    }
+
+    function _calculateFeesFromSwapAndWithdraw(uint16 swapFee, uint16 withdrawFee) internal pure returns (uint24) {
+        if (swapFee > FEE_MAX) revert FeeOutOfBounds(swapFee);
```
snreynolds, line 157: After making those other changes, since the fee will not be passed as a uint16, you don't need to check for FeeOutOfBounds because they cannot exceed the 24 bits allocated. Also FEE_MAX is misleading because the actual denominator that a protocol fee is allowed to be is already defined by MIN_PROTOCOL_FEE_DENOMINATOR

snreynolds (changes requested):

contracts/libraries/Pool.sol
HrikB: Don't think this function is needed at all anymore

test/foundry-tests/Fees.t.sol
HrikB: Well originally, the `> 2 ** 12` was not being done in the singleton, it was being done in the `MockHooks` and `ProtocolFeeControllerTest` test contracts. However, I do think it makes sense to move those checks into the singleton so I have done that, and then added the test for it.

contracts/Fees.sol
```diff
@@ -28,43 +29,49 @@ abstract contract Fees is IFees, Owned {
...
             _checkProtocolFee(protocolWithdrawFee);
         }
     }
 
     /// @notice There is no cap on the hook fee, but it is specified as a percentage taken on the amount after the protocol fee is applied, if there is a protocol fee.
-    function _fetchHookFees(PoolKey memory key) internal view returns (uint8 hookSwapFee, uint8 hookWithdrawFee) {
+    function _fetchHookFees(PoolKey memory key) internal view returns (uint24 hookFees) {
         if (key.fee.hasHookSwapFee()) {
-            hookSwapFee = IHookFeeManager(address(key.hooks)).getHookSwapFee(key);
+            uint16 swapFee = IHookFeeManager(address(key.hooks)).getHookSwapFee(key);
```
snreynolds, line 59: What do you think about doing something like this? Allowing the integrator to return a full uint24 and then just making sure each 12 bits are only set if there is a fee set.

```
uint24 hookFee = IHookFeeManager(address(key.hooks)).getHookFee(key);
uint24 swapFeeMask = key.fee.hasHookSwapFee() ?  0xFFF : 0
uint24 withdrawFeeMask = key.fee.hasHookWithdrawFee() ? 0xFFF : 0
uint24 fullFeeMask = swapFeeMask << 12 | withdrawFeeMask
hookFee = hookFee & fullFeeMask
```

snreynolds: And then I think for the future it could be nice to have a library to handle parsing uint24s to uint16 fees and vice versa but that can be separate from this PR

contracts/Fees.sol
```diff
@@ -28,43 +29,49 @@ abstract contract Fees is IFees, Owned {
...
         if (key.fee.hasHookWithdrawFee()) {
-            hookWithdrawFee = IHookFeeManager(address(key.hooks)).getHookWithdrawFee(key);
+            uint16 withdrawFee = IHookFeeManager(address(key.hooks)).getHookWithdrawFee(key);
+            if (withdrawFee > MAX_FEE_DENOMINATOR) revert FeeDenominatorOutOfBounds(withdrawFee);
+            hookFees |= withdrawFee;
         }
     }
 
-    function _checkProtocolFee(uint8 fee) internal pure {
+    function _checkProtocolFee(uint16 fee) internal pure {
```
snreynolds, line 71: Can we leave an @dev comment here saying that only 12 least significant bits are used even though this is a uint16.

contracts/Fees.sol
snreynolds: If you all like this suggestion, would then probably suggest a similar interface for setting protocol fees so just having it return the uint24

snreynolds (changes requested):

contracts/Fees.sol
HrikB: Yeah I'm a fan of this. I feel that having unintuitive UX in the pool is OK given its not meant to be directly interacted with by non-technical folk. 

When writing the parsing library you mentioned, could be nice to use custom Value Types... but that can be discussed separately.

contracts/Fees.sol
```diff
@@ -15,6 +15,7 @@ abstract contract Fees is IFees, Owned {
     using CurrencyLibrary for Currency;
 
     uint8 public constant MIN_PROTOCOL_FEE_DENOMINATOR = 4;
+    uint16 private constant MAX_FEE_DENOMINATOR = 0xFFF;
```
ewilz, line 18: think this is leftover?

ewilz:
need to update snapshots

contracts/Fees.sol
```diff
@@ -28,43 +28,44 @@ abstract contract Fees is IFees, Owned {
...
-        }
-
-        if (key.fee.hasHookWithdrawFee()) {
-            hookWithdrawFee = IHookFeeManager(address(key.hooks)).getHookWithdrawFee(key);
+    function _fetchHookFees(PoolKey memory key) internal view returns (uint24 hookFees) {
+        if (address(key.hooks) != address(0)) {
+            try IHookFeeManager(address(key.hooks)).getHookFees(key) returns (uint24 hookFeesRaw) {
+                uint24 swapFeeMask = key.fee.hasHookSwapFee() ? 0xFFF : 0;
+                uint24 withdrawFeeMask = key.fee.hasHookWithdrawFee() ? 0xFFF : 0;
+                uint24 fullFeeMask = swapFeeMask << 12 | withdrawFeeMask;
```
snreynolds, line 58: Instead of shifting here actually, we can just set `swapFeeMask = 0xFFF000`

contracts/interfaces/IFees.sol
```diff
@@ -4,6 +4,9 @@ pragma solidity ^0.8.19;
 import {Currency} from "../types/Currency.sol";
 
 interface IFees {
+    /// @notice Thrown when fee denominator exceeds FEE_DENOMINATOR_MAX
+    /// @param fee The invalid fee
+    error FeeDenominatorOutOfBounds(uint16 fee);
```
snreynolds, line 9: this is unused

snreynolds (changes requested):

snreynolds (changes requested):

ewilz:
snapshots need to be updated!

HrikB:
> snapshots need to be updated!

Hmm... think they are. `yarn run snapshots` isn't producing a diff for me

ewilz (approved):

Issue #203, opened 2023-06-12
Consider fee accrual as 1155 balance


hensha256:
### Components

Singleton, 1155 Balances

### Describe the suggested feature and problem it solves.

Currently fees accrued to the protocol and to hooks are stored in mappings. However the ERC1155 implementation of the PoolManager should allow us to actually accrue these as 1155 minting instead of having separate mappings.

### Describe the desired implementation.

Remove the 2 mappings `protocolFeesAccrued` and `hookFeesAccrued` and instead mint 1155s, similarly to in `PoolManager.mint`

abhithory:
hi @hensha256 , i want to work on this issue. can you please assigne it to me?

hensha256:
Sure! Feel free to tag me here if you have any questions!

abhithory:
ok. Thank you

abhithory:
Hi @hensha256 ,

From reading this and checking contracts I am able to understand that:
- currently, we are storing accrued fees from the protocol and hooks in mappings (protocolFeesAccrued and hookFeesAccrued)

"accrued fees" - refer to the fees that have accumulated or been earned by the protocol and hooks.

- But in PoolManager Erc115 is already implemented. So we should directly mint erc115 nft for this accrued fee instead of storing them in mappings.

Am I thinking right?? or is there anything else?


Also if it is right then I am thinking to solve it this way:---

1.  Remove mappings (protocolFeesAccrued and hookFeesAccrued)

2. Find all places where fees are added in these mappings

3. then instead of adding fees in mapping directly mint nft with it


also please tell me is it right way to approach?

abhithory:
So now for minting ERC1155, it is required:
address to mint, nft id, amount.

so I can set the amount to 1.

what should i set nft is??  currency key? (key.currency0)

and then what to whom should i mint this nft?? to this contract?? (address(this)) or to someone else??

abhithory:
Hi @hensha256  , can you please tell me these things??

So now for minting ERC1155, it is required:
address to mint, nft id, amount.

so I can set the amount to 1.

what should i set nft is?? currency key? (key.currency0)

and then what to whom should i mint this nft?? to this contract?? (address(this)) or to someone else??

in "collectProtocolFees" function from where should I subtract the amount??

should i create a new mapping for both of the two?? in which we will maintain fees accrued of and then subtract in this function??

and also what should be the key for this mapping?? nft id or same as this??

hensha256:
[...]

PR #289, opened 2023-06-30
Fee accrual as 1155 balance

kevin-fruitful:
## Related Issue
Referencing issue #[203](https://github.com/Uniswap/v4-core/issues/203)

## Description of changes

Removed the 2 mappings `protocolFeesAccrued` and `hookFeesAccrued`.
`swap` and `modifyPosition` now mint ERC1155 currencies to the `protocolFeeController` and a pool's `hook` contract.

`collectProtocolFees` and `collectHookFees` now burn ERC1155 currencies.

Respective tests have been updated. Implemented `onERC1155Received` and `onERC1155BatchReceived` for `ProtocolFeeControllerTest` and `MockHooks` test contracts.

## Additional thoughts

Change `_balances` from private to internal such that the variable can be read directly instead of going through `balanceOf`.


contracts/PoolManager.sol
```diff
@@ -479,8 +481,9 @@ contract PoolManager is IPoolManager, Owned, NoDelegateCall, ERC1155, IERC1155Re
     {
         if (msg.sender != owner && msg.sender != address(protocolFeeController)) revert InvalidCaller();
 
-        amountCollected = (amount == 0) ? protocolFeesAccrued[currency] : amount;
-        protocolFeesAccrued[currency] -= amountCollected;
+        amountCollected = (amount == 0) ? balanceOf(address(protocolFeeController), currency.toId()) : amount;
+
+        _burn(address(protocolFeeController), currency.toId(), amountCollected);
```
hiroshitash, line 486: I'm not an official reviewer so I could be wrong. I just want someone more knowledgeable to take a look.

I wonder if we want to call `_burn` when `msg.sender == address(protocolFeeController)` since calling `currency.transfer` below.

contracts/PoolManager.sol
```diff
@@ -490,10 +493,10 @@ contract PoolManager is IPoolManager, Owned, NoDelegateCall, ERC1155, IERC1155Re
     {
         address hookAddress = msg.sender;
 
-        amountCollected = (amount == 0) ? hookFeesAccrued[hookAddress][currency] : amount;
+        amountCollected = (amount == 0) ? balanceOf(address(hookAddress), currency.toId()) : amount;
         recipient = (recipient == address(0)) ? hookAddress : recipient;
 
-        hookFeesAccrued[hookAddress][currency] -= amountCollected;
+        _burn(hookAddress, currency.toId(), amountCollected);
```
hiroshitash, line 499: Same here.

EphemerisNetwork (approved):

contracts/PoolManager.sol
kevin-fruitful: Thanks for the review!

In this singleton design, all ERC20s are held by the Pool Manager contract. Individual pool balances of ERC20s (aka currencies) are being accounted for within this Pool Manager contract.  

This internal call `burn()` is decreasing the ERC1155 balance (or "internal" balance) of the protocol fee controller's fee token (some ERC20). Then, `currency.transfer` is transferring (`ERC20.transfer()`) the fee token rom the pool manager to the (external) protocol fee controller.

Let me know if that makes sense or if I'm misunderstanding something! Thanks!

hensha256:
Hi! Going to close this PR now for a few reasons

1. we have removed hook fees accruing in the contract and instead they are achieved in other methods (#432)
2. we now use 6909s instead of 1155s (#446)
3. we tried to accrue protocol fees as 6909s (#449) but closed this as it made `swap` costs increase dramatically due to the event emitted when minting a 6909

Thanks again for your submission

PR #299, opened 2023-07-06
Custom swap accounting with NoOps

emmaguo13:
Inspired by NoOp design from https://github.com/Uniswap/v4-core/pull/291. @thogard785's commit is merged with this PR


contracts/PoolManager.sol
```diff
@@ -225,8 +257,23 @@ contract PoolManager is IPoolManager, Owned, NoDelegateCall, ERC1155, IERC1155Re
...
+                    _accountNoOp(key, hookDelta);
+                    if (key.hooks.shouldCallAfterModifyPosition()) {
+                        if (
+                            key.hooks.afterModifyPosition(msg.sender, key, params, delta)
+                                != IHooks.afterModifyPosition.selector
+                        ) {
+                            revert Hooks.InvalidHookResponse();
+                        }
+                    }
+                    return hookDelta;
```
NoahZinsmeister, line 273: i guess because it makes this return a little redundant...

contracts/types/BalanceDelta.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity ^0.8.20;
 
+import "forge-std/console2.sol";
```
NoahZinsmeister, line 4: unused

contracts/types/BalanceDelta.sol
```diff
@@ -14,6 +16,23 @@ function toBalanceDelta(int128 _amount0, int128 _amount1) pure returns (BalanceD
...
+        amt0 := and(0x000000000000000000000000000000000000ffffffffffffffffffffffffffff, input)
+        amt1 := and(0x000000000000000000000000000000000000ffffffffffffffffffffffffffff, shr(112, input))
+    }
+    int128 _amount0 = int128(amt0);
+    int128 _amount1 = int128(amt1);
+
+    assembly {
+        balanceDelta :=
+            or(shl(128, _amount0), and(0x00000000000000000000000000000000ffffffffffffffffffffffffffffffff, _amount1))
+    }
```
NoahZinsmeister, line 33: this should just be `toBalanceDelta`

NoahZinsmeister (changes requested):
i wonder if it's worth allowing pools to use a returned hooksDelta even in non-NO_OP cases...one reason is i think it's awkward if initialize can NO_OP (for example, sqrtPriceX96 remains 0 in this case, which i think means the pool is borked unless it exclusively uses no-ops going forward, which is fine for some pools but maybe not for others, and feels like a bit of a footgun). so, i think we should consider disallowing no-op on initialize, but the delta logic could still be useful for hooks to levy an initial fee or incentivize the creator or something?

i also don't really think we should allow calling `after*` hooks when NO_OP is used

finally i think a version of `_accountNoOp` that applies fees would be valuable to see. for the sake of clarity it might make sense to write that as a stacked PR on top of this one

emmaguo13:
NoOping on initialize will prevent the pool with that hook from being included in our pools mapping, so I think it should be removed.

contracts/test/GracefulReturnTestHooks.sol
```diff
@@ -0,0 +1,238 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+import {Hooks} from "../libraries/Hooks.sol";
+import {IHooks} from "../interfaces/IHooks.sol";
+import {IPoolManager} from "../interfaces/IPoolManager.sol";
+import {BalanceDelta, toBalanceDelta} from "../types/BalanceDelta.sol";
```
0xb10ckdev, line 7: nit: unused `toBalanceDelta` import

contracts/test/NoOpTestHooks.sol
```diff
@@ -0,0 +1,256 @@
...
+import {Hooks} from "../libraries/Hooks.sol";
+import {IHooks} from "../interfaces/IHooks.sol";
+import {IPoolManager} from "../interfaces/IPoolManager.sol";
+import {BalanceDelta, toBalanceDelta} from "../types/BalanceDelta.sol";
+import {IHookFeeManager} from "../interfaces/IHookFeeManager.sol";
+import {PoolId, PoolIdLibrary} from "../libraries/PoolId.sol";
+import {IERC1155Receiver} from "@openzeppelin/contracts/token/ERC1155/IERC1155Receiver.sol";
+import {CurrencyLibrary, Currency} from "../libraries/CurrencyLibrary.sol";
+import {IERC20Minimal} from "../interfaces/external/IERC20Minimal.sol";
+import {ILockCallback} from "../interfaces/callback/ILockCallback.sol";
```
0xb10ckdev, line 13: nit: unused import

contracts/PoolManager.sol
```diff
@@ -108,6 +108,25 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
         return pools[id].positions.get(owner, tickLower, tickUpper);
     }
 
+    function _accountNoOp(PoolKey memory key, BalanceDelta hookDelta) internal {
+        // NOTE: we are in the lock state of the user, not the hook, unless the hook is the user
+        _accountDelta(key.currency0, hookDelta.amount0());
```
ewilz, line 113: nit: why not just use `_accountPoolBalanceDelta` for the whole hookDelta?

contracts/PoolManager.sol
```diff
@@ -108,6 +108,25 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
         return pools[id].positions.get(owner, tickLower, tickUpper);
     }
 
+    function _accountNoOp(PoolKey memory key, BalanceDelta hookDelta) internal {
+        // NOTE: we are in the lock state of the user, not the hook, unless the hook is the user
```
ewilz, line 112: I think it'd be neater to take comments out of the function, they feel a bit inconsistent, and the code feels self-exlanatory, and just write one comment above along the lines of:
```
/// @dev on noOp, rebalance 1155 balances of hook and account deltas to the locker
```

contracts/test/GracefulReturnTestHooks.sol
```diff
@@ -0,0 +1,221 @@
...
+
+    function setSwapFee(PoolKey calldata key, uint8 value) external {
+        swapFees[key.toId()] = value;
+    }
+
+    function setWithdrawFee(PoolKey calldata key, uint8 value) external {
+        withdrawFees[key.toId()] = value;
+    }
+
+    function simulateHead(address caller, uint256 nonce, uint256 stage, bytes32 head) external pure returns (bytes32) {
```
ewilz, line 172: i must say..... i am very confused

test/foundry-tests/PoolManager.t.sol
```diff
@@ -984,6 +986,153 @@ contract PoolManagerTest is Test, Deployers, TokenFixture, GasSnapshot, IERC1155
...
+
+        // Donate
+        donateRouter.donate(key, 0, 0);
+
+        // should have donated 1 eth based on hook donate logic
+        assertEq(manager.balanceOf(hookAddr, CurrencyLibrary.toId(currency0)), 12 ether);
+        assertEq(manager.balanceOf(hookAddr, CurrencyLibrary.toId(currency1)), 9 ether);
+    }
+
+    function testPoolManagerGracefulReturnsOnHook(uint160 sqrtPriceX96) public {
```
ewilz, line 1051: I'd say this description does not explain what the test is testing, but just saying that it's testing a test contract. IMO tests should generally be very targeted (like the majority of the tests in this file) and only in certain special cases a test can be very comprehensive and go through many many things. I'm not completely following this whole GracefulReturns style testing.... it seems a bit confusing and unreadable to me

yarn.lock
```diff
@@ -1625,10 +1625,10 @@
     "@types/sinon-chai" "^3.2.3"
     "@types/web3" "1.0.19"
 
-"@openzeppelin/contracts@4.4.2":
-  version "4.4.2"
-  resolved "https://registry.yarnpkg.com/@openzeppelin/contracts/-/contracts-4.4.2.tgz#4e889c9c66e736f7de189a53f8ba5b8d789425c2"
-  integrity sha512-NyJV7sJgoGYqbtNUWgzzOGW4T6rR19FmX1IJgXGdapGPWsuMelGJn9h03nos0iqfforCbCB0iYIR0MtIuIFLLw==
+"@openzeppelin/contracts@^4.9.2":
```
ewilz, line 1628: por que?

package.json
```diff
@@ -27,7 +27,7 @@
     "node": ">=10"
   },
   "dependencies": {
-    "@openzeppelin/contracts": "4.4.2"
+    "@openzeppelin/contracts": "^4.9.2"
```
ewilz, line 30: why?

contracts/PoolManager.sol
```diff
@@ -108,6 +108,25 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
         return pools[id].positions.get(owner, tickLower, tickUpper);
     }
 
+    function _accountNoOp(PoolKey memory key, BalanceDelta hookDelta) internal {
```
ewilz, line 111: nit: this function would be more well organized below `_accountPoolBalanceDelta` or at least in that general vicinity

contracts/test/GracefulReturnTestHooks.sol
emmaguo13: This is all from #291, I can remove it after I merge my fee changes, the test is not super necessary

contracts/PoolManager.sol
emmaguo13: updated in most current commit, but it increases gas for modifyPosition by 56 and swap by 17

contracts/test/GracefulReturnTestHooks.sol
thogard785: The original idea test concept that prompted my PR was using a sequencer inside the V4 hook, so some of that logic carried forward into the test.  My thinking was that sequencing the storage writes and using that to verify that *all* storage writes were successfully saved in the right order would help future-proof the test for transient storage... but i doubt it'll actually be relevant.

snreynolds:
Closing in favor of this NoOp impl https://github.com/Uniswap/v4-core/pull/324

PR #300, opened 2023-07-06
feat: add lock target

### LINKED ISSUE #298: Consider allowing EOAs to call `lock` directly
### Component

Lock and Call

we could allow EOAs to call `lock` directly on the core, with callers specifying a router address on which `lockAcquired` would be called

the upside of this approach is that the `msg.sender` could be trustlessly threaded through to hooks, for use cases such as https://github.com/jdubpark/Uniswap-Hooks/blob/main/contracts/OldAccountHook.sol

the downside (?) is that anyone could trigger the canonical PoolManager to call `lockAcquired(...)` on any address, with at least partial control over the calldata.

h/t @marktoda for the idea

_No response_

### Describe alternatives.

_No response_

### Additional context.

_No response_

marktoda:
This commit adds a lockTarget parameter to lock where any caller can
specify a locker to handle the routing logic on their behalf. This can
allow safe storage via tstore and passing usage of the lock initiator
to hook callbacks

closes #298

## Related Issue

## Description of changes


marktoda:
[WIP] needs tests

ewilz:
I imagine we won't see the true gas changes this involves until the tests start calling PoolManager directly....I'd honestly expect gas to go _down_ rather than up? I just dont' see a huge benefit tho, bc ultimately a router will need to call swap, and msg.sender will still be a router..... am I missing something?

contracts/PoolManager.sol
```diff
@@ -171,12 +171,12 @@ contract PoolManager is IPoolManager, Owned, NoDelegateCall, ERC1155, IERC1155Re
     }
 
     /// @inheritdoc IPoolManager
-    function lock(bytes calldata data) external override returns (bytes memory result) {
+    function lock(address lockTarget, bytes calldata data) external override returns (bytes memory result) {
```
ewilz, line 174: I imagine we won't see the true gas changes this involves until the tests start calling PoolManager directly....I'd honestly expect gas to go down rather than up? I just dont' see a huge benefit tho, bc ultimately a router will need to call swap, and msg.sender will still be a router..... am I missing something?

marktoda: Yeah gas goes down about 2k since it avoids the extra call!

Basic idea is if a hook needs to know the _swapper_ address that currently requires a trusted router to pass through the swapper data. With this, the swapper could just call the poolmanager directly and hook could use PM's msg.sender to determine the swapper

marktoda:
if we do this, future TODO is to tstore the actual msg.sender and pass it as a param to hook calls

contracts/PoolManager.sol
azflin: >hook could use PM's msg.sender to determine the swapper

need a way to pass PM's msg.sender into hook. I've discussed one way here: https://www.notion.so/uniswaplabs/V4-Arbitrary-Data-msg-sender-73d4a2d3f43241049249af6278abd2d5?pvs=4#0327195ab2ca4827829e9d64c398a38b

contracts/interfaces/IPoolManager.sol
```diff
@@ -162,9 +162,10 @@ interface IPoolManager is IERC1155 {
     function getCurrencyDelta(uint256 id, Currency currency) external view returns (int256);
 
     /// @notice All operations go through this function
+    /// @param lockTarget The address to call the callback on
     /// @param data Any data to pass to the callback, via `ILockCallback(msg.sender).lockCallback(data)`
     /// @return The data returned by the call to `ILockCallback(msg.sender).lockCallback(data)`
-    function lock(bytes calldata data) external returns (bytes memory);
+    function lock(address lockTarget, bytes calldata data) external returns (bytes memory);
```
azflin, line 168: I thought your original idea was to add a new lock function, instead of modifying existing one. Why change of mind?

marktoda: i always planned on just modifying the existing one, not a huge benefit of having a whole separate lock entrypoint that I can see?

azflin: Fair enough. Maybe tiny gas savings in calling `lock` with 1 param instead of 2 is the only benefit

test/foundry-tests/PoolManager.t.sol
```diff
@@ -808,6 +839,34 @@ contract PoolManagerTest is Test, Deployers, TokenFixture, GasSnapshot, IERC1155
...
+            PoolSwapTest.TestSettings({withdrawTokens: true, settleUsingTransfer: true});
+
+        manager.initialize(key, SQRT_RATIO_1_1);
+        swapRouter.swap(key, params, testSettings);
+
+        params = IPoolManager.SwapParams({zeroForOne: true, amountSpecified: 100, sqrtPriceLimitX96: SQRT_RATIO_1_4});
+        testSettings = PoolSwapTest.TestSettings({withdrawTokens: false, settleUsingTransfer: false});
+
+        snapStart("simpleSwapEOAInitiated");
+        manager.lock(
```
emmaguo13, line 864: maybe check that the the swap emitted a swap event properly : )

contracts/PoolManager.sol
ewilz: >With this, the swapper could just call the poolmanager directly and hook could use PM's msg.sender to determine the swapper

my confusion here is: the hook doesn't get called until we call swap (which is done by the router, not the EOA which called `lock`, so now that msg.sender is lost by the tiem we're in swap.)

I really didn't (don't) like the idea of tracking `msg.sender` in core, but now I'm realizing that it's so hard to do it from the router, bc a hook would have to be 4ever paired to a router which doesn't seem ideal. I think the t-storing the lock's `msg.sender`in core is not going to be straightforward, bc just like locker addr, it will have to be recursive, and I'm still wondering whether we _really_ want to support the permissioned usecase enough to merit the complexity...

azflin: > my confusion here is: the hook get's called inside swap, which is called by the router, NOT the EOA which called lock, so by the time we're calling the hook, the original msg.sender is lost.

Yes this is true. I addressed this issue above in https://github.com/Uniswap/v4-core/pull/300#discussion_r1256015791.

> I think the t-storing the lock's msg.sender in core is not going to be straightforward, bc just like locker addr, it will have to be recursive
I don't follow - can you be more specific? I believe the method I've described in https://github.com/Uniswap/v4-core/pull/300#discussion_r1256015791 is sufficient.

NoahZinsmeister (commented):
my overall concern with this feature after reading the PR is that it doesn't actually guarantee that the `lock` caller is a "real user" and not a contract/router/whatever; in fact it locks users into a specific calling format if a hook is expecting this

i'm also worried that this isn't going to scale in an AA world and that hooks are just going to have to resort to other identification methods instead

imo we should add arbitrary calldata as a 1st class citizen and let that be the bailout for hooks that want this behavior, rather than adding this feature

marktoda:
Seemingly pretty serious gas improvements:

simple swap: 203264
simple swap EOA initiated: 174949
simple swap native: 191605
simple swap EOA initiated native: 168498

src/libraries/Lockers.sol
```diff
@@ -58,7 +66,16 @@ library Lockers {
...
     function getLocker(uint256 i) internal view returns (address locker) {
-        uint256 slot = LOCKERS_SLOT + i;
+        // first slot of the ith array item
+        uint256 slot = LOCKERS_SLOT + (i * LOCKER_STRUCT_SIZE);
+        assembly {
+            locker := tload(slot)
+        }
+    }
+
+    function getLockOriginator(uint256 i) internal view returns (address locker) {
```
marktoda, line 76: name suggestions appreciated! I'm currently thinking maybe we should call this `locker` and call the other `lockTarget`

test/PoolManager.t.sol
```diff
@@ -213,6 +214,68 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
+            PoolSwapTest.TestSettings({withdrawTokens: false, settleUsingTransfer: true, currencyAlreadySent: true});
+
+        snapStart("simpleSwapNativeEOAInitiated");
+        manager.lock{value: 100}(
+            address(swapRouter),
+            abi.encode(PoolSwapTest.CallbackData(address(this), testSettings, nativeKey, params, ZERO_BYTES))
+        );
+        snapEnd();
+    }
+
```
marktoda, line 278: planning on adding some more tests here fyi

src/libraries/Lockers.sol
```diff
@@ -8,24 +8,32 @@ library Lockers {
...
 
+    // The number of slots per item in the lockers array
+    uint256 constant LOCKER_STRUCT_SIZE = 2;
+
     // The slot holding the number of nonzero deltas.
     uint256 constant NONZERO_DELTA_COUNT = uint256(keccak256("NonzeroDeltaCount")) - 1;
 
-    function push(address locker) internal {
+    // pushes an address tuple (address locker, address lockOriginator)
+    // to the locker array, so each length of the array represents 2 slots of tstorage
```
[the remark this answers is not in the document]
snreynolds: Maybe we'd consider actually changing it to an address[] array and then a sep mapping. So we are not repeatedly saving 2 of the same addresses in the callback case (when originator == implementer)

src/libraries/Lockers.sol
snreynolds: yeah heh I was confused by the name. so there are two addrs now
- the address that implements the lockCallback (this is the locker) ?
- the address that triggers the lock call (this is the lockCaller) ?

snreynolds: thats the best i could come up w though 😄

test/PoolManager.t.sol
snreynolds: Yeah maybe a revert test with a lockTarget of an invalid address ? (address0, or addr that doesnt implement lockCallback)

snreynolds (changes requested):

src/libraries/Lockers.sol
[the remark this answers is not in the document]
marktoda: wait wdym by a sep mapping? like from `locker` => `lockOriginator`? Don't we want it to be split up by lock though, not by locker? i.e. I can have multiple locks with the same locker but different lockOriginators

src/libraries/Lockers.sol
marktoda: ahh lockCaller is better

src/interfaces/callback/ILockCallback.sol
```diff
@@ -3,7 +3,8 @@ pragma solidity ^0.8.20;
 
 interface ILockCallback {
     /// @notice Called by the pool manager on `msg.sender` when a lock is acquired
+    /// @param lockOriginator The address that originally locked the PoolManager
     /// @param data The data that was passed to the call to lock
     /// @return Any data that you want to be returned from the lock call
-    function lockAcquired(bytes calldata data) external returns (bytes memory);
+    function lockAcquired(address lockOriginator, bytes calldata data) external returns (bytes memory);
```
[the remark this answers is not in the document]
marktoda: yeah updated to lockCaller

src/libraries/Lockers.sol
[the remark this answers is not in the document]
snreynolds: oh yeah thats true can have multiple of the same locker

snreynolds (approved):
lgtm 🚢

src/PoolManager.sol
```diff
@@ -149,11 +149,11 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
     }
 
     /// @inheritdoc IPoolManager
-    function lock(bytes calldata data) external override returns (bytes memory result) {
-        Lockers.push(msg.sender);
+    function lock(address lockTarget, bytes calldata data) external payable override returns (bytes memory result) {
```
hensha256, line 152: did you consider adding a `if (msg.value > 0) settle()` to the lock function?
definitely not needed but it would probably bring down gas further for our ETH input cases (which is a huge % of our volume)

hensha256: it just feels weird that the ETH is sent in this function but is actually used in a different functions

ewilz:
are gas savings the only reason for this PR? bc I imagine we'll inevitably go through the router anyway since we tend to aggregate across versions, so I imagine this will actually be a gas increase overall for our app? UNless we want to special case v4-only swaps and get them for that subset of swaps...

PR #308, opened 2023-07-13
Implement protocol fees for NoOps

emmaguo13:


contracts/PoolManager.sol
```diff
@@ -280,12 +280,55 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
         returns (BalanceDelta delta)
     {
         if (key.hooks.shouldCallBeforeSwap()) {
+            if (key.hooks.isNoOp() && params.amountSpecified > 0) {
```
ewilz, line 283: spoke about getting rid of selector if we do address flag.

emmaguo13:
might move the fee logic to a function in Pool.sol, but for now, keeping in PoolManager (makes manager code a bit clunky unfortunately)

contracts/libraries/Hooks.sol
ewilz, line 1: if we go with noop address bit, isValidHookAddress should probably check that a noop hook has a before hook for swap/modify/donate since otherwise the hook causes wonky pool state

PR #309, opened 2023-07-14
test: replace a line that perform the same logic with more unified li…

usgeeus:
…ne that sets code

## Related Issue
Replaces a line that performs the same logic with more unified and concise line that sets code

## Description of changes
see commit


ewilz:
thanks, but we're currently phasing out hardhat, this file will be gone soon

PR #316, opened 2023-07-19
.env

nagrarohit:
Just added .env.example file to make it a little bit smoother.


ewilz:
we're working to remove hardhat which is the only runner that would currently use the .env file, but still I don't see any references to the majority of example params listed here...so I don't think this is relevant right now

nagrarohit:
ok i understand @ewilz , is there anything else i can do to make some contributions.

Issue #318, opened 2023-07-25
[Bug]: Hook can steal tokens on withdrawal 


damianrusinek:
### Describe the bug

The withdrawal fee fetched from hook is never validated in terms of max value (as are the protocol fees for example). This allows the hook to set value 1 as the hook fee and when fees are collected ([_calculateExternalFees](https://github.com/Uniswap/v4-core/blob/main/contracts/libraries/Pool.sol#L275)) all withdrawn tokens (minus 1 due to rounding) is accounted for the hook.

Things that make the attack harder:
* User can try to _take_ correct amount of tokens and the withdrawal will revert, but I am sure there will be routers that will take what is allowed after withdrawal.

Things that make the attack easier and more impactful:
* Hook owner can front-run withdrawals, update the fees in hook and fetch them (via _setHookFees_). 
* Hook owner can make it unprofitable to withdraw, resulting in DoS on withdrawal.

### Expected Behavior

The withdrawal fees should limited. Using the same check as is for swap fees (value lower than 4 not accepted) still makes it possible to have 25% withdrawal fee which is quite big. Swap fee is limited to 1%.

### To Reproduce

git checkout poc/hook_fee_bug
forge test -vv --mp test/foundry-tests/Fees.t.sol --mt testHookFeeOnWithdrawalShouldNotSteallAllWithdrawnTokens

### Additional context

I have added the PoC in PR: https://github.com/Uniswap/v4-core/pull/319

damianrusinek:
I can make a gist with test if you want, so that you do not need to clone my repo.

snreynolds:
> The withdrawal fee fetched from hook is never validated in terms of max value

This was actually an intentional design for hook fees. I am open to changing that assumption, but there were a few use cases that we thought maybe would warrant a 100% fee... ie hook has a special way of distributing fees to LPs

damianrusinek:
I understand the business need behind the mentioned functionality, but I would like to highlight some concerns which convince me that it is still a vulnerability:
* The rarity of the mentioned case leads me to believe it is an infrequent scenario.
* The current implementation grants significant power to the hook, allowing it to modify fees without any restrictions, including the possibility of front-running and enabling an upgradeable contract.
* While the functionality is labeled as a "withdrawal fee," it appears more akin to a complete takeover of withdrawn tokens rather than a standard fee structure.
[...]

PR #319, opened 2023-07-25
Added POC that shows bug on withdrawal hook fee.

damianrusinek:
## Related Issue
This PR refers to this https://github.com/Uniswap/v4-core/issues/318 issue but does not resolve it yet as it requires some business decision.

## Description of changes
Added test that is failed when the withdrawal fee is greater than 25% as it was assumed for swap fee, but swap fees were also limited to 1% and 25% referred to distribution between protocol and hook. Therefore, the test will need to be changed when the decision how to solve this problem is made.

### DIFF
--- a/test/foundry-tests/Fees.t.sol
+++ b/test/foundry-tests/Fees.t.sol
@@ -543,4 +543,35 @@ contract FeesTest is Test, Deployers, TokenFixture, GasSnapshot {
             fee = value << 4;
         }
     }
+
+    function testHookFeeOnWithdrawalShouldNotSteallAllWithdrawnTokens() public {
+
+        int128 liquidity = 10e18;
+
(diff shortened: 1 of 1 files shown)


test/foundry-tests/Fees.t.sol
```diff
@@ -543,4 +543,35 @@ contract FeesTest is Test, Deployers, TokenFixture, GasSnapshot {
             fee = value << 4;
         }
     }
+
+    function testHookFeeOnWithdrawalShouldNotSteallAllWithdrawnTokens() public {
```
0xb10ckdev, line 547: Typo: `Steall`

hensha256:
Closing - left a comment in #318 about why im closing it.

PR #324, opened 2023-07-28
NoOp implementation with flag

emmaguo13:
A NoOp implementation with a NoOp flag is helpful to be able to make checks on the hook before the pool initializes with the hook. For instance, if we wanted to add checks that hooks that NoOp on beforeSwap are not able to take a protocol fee, having a flag would be helpful (this example is not implemented in this PR, but could be built on top of this PR).


.gitmodules
```diff
@@ -9,4 +9,4 @@
 	url = https://github.com/transmissions11/solmate
 [submodule "lib/openzeppelin-contracts"]
 	path = lib/openzeppelin-contracts
-	url = https://github.com/OpenZeppelin/openzeppelin-contracts
```
ewilz, line 12: I think their url does auto direct to the SnakeCase, so this maybe does seem more correct?

snreynolds (changes requested):

contracts/PoolManager.sol
```diff
@@ -196,8 +196,13 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
         returns (BalanceDelta delta)
     {
         if (key.hooks.shouldCallBeforeModifyPosition()) {
-            if (key.hooks.beforeModifyPosition(msg.sender, key, params) != IHooks.beforeModifyPosition.selector) {
-                revert Hooks.InvalidHookResponse();
+            bytes4 selector = key.hooks.beforeModifyPosition(msg.sender, key, params);
+            if (selector != IHooks.beforeModifyPosition.selector) {
+                if (key.hooks.isNoOp() && selector == Hooks.RETURN_BEFORE_MODIFY) {
```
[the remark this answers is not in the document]
ewilz: maybe calling it `Hooks.NOOP_SELECTOR` would be more readable, (do need a different one for each hook callback?)

but yah, agree the way the code is, it doesn't _have_ to return the noop selector

test/foundry-tests/Hooks.t.sol
```diff
@@ -227,7 +230,7 @@ contract HooksTest is Test, Deployers, GasSnapshot {
     }
 
     function testValidateHookAddressBeforeAndAfterInitialize(uint152 addr) public {
```
snreynolds, line 232: It was a little unlcear why we have to now add the mask `0x007ffffFfffffffffFffffFFfFFFFFFffFFfFFff`to the address.. We don't need it before bc we require inputs to be uint152 which works nicely with / it guarantees 8 cleared upper bits. Maybe instead we now allow it to be uint160 and mask it or shift it right 9 bits.

snreynolds: or add comment?

test/foundry-tests/PoolManager.t.sol
```diff
@@ -984,6 +985,106 @@ contract PoolManagerTest is Test, Deployers, TokenFixture, GasSnapshot, IERC1155
...
+
+        // Donate
+        snapStart("donate with noop");
+        delta = donateRouter.donate(key, 0, 0);
+        snapStart("donate with noop");
+
+        assertEq(delta.amount0(), 0);
+        assertEq(delta.amount1(), 0);
+    }
+
```
snreynolds, line 1040: Would add a test that initializing a pool with the address that just enables NoOp and nothing else reverts successfully

snreynolds (changes requested):

src/PoolManager.sol
```diff
@@ -184,16 +184,20 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
         IPoolManager.ModifyPositionParams memory params,
         bytes calldata hookData
     ) external override noDelegateCall onlyByLocker returns (BalanceDelta delta) {
+        PoolId id = key.toId();
```
snreynolds, line 187: Why move this to the top level? Seems cleaner to have it in` Pool.sol`

test/Hooks.t.sol
```diff
@@ -168,8 +168,8 @@ contract HooksTest is Test, Deployers, GasSnapshot {
     }
 
     // hook validation
-    function testValidateHookAddressNoHooks(uint152 addr) public {
-        uint160 preAddr = uint160(uint256(addr));
+    function testValidateHookAddressNoHooks(uint160 addr) public {
+        uint160 preAddr = addr & uint160(0x007ffffFfffffffffFffffFFfFFFFFFffFFfFFff);
```
snreynolds, line 172: I have a change to this file on this branch that I think will make updating these tests a lot easier when we introduce new flags. See here -(just flagging) since we'd have conflicts. https://github.com/Uniswap/v4-core/blob/access-lock/test/Hooks.t.sol#L32

test/Hooks.t.sol
```diff
@@ -182,7 +182,8 @@ contract HooksTest is Test, Deployers, GasSnapshot {
                 beforeSwap: false,
                 afterSwap: false,
                 beforeDonate: false,
-                afterDonate: false
+                afterDonate: false,
+                noOp: false
```
snreynolds, line 186: Add `assertFalse(Hooks.shouldAllowNoOp(hookAddr));` All the other tests where you add noOp: false needs this statement as well

test/Hooks.t.sol
```diff
@@ -578,15 +628,16 @@ contract HooksTest is Test, Deployers, GasSnapshot {
                 beforeSwap: true,
                 afterSwap: true,
                 beforeDonate: true,
-                afterDonate: true
+                afterDonate: true,
+                noOp: false
```
snreynolds, line 632: this should be set to true?

test/PoolManager.t.sol
```diff
@@ -1193,6 +1210,143 @@ contract PoolManagerTest is Test, Deployers, TokenFixture, GasSnapshot {
...
+        // Swap
+        IPoolManager.SwapParams memory swapParams =
+            IPoolManager.SwapParams({zeroForOne: true, amountSpecified: 100, sqrtPriceLimitX96: SQRT_RATIO_1_2});
+
+        PoolSwapTest.TestSettings memory testSettings =
+            PoolSwapTest.TestSettings({withdrawTokens: true, settleUsingTransfer: true});
+
+        delta = swapRouter.swap(key, swapParams, testSettings, ZERO_BYTES);
+
+        assertEq(delta.amount0(), 0);
```
snreynolds, line 1297: I think delta will be 0 even if NoOp doesn't work here because there is no underlying liquidity in the pool. (It NoOps the add liquidity before). 

So I think we should check that delta is 0 and there are reserves in the pool manager

test/PoolManager.t.sol
```diff
@@ -1193,6 +1210,143 @@ contract PoolManagerTest is Test, Deployers, TokenFixture, GasSnapshot {
...
+
+        // Donate
+        delta = donateRouter.donate(key, 1 ether, 1 ether, ZERO_BYTES);
+
+        assertEq(delta.amount0(), 0);
+        assertEq(delta.amount0(), 0);
+        assertEq(manager.reservesOf(currency0), 0);
+        assertEq(manager.reservesOf(currency1), 0);
+    }
+
```
snreynolds, line 1310: Could also add a test on a pool that has a NoOp flag, beforeInitializeFlag and beforeSwap flag. Then return a NoOp selector in initialize and ensure it fails

snreynolds (changes requested):

src/PoolManager.sol
hensha256: Because if its in the pool, then its _after_ the hook call. So that means on an uninitialized pool, a `swap` transaction succeeds with a hook that NoOps (even though the pool doesnt exist). So I've moved it top level so that we can maintain the invariant that actions on an uninitialized pool always revert

test/Hooks.t.sol
hensha256: oh niceeeeee, great change

test/PoolManager.t.sol
hensha256: > I think delta will be 0 even if NoOp doesn't work here because there is no underlying liquidity in the pool. (It NoOps the add liquidity before).

Thats true - it would actually revert in the `swapRouter` for that because it checks if noop was allowed when delta was 0. So I'll just remove these checks - its handled in the other contract. (And reserves are checked on the next line)

test/PoolManager.t.sol
hensha256: yeah good idea

src/PoolManager.sol
marktoda: maybe new modifier `onlyInitializedPool(key)`? that duplicates id hashing in code but id guess it gets compiled out, could check?

src/PoolManager.sol
```diff
@@ -185,16 +185,20 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
+        if (pools[id].isNotInitialized()) revert PoolNotInitialized();
+
         if (key.hooks.shouldCallBeforeModifyPosition()) {
-            if (
-                key.hooks.beforeModifyPosition(msg.sender, key, params, hookData)
-                    != IHooks.beforeModifyPosition.selector
-            ) {
-                revert Hooks.InvalidHookResponse();
+            bytes4 selector = key.hooks.beforeModifyPosition(msg.sender, key, params, hookData);
```
marktoda, line 192: the length of the per-hook block expands once again https://github.com/Uniswap/v4-core/pull/404/files#r1397713947

I might tinker with some refactors here later

src/PoolManager.sol
```diff
@@ -185,16 +185,20 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
+            if (selector != IHooks.beforeModifyPosition.selector) {
+                if (key.hooks.isValidNoOpCall(selector)) {
+                    return BalanceDelta.wrap(0);
```
marktoda, line 195: Maybe some return indication here to tell the caller that custom accounting was used?
Either a special sentinel value (though that could probs always conflict with a truly possible balancedelta) or a separate boolean?

src/PoolManager.sol
```diff
@@ -185,16 +185,20 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
+
         if (key.hooks.shouldCallBeforeModifyPosition()) {
-            if (
-                key.hooks.beforeModifyPosition(msg.sender, key, params, hookData)
-                    != IHooks.beforeModifyPosition.selector
-            ) {
-                revert Hooks.InvalidHookResponse();
+            bytes4 selector = key.hooks.beforeModifyPosition(msg.sender, key, params, hookData);
+            if (selector != IHooks.beforeModifyPosition.selector) {
+                if (key.hooks.isValidNoOpCall(selector)) {
```
marktoda, line 194: can flatten these ifs with
```
if (IisValidNoop(selector) => return 0
else if (selector != correct selector) => revert

```

marktoda: downside is one extra branch and check in the happy not-noop path but personally think thats worth the flattening

src/libraries/Hooks.sol
```diff
@@ -58,7 +63,12 @@ library Hooks {
     /// @notice Ensures that the hook address includes at least one hook flag or dynamic fees, or is the 0 address
     /// @param hook The hook to verify
     function isValidHookAddress(IHooks hook, uint24 fee) internal pure returns (bool) {
-        // If there is no hook contract set, then fee cannot be dynamic and there cannot be a hook fee on swap or withdrawal.
```
marktoda, line 61: this comment still applies to the bottom part right?

src/test/PoolDonateTest.sol
```diff
@@ -54,6 +57,14 @@ contract PoolDonateTest is PoolTestBase, Test {
 
         assertEq(reserveBefore0, reserveAfter0);
         assertEq(reserveBefore1, reserveAfter1);
+
+        if (BalanceDelta.unwrap(delta) == 0 && (data.amount0 > 0 || data.amount1 > 0)) {
+            // The donate did not work - so check that this hook is allowed to NoOp
+            // Then we can return, as we dont need to settle
+            assert(data.key.hooks.shouldAllowNoOp());
```
marktoda, line 64: assert comments help w/ debugging failing tests

test/PoolManager.t.sol
```diff
@@ -651,6 +651,180 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
+        // Now we let the modify position succeed (so we can test donate)
+        mockHooks.setReturnValue(mockHooks.afterModifyPosition.selector, mockHooks.afterModifyPosition.selector);
+        modifyPositionRouter.modifyPosition(key, LIQ_PARAMS, ZERO_BYTES);
+
+        // Fails at afterDonate hook when it returns a NoOp
+        mockHooks.setReturnValue(mockHooks.afterDonate.selector, Hooks.NO_OP_SELECTOR);
+        vm.expectRevert(Hooks.InvalidHookResponse.selector);
+        donateRouter.donate(key, 100, 100, ZERO_BYTES);
+    }
+
```
marktoda, line 827: No test for reverts if noop flag not set but noop sentinel returned?

src/libraries/Hooks.sol
zhongeric, line 61: nit: keep comment?

test/PoolManager.t.sol
```diff
@@ -651,6 +651,180 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
+        modifyPositionRouter.modifyPosition(key, LIQ_PARAMS, ZERO_BYTES);
+
+        // Now we let the modify position succeed (so we can test donate)
+        mockHooks.setReturnValue(mockHooks.afterModifyPosition.selector, mockHooks.afterModifyPosition.selector);
+        modifyPositionRouter.modifyPosition(key, LIQ_PARAMS, ZERO_BYTES);
+
+        // Fails at afterDonate hook when it returns a NoOp
+        mockHooks.setReturnValue(mockHooks.afterDonate.selector, Hooks.NO_OP_SELECTOR);
+        vm.expectRevert(Hooks.InvalidHookResponse.selector);
+        donateRouter.donate(key, 100, 100, ZERO_BYTES);
```
zhongeric, line 825: missing coverage for forbidden afterSwap?

src/PoolManager.sol
snreynolds: modifier or helper func sgtm :)

src/PoolManager.sol
snreynolds: Makes sense, I think both this and AccessLock introduce some spaghetti paths that we can clean up in a follow up?

src/PoolManager.sol
snreynolds: Hmm interesting.. I like that idea. That would definitely make writing tests (or any integrating contract) a little easier so I'm in favor.

src/PoolManager.sol
```diff
@@ -295,13 +305,21 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
+            if (selector != IHooks.beforeDonate.selector) {
+                if (key.hooks.isValidNoOpCall(selector)) {
+                    return BalanceDelta.wrap(0);
+                } else {
+                    revert Hooks.InvalidHookResponse();
+                }
             }
         }
 
-        delta = _getPool(key).donate(amount0, amount1);
```
snreynolds, line 304: Remove _getPool() then?

src/libraries/Hooks.sol
```diff
@@ -98,4 +108,12 @@ library Hooks {
     function shouldCallAfterDonate(IHooks self) internal pure returns (bool) {
         return uint256(uint160(address(self))) & AFTER_DONATE_FLAG != 0;
     }
+
+    function shouldAllowNoOp(IHooks self) internal pure returns (bool) {
```
snreynolds, line 112: `hasPermissionToNoOp`?

src/test/PoolSwapTest.sol
```diff
@@ -58,12 +61,16 @@ contract PoolSwapTest is Test, PoolTestBase {
         (,, uint256 reserveAfter0, int256 deltaAfter0) = _fetchBalances(data.key.currency0, data.sender);
         (,, uint256 reserveAfter1, int256 deltaAfter1) = _fetchBalances(data.key.currency1, data.sender);
 
-        // Make sure youve added liquidity to the test pool!
-        if (BalanceDelta.unwrap(delta) == 0) revert NoSwapOccurred();
```
snreynolds, line 62: Ah nice - yeah I think also this isn't true for example when you swap on a pool whos liquidity is entirely in one token. Edge case.. but good to remove imo

snreynolds: and now its necessary for no op :P

test/Hooks.t.sol
```diff
@@ -536,15 +601,16 @@ contract HooksTest is Test, Deployers, GasSnapshot {
...
```
snreynolds, line 605: I think this still needs to be set to true?

snreynolds (changes requested):

src/PoolManager.sol
hensha256: yeah I think we should wait to get both this and AccessLock in given they both change it - and then we can do that after :)

src/PoolManager.sol
hensha256: Modifier put gas up by 1000 for swaps because of repeated hashing. I've put in a helper function instead

src/PoolManager.sol
hensha256: Changed it to return the maximum int256... it feels unlikely that a legit trade would ever return that?

snreynolds (commented):
lookin goood! just a few comments.

Lastly, I think it will be good to write some NoOp tests with nested hooks/re-entrancy. Like swap on a hook who calls swap on a different hook that first NoOps and then doesn't... and vice versa

src/types/BalanceDelta.sol
```diff
@@ -22,7 +22,13 @@ function sub(BalanceDelta a, BalanceDelta b) pure returns (BalanceDelta) {
     return toBalanceDelta(a.amount0() - b.amount0(), a.amount1() - b.amount1());
 }
 
+function eq(BalanceDelta a, BalanceDelta b) pure returns (bool) {
+    return a.amount0() == b.amount0() && a.amount1() == b.amount1();
+}
+
 library BalanceDeltaLibrary {
+    BalanceDelta public constant MAXIMUM_DELTA = BalanceDelta.wrap(int256(type(uint256).max));
```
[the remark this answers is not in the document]
hensha256: I just did it because Mark requested we use a sentinel value [here](https://github.com/Uniswap/v4-core/pull/324#discussion_r1397789015) - and 0 isnt a sentinel value really in this situation. Will add a comment!

test/Hooks.t.sol
```diff
@@ -519,12 +548,48 @@ contract HooksTest is Test, Deployers, GasSnapshot {
...
+        assertTrue(Hooks.shouldCallBeforeDonate(hookAddr));
+        assertFalse(Hooks.shouldCallAfterDonate(hookAddr));
+        assertTrue(Hooks.hasPermissionToNoOp(hookAddr));
+    }
+
+    function testValidateHookAddressFailsAllHooks(uint160 addr, uint8 mask) public {
+        uint160 preAddr = addr & uint160(0x007ffffFfffffffffFffffFFfFFFFFFffFFfFFff);
         vm.assume(mask != 0xff);
-        IHooks hookAddr = IHooks(address(uint160(preAddr) | (uint160(mask) << 152)));
+        IHooks hookAddr = IHooks(address(preAddr | (uint160(mask) << 152)));
```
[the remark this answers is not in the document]
hensha256: Oh i see this test doesnt even do what it wants - it can only set a flag up to 0xff but we now have more flags than that. Will fix it.

hensha256:
> Lastly, I think it will be good to write some NoOp tests with nested hooks/re-entrancy. Like swap on a hook who calls swap on a different hook that first NoOps and then doesn't... and vice versa

@snreynolds this is a post-access lock PR right? I'll add to the testing issue that we need to test the interactions of NoOp and Access Lock with this example!

zhongeric (approved):
Looks good!

snreynolds (approved):
🚢 🎊

PR #328, opened 2023-08-01
Use specific release tag for OpenZeppelin Contracts

Amxx:
Has mentionned [here](https://github.com/OpenZeppelin/openzeppelin-contracts#foundry-git), the default (master) branch of [Openzeppelin Contracts](https://github.com/OpenZeppelin/openzeppelin-contracts) is not safe for production. We recommand using a release branch or tag.


Amxx:
I might be missing something about how submodules work. Apparently the version used is fixed to 4.4.2 somewhere outside the `.gitmodules` file.

PR #329, opened 2023-08-02
Use a fixed tag for openzeppelin-contracts submodule

Amxx:
## Related Issue
Currently, the submodule for `lib/openzeppelin-contracts` does not point to any particular tag or branch. The consequence is that is someone does `forge update`, the latest commit from the default branch will be fetched.

Has mentionned [here](https://github.com/OpenZeppelin/openzeppelin-contracts#foundry-git), the default (master) branch of [Openzeppelin Contracts](https://github.com/OpenZeppelin/openzeppelin-contracts) is not safe for production. We recommand using a release branch or tag.

## Description of changes

Change the submodule file so that in the event of a `forge update`, the latest version is fetched.

Note: I arbitrarily set the tag to the latest release. Currently the submodule follows an earlier version (4.4). 4.9.3 is expected to be the latest release of the 4.x cycle. 5.0 will include breaking changes to ERC1155.

At this time, there is no way to point to the latest release of OpenZeppelin Contracts without in a future proof manner (that is comming). Still I think this change would act as a fail-safe, preventing someone from accidentally fetching an unsafe version of the dependency.


hensha256 (approved):
Thank youuuu!

hensha256:
Ah sorry - you havent signed your commit 😢 you can sign it in an interactive rebase!

Amxx:
@hensha256 Should be good now

ewilz:
should we stick with the original zeppelin version we've been on? 4.4.2?

ewilz (approved):

PR #330, opened 2023-08-03
CurrencyLibrary.balanceOf() for integrators

ewilz:
## Related Issue

```        
//ugly :( 
assertEq(IERC20(Currency.unwrap(currency0)).balanceOf(address(this)), 9 ether);
```

updatee: only adding balanceOf() and keeping out exclusively erc20 funcionality out of currency library...should just be unwrapped as an erc20

## Description of changes


marktoda (commented):
Ah yea this makes sense to me, makes it easier to use this custom type in hooks / integrations and stuff

Should we write / port some tests at least for the transferFrom impl? Also do we know what the actual gas savings is by inlining the transferFrom code rather than just importing solmate :D

snreynolds:
I like it, I think we just need to update the tests to use this?

ewilz:
@marktoda ughh I actually think I want to take out `transferFrom` since it's not compliant with native currency. Feels like a footgun :((((( actually to add ERC20 specific things here, so maybe just want to add balanceOf??

marktoda:
> transferFrom since it's not compliant with native currency

++ makes sense

> so maybe just want to add balanceOf

yea think this one is useful for sure

emmaguo13 (commented):
looks good to me 👍

hensha256 (approved):

PR #331, opened 2023-08-07
add snapshot to CI

snreynolds:
## Related Issue

We should be running forge snapshots on all PRs locally & on CI. There was an issue with getting the fuzz snapshots locally to match the CI. Hopefully we can resolve it in this PR.

## Description of changes


snreynolds:
trying to fix #296

marktoda:
tl;dr had to set the fuzz seed _in the environment_ because forge snapshot does not respect it from foundry.toml currently (issue below). So now we should use the `justfile` to run with the proper seed i.e. `just snapshots`

Reference to foundry issue https://github.com/foundry-rs/foundry/issues/6179

foundry.toml
```diff
@@ -8,7 +8,11 @@ ffi = true
 fs_permissions = [{ access = "read-write", path = ".forge-snapshots/"}, { access = "read", path = "./foundry-out"}]
 evm_version = 'paris'
 
-[profile.ci]
-fuzz_runs = 100000
+[profile.default.fuzz]
+runs = 100
```
snreynolds, line 12: why are the runs different in default vs ci?

marktoda: So it runs faster locally! Lots of runs can make it quite slow esp with lot of fuzz tests

marktoda: So it runs faster locally! Lots of runs can make it quite slow esp with lot of fuzz tests

snreynolds:
closing this in favor of this https://github.com/Uniswap/v4-core/pull/401 just squashes these, rebases main, and adds a tolerance

PR #335, opened 2023-08-10
Move MockERC20.sol to contracts/test

emmaguo13:
## Description of changes

MockERC20 should be used instead of TestERC20 in some cases, as it exposes more functions/external values for tests to use.


emmaguo13:
I did have to change the solmate remappings in remappings.txt, otherwise it would not work with hardhat 😐, and I wasn't able to yarn install a specific commit hash, so I went 1 version above the version installed in forge lib.

PR #337, opened 2023-08-13
feat: add PoolManager test

partylikeits1983:
## Description of changes
Adding test to PoolManager.t.sol to test the revert of an already initialized pool. This test expects revert when attempting to initialize an already initialized pool with the same exact PoolKey.


ewilz (approved):

ewilz:
good call

PR #332, opened 2023-08-09
Hooks accept arbitrary bytes from swapper + hook.getFee() accepts parameters

ewilz:
## Related Issue
Which issue does this pull request resolve?
fixes: https://github.com/Uniswap/v4-core/issues/307

about an 850 gas overhead for zero bytes for arbitrary data

## Description of changes
add arbitrary bytes for hooks

PR #343, opened 2023-08-25
Support arbitrary bytes on test routers

saucepoint:
## Related Issue

The test routers (i.e. `PoolSwapTest`) did not support arbitrary data arguments (new in #332). This PR updates the test routers to support the flow of arbitrary bytes from the router to the hooks.

## Description of changes

1. Updated the main function(s) in `PoolSwapTest`, `PoolModifyPositionTest`, and `PoolDonateTest` to support arbitrary data args

2. Updated `MockHooks` to write arbitrary data to storage. Used to verify that test routers are providing arbitrary bytes correctly (`Hooks.t.sol`)


saucepoint:
@ewilz any chance i can get reviewers on this?

snreynolds (dismissed):
lgtm!

contracts/test/PoolSwapTest.sol
```diff
@@ -35,21 +36,29 @@ contract PoolSwapTest is ILockCallback {
         payable
         returns (BalanceDelta delta)
     {
-        delta =
-            abi.decode(manager.lock(abi.encode(CallbackData(msg.sender, testSettings, key, params))), (BalanceDelta));
+        return swap(key, params, testSettings, new bytes(0));
```
ewilz, line 39: looks like gas went up by >1k for all swaps...wonder if consolidating the swap function here makes more sense than having 2 external calls (or just keeping them separate)

contracts/test/PoolModifyPositionTest.sol
```diff
@@ -22,14 +22,23 @@ contract PoolModifyPositionTest is ILockCallback {
...
+        bytes hookData;
     }
 
     function modifyPosition(PoolKey memory key, IPoolManager.ModifyPositionParams memory params)
         external
         payable
         returns (BalanceDelta delta)
     {
-        delta = abi.decode(manager.lock(abi.encode(CallbackData(msg.sender, key, params))), (BalanceDelta));
+        return modifyPosition(key, params, new bytes(0));
```
ewilz, line 33: same as swap, I wonder if we can minimize the impact on gas by the test contracts

snreynolds:
Also looks like the hardhat tests are failing. Since we still have some coverage with hardhat unfortunately we should get those updated even though we are in the process of deprecating. Can reproduce with `yarn test`.

contracts/test/PoolSwapTest.sol
saucepoint: I prefer removing the old function signature and moving everything over to the new function signature w/ `bytes memory hookData` (instead of duplicating the logic). I wanted to keep the original for backwards compatibility with v4-periphery and community repos, but I suppose theres been breaking changes already so might as well keep it up 😝 

Will try it for both swap + modifyPosition

contracts/test/PoolModifyPositionTest.sol
saucepoint: I removed the old `modifyPosition()` and the gas is practically the same, or worse in some cases 🤔 

Here's the diff
https://github.com/saucepoint/v4-core/pull/1/files

saucepoint:
To recap the outstanding changes, there's two paths we could take:

1) Remove overloaded functions by consolidating to the function(s) with `bytes memory` parameters
    - Gas improvements are negligible or worse
    - Update hardhat tests to the new function signatures
    - Update v4-periphery to the new function signatures
    - Requires users to provide ZERO_BYTES when using the test routers
    - More work, but cleaner codebase

2) Keep overloaded functions
    - Hardhat tests are ugly because they need to reference the full function signature
    - backwards compatible with v4-periphery and existing repos
    - cleaner interface for users that dont need arbitrary bytes

snreynolds:
> To recap the outstanding changes, there's two paths we could take:
> 
> 1. Remove overloaded functions by consolidating to the function(s) with `bytes memory` parameters
>    
>    * Gas improvements are negligible or worse
>    * Update hardhat tests to the new function signatures
>    * Update v4-periphery to the new function signatures
>    * Requires users to provide ZERO_BYTES when using the test routers
>    * More work, but cleaner codebase
> 2. Keep overloaded functions
>    
>    * Hardhat tests are ugly because they need to reference the full function signature. But these are getting deprecated anyway 🙏
>    * backwards compatible with v4-periphery and existing repos
>    * cleaner interface for users that dont need arbitrary bytes

I'm in favor of option #1. Don't think we need to maintain backwards compatibility in tests - we should just update them to match the new functionality. Gas seems to just be overhead of the new ZERO_BYTES param so I'm not that worried about it but curious to hear your thoughts @ewilz.

contracts/test/PoolSwapTest.sol
```diff
@@ -23,33 +23,34 @@ contract PoolSwapTest is ILockCallback {
...
-        returns (BalanceDelta delta)
-    {
-        delta =
-            abi.decode(manager.lock(abi.encode(CallbackData(msg.sender, testSettings, key, params))), (BalanceDelta));
+    function swap(
+        PoolKey memory key,
+        IPoolManager.SwapParams memory params,
+        TestSettings memory testSettings,
+        bytes memory hookData
+    ) public payable returns (BalanceDelta delta) {
```
snreynolds, line 39: external?

contracts/test/PoolDonateTest.sol
```diff
@@ -24,14 +24,25 @@ contract PoolDonateTest is ILockCallback {
         PoolKey key;
         uint256 amount0;
         uint256 amount1;
+        bytes hookData;
     }
 
     function donate(PoolKey memory key, uint256 amount0, uint256 amount1)
```
snreynolds, line 30: Remove overloaded func in favor of just using the new func signature below

contracts/test/PoolDonateTest.sol
```diff
@@ -24,14 +24,25 @@ contract PoolDonateTest is ILockCallback {
...
         external
         payable
         returns (BalanceDelta delta)
     {
-        delta = abi.decode(manager.lock(abi.encode(CallbackData(msg.sender, key, amount0, amount1))), (BalanceDelta));
+        return donate(key, amount0, amount1, new bytes(0));
+    }
+
+    function donate(PoolKey memory key, uint256 amount0, uint256 amount1, bytes memory hookData)
+        public
```
snreynolds, line 39: external

contracts/test/PoolModifyPositionTest.sol
```diff
@@ -22,14 +22,15 @@ contract PoolModifyPositionTest is ILockCallback {
         address sender;
         PoolKey key;
         IPoolManager.ModifyPositionParams params;
+        bytes hookData;
     }
 
-    function modifyPosition(PoolKey memory key, IPoolManager.ModifyPositionParams memory params)
-        external
+    function modifyPosition(PoolKey memory key, IPoolManager.ModifyPositionParams memory params, bytes memory hookData)
+        public
```
snreynolds, line 29: external

snreynolds (dismissed):

snreynolds:
@saucepoint Hm looks like hardhat tests are still failing, and I think you need to run yarn prettier but other than that happy to merge!

snreynolds (approved):

snreynolds:
Closing in favor of https://github.com/Uniswap/v4-core/pull/361

PR #353, opened 2023-09-08
use custom storage for lockData

snreynolds:
## Related Issue
By changing the lockData structure to use custom storage, we are setting up the repository to leverage tstore/tload.
## Description of changes
Changes the LockDataLibrary to account for the sentinel node in storage.


contracts/interfaces/IPoolManager.sol
```diff
@@ -112,19 +113,11 @@ interface IPoolManager is IFees, IERC1155 {
...
-        uint128 nonzeroDeltaCount;
-    }
-
     /// @notice Returns the locker in the ith position of the locker queue.
     function getLock(uint256 i) external view returns (address locker);
 
-    /// @notice Returns lock data
-    function lockData() external view returns (uint128 length, uint128 nonzeroDeltaCount);
+    /// @notice Returns the sentinel node for the LockData structure.
+    function getLockSentinel() external view returns (LockSentinel sentinel);
```
ewilz, line 120: whats the problem with continuing to call this `LockData`? I think I feel extremely opposed to the term "sentinel" as it's not descriptive and means [something else](https://en.wikipedia.org/wiki/Sentinel_value) in the context of computer programming :P

contracts/libraries/LockDataLibrary.sol
```diff
@@ -2,49 +2,92 @@
...
-            // update the sentinel storage slot
-            self.length = length + 1;
+            lockData.update(_length + 1, lockData.nonzeroDeltaCount());
         }
     }
 
     /// @dev Pops a locker off the end of the queue. Note that no storage gets cleared.
-    function pop(IPoolManager.LockData storage self) internal {
+    function pop(LockData lockData) internal {
+        uint128 _length = lockData.length();
```
0xb10ckdev, line 36: It doesn't need to save the length to a temporary variable.
`lockData.length()` can be used directly in the below code

contracts/libraries/LockDataLibrary.sol
```diff
@@ -2,49 +2,91 @@
...
+    }
+
+    function nonzeroDeltaCount(LockData lockData) internal pure returns (uint128 _nonzeroDeltaCount) {
+        /// @solidity memory-safe-assembly
+        assembly {
+            _nonzeroDeltaCount := lockData
+        }
+    }
+
+    function update(LockData lockData, uint128 _length, uint128 _nonzeroDeltaCount) internal {
```
ewilz, line 55: for readability/less footguns, wonder if we can break this into 3 functions:
`lockData.incrementNonzeroDeltaCount()`
`lockData.decrementNonzeroDeltaCount()`
`lockData.delete()`

probably at the expense of some bytecode savings, but I think it would make a lot of the code nicer to look at

contracts/types/LockData.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.20;
+
+import {LockDataLibrary} from "../libraries/LockDataLibrary.sol";
```
ewilz, line 4: keeping this in a separate file is breaking convention with types Currency + BalanceDelta. Wonder if it's worth migrating the library here

ewilz:
can rename PR to something like: use custom storage for LockData?

contracts/PoolManager.sol
```diff
@@ -138,14 +135,16 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
 
     /// @inheritdoc IPoolManager
     function lock(bytes calldata data) external override returns (bytes memory result) {
+        LockData lockData = LockDataLibrary.getLockData();
```
ewilz, line 138: something definitely feels a bit weird to me about using pop/push (write functionality) on a type uint256 that actually does storage writes. I wish there was a nice way to return a reference to storage here instead of recaching. Feels a bit dirty but can't think of any great solutions...

ewilz: once there's a full solc library that has the transient keyword, I think it's worth toying around with having a storage slot (unless we can call things directly on storage here with the current lib). I know it saves gas, but I'm wondering if the code cleanliness/quality is getting a little bit wonky here

ewilz (changes requested):
Wondering if we can tighten the library up by changing update to more opinionated functions otherwise library integrators can just input incoherent data right into storage. 

Porbably not good the library performs storage writes using any arbitrary uint256 even if it doesn't match what's currently in storage. For push and pop, can we just call like this:
```
DataLockLibrary.push()
DataLockLibrary.pop()
```

then inline `LockData lockData = getLockData();` as the first line of those functions instead?

contracts/PoolManager.sol
snreynolds: Yeah it is a bit wonky..let me know if you can think of another way, but I had to move away from a library for storage pointers because we don't yet have the `transient` keyword. 

We can rewrite the lib to use the transient keyword when there is solidity support.

snreynolds:
> Wondering if we can tighten the library up by changing update to more opinionated functions otherwise library integrators can just input incoherent data right into storage.
> 
> Porbably not good the library performs storage writes using any arbitrary uint256 even if it doesn't match what's currently in storage. For push and pop, can we just call like this:
> 
> ```
> DataLockLibrary.push()
> DataLockLibrary.pop()
> ```
> 
> then inline `LockData lockData = getLockData();` as the first line of those functions instead of getting it outside?

The reason I didn't do that is because as its written right now it would be easy to replace the in-memory lockData variable with a transient storage variable `lockData` and then write the library over that. So I'm just prepping for a more minimal diff later on. 

I can change the `update` function though so that we have more clear actions over the lockData var

contracts/types/LockData.sol
snreynolds: moved it to types/LockData.sol to match

contracts/libraries/LockDataLibrary.sol
snreynolds: Ok i added these functions but they all use similar code paths.. which I separated into toLockData and _update which does the final write

contracts/types/LockData.sol
```diff
@@ -0,0 +1,123 @@
...
+    function decrementNonzeroDeltaCount(LockData lockData) internal {
+        uint128 currLength = lockData.length();
+        uint128 newNonzeroDeltaCount = lockData.nonzeroDeltaCount() - 1;
+        LockData newLockData = toLockData(currLength, newNonzeroDeltaCount);
+        _update(newLockData);
+    }
+
+    function incrementNonzeroDeltaCount(LockData lockData) internal {
+        uint128 currLength = lockData.length();
+        uint128 newNonzeroDeltaCount = lockData.nonzeroDeltaCount() + 1;
```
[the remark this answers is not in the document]
hensha256: same with decrement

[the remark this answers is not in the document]
hensha256: although im pretty sure, if youre happy with `unchecked`, you could actually just do
```
unchecked{ uint256(lockData)++ }
```

contracts/types/LockData.sol
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    function nonzeroDeltaCount(LockData lockData) internal pure returns (uint128 _nonzeroDeltaCount) {
+        /// @solidity memory-safe-assembly
+        assembly {
+            _nonzeroDeltaCount := lockData
+        }
+    }
+
+    function _update(LockData lockData) internal {
```
hensha256, line 73: the use of `_` feels super sporadic - all the functions are internal to the library but some have _ and some dont?

contracts/types/LockData.sol
```diff
@@ -0,0 +1,123 @@
...
+            )
+    }
+}
+
+/// @dev This library manages a custom storage implementation for a queue
+///      that tracks current lockers. The LOCK_DATA storage slot for this data structure,
+///      passed in as a LockData custom type that is a uint256, stores not just the current
+///      length of the queue but also the global count of non-zero deltas across all lockers.
+///      The values of the data structure start at LOCKERS, and each value is a locker address.
+library LockDataLibrary {
```
hensha256, line 28: I found this library quite hard to read tbh. i think a few things that would help:
- Function ordering - currently you have to really jump around to read the library. e.g. `_update` is defined in the middle of the library, and there are functions both above and below it that call it. I'd personally rather the functions were in an order where all the function they call are defined above it (or below it if you prefer).
- Comments! Lots of the functions have no comments at all

contracts/types/LockData.sol
```diff
@@ -0,0 +1,123 @@
...
+    using LockDataLibrary for LockData;
+
+    uint256 private constant LOCK_DATA = uint256(keccak256("LockData"));
```
hensha256, line 31: maybe `LOCK_DATA_SLOT` is more descriptive? same for `LOCKERS_SLOT`

contracts/types/LockData.sol
```diff
@@ -0,0 +1,123 @@
...
+        unchecked {
+            uint256 lockerSlot = LOCKERS + _length; // not in assembly because LOCKERS is in the library scope
+
+            /// @solidity memory-safe-assembly
+            assembly {
+                // in the next storage slot, write the locker
+                sstore(lockerSlot, locker)
+            }
+            LockData newLockData = toLockData(_length + 1, lockData.nonzeroDeltaCount());
+            _update(newLockData);
```
hensha256, line 47: `newLockData._update()` ?

contracts/types/LockData.sol
```diff
@@ -0,0 +1,123 @@
...
+
+/// @notice The LockData holds global information for the lock data structure: the length and nonzeroDeltaCount.
+/// @dev The left most 128 bits is the length, or total number of active lockers.
+/// @dev The right most 128 bits is the nonzeroDeltaCount, or total number of nonzero deltas over all active + completed lockers.
+/// @dev Located in transient storage.
+type LockData is uint256;
+
+using LockDataLibrary for LockData global;
+
+function toLockData(uint128 _length, uint128 _nonzeroDeltaCount) pure returns (LockData lockData) {
```
hensha256, line 12: Given its going to be transient storage, should we consider just putting these in different slots? Every function that writes to this slot, only updates 1 of the 2 values - they never _both_ get updated. It feels like it would be easier to just separate them?
`push` - only touches length
`pull` - only touches length
`incrementNonzeroDeltaCount` - only touches delta count
`decrementNonzeroDeltaCount` - only touches delta count

Imo it might make more sense to have
```
NON_ZERO_DELTA_COUNT_SLOT = x
LOCKERS_SLOT = y
```

where
```
sload(LOCKERS_SLOT) => length
sload(LOCKERS_SLOT+1) => first locker
```
this is more like how actual arrays are encoded in storage

hensha256 (changes requested):
Left some thoughts about the setup of the library!

contracts/types/LockData.sol
snreynolds: interesting, yeah we could do that and compare gas. tload/tstore still 100 gas so its not crazy to pack though

snreynolds:
closing in favor of separating the libs diff shown here https://github.com/Uniswap/v4-core/pull/388

Issue #315, opened 2023-07-18
Migrate All tests to Foundry


ewilz:
### Describe the suggested feature and problem it solves.

Needing access to node-js is causing some dev-x issues, and instead of patching those, it seems like somewhat low effort to just get all our tests migrated and delete hardhat.

merged migrations:

- [x] sqrtPriceMath https://github.com/Uniswap/v4-core/pull/317
- [x] tickMath https://github.com/Uniswap/v4-core/pull/306
- [x] FullMath (https://github.com/Uniswap/v4-core/pull/314))
- [x] TickBitmap https://github.com/Uniswap/v4-core/pull/320
- [ ] SwapMath https://github.com/Uniswap/v4-core/pull/323
- [ ] PoolManager

ewilz:
If anyone feels like contributing to this, please leave a comment with which file you're migrating. This would be SO HELPFUL!

MwitahJob:
Working on swapMath.spec.ts

MwitahJob:
Working on swapMath.spec.ts

MwitahJob:
Hello could you help me a little, they are some functions imported into the hardhat test which are written in typescript.

``` solidity
import { encodeSqrtPriceX96, expandTo18Decimals } from './shared/utilities'
```

If you don't mind, how do I go around it? Sorry it's my first time.🙂

gitcoindev:
Hi! I would like to contribute and pick up SqrtPriceMath, will open a PR till Tuesday EOD.

gitcoindev:
Hi @ewilz I second the question from @MwitahJob . just after I created a skeleton I encountered the same functions. If I understand correctly, the typescript helper function implementations used in tests must be ported to solidity.

In this particular case for expandTo18Decimals we have number * 10^18

```
export function expandTo18Decimals(n: number): BigNumber {
  return BigNumber.from(n).mul(BigNumber.from(10).pow(18))
}
```

and we could potentially use an example from Solidity types https://docs.soliditylang.org/en/v0.8.20/types.html for `UFixed256x18` 

```
 /// Turns a uint256 into a UFixed256x18 of the same value.
    /// Reverts if the integer is too large.
    function toUFixed256x18(uint256 a) internal pure returns (UFixed256x18) {
        return UFixed256x18.wrap(a * multiplier);
    }
```

I just want to double check if this is the correct way or perhaps Uniswap already has a preferred Solidity library to use?

gitcoindev:
[...]

PR #354, opened 2023-09-11
Migrate functionality from setProtocolFeeController to testPoolManagerFeeControllerSet

smatthewenglish:
## Related Issue

This PR addresses part of Issue #315, porting `PoolManager` test cases from hardhat to foundry. 

## Description of changes

Finish porting the final piece of functionality from `setProtocolFeeController` in `PoolManager.spec.ts` to the associated test case `testPoolManagerFeeControllerSet` in `PoolManager.t.sol`. 

These tests were almost the same, except that `testPoolManagerFeeControllerSet` didn't check for the `ProtocolFeeControllerUpdated` event, now it checks for that event so it seems safe to remove the redundant test `setProtocolFeeController` from `PoolManager.spec.ts`.

### DIFF
--- a/test/foundry-tests/PoolManager.t.sol
+++ b/test/foundry-tests/PoolManager.t.sol
@@ -38,6 +38,7 @@ contract PoolManagerTest is Test, Deployers, TokenFixture, GasSnapshot, IERC1155
     using CurrencyLibrary for Currency;
 
     event LockAcquired();
+    event ProtocolFeeControllerUpdated(address protocolFeeController);
     event Initialize(
         PoolId indexed poolId,
         Currency indexed currency0,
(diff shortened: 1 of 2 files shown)


test/foundry-tests/PoolManager.t.sol
```diff
@@ -340,6 +341,8 @@ contract PoolManagerTest is Test, Deployers, TokenFixture, GasSnapshot, IERC1155
 
     function testPoolManagerFeeControllerSet() public {
         assertEq(address(manager.protocolFeeController()), address(0));
+        vm.expectEmit(false, false, false, true);
```
emagici, line 344:
```suggestion
        vm.expectEmit(true, true, true, true, address(manager));
```

smatthewenglish: Thanks for the suggestion!

In my opinion it makes more sense to specify the first three values as `false` since the purpose of those arguments is to evaluate indexed topics, and the `ProtocolFeeControllerUpdated` event doesn't have any log values encoded as topics. 

Specifying the emitting contract though does seem like it adds value, since it's more comprehensive. I'll update it accordingly!

ewilz (approved):

Issue #347, opened 2023-09-01
Explore consolidating hook.beforeSwap() and hook.getFee()


ewilz:
### Describe the suggested feature and problem it solves.

These calls are extremely similar, except one returns a fee value and one returns a safety selector value. You'd probably only need to use one or the other depending on whether you need a dynamic fee. Otherwise, they both have all the same access/capabilities.

relevant code
```
        if (key.hooks.shouldCallBeforeSwap()) {
            if (key.hooks.beforeSwap(msg.sender, key, params, hookData) != IHooks.beforeSwap.selector) {
                revert Hooks.InvalidHookResponse();
            }
        }

        // Set the total swap fee, either through the hook or as the static fee set an initialization.
        uint24 totalSwapFee;
        if (key.fee.isDynamicFee()) {
            totalSwapFee = IDynamicFeeManager(address(key.hooks)).getFee(msg.sender, key, params, hookData);
            if (totalSwapFee >= 1000000) revert FeeTooLarge();
        } else {
            // clear the top 4 bits since they may be flagged for hook fees
            totalSwapFee = key.fee.getStaticFee();
        }
```

pros: 
- code consolidation
- eliminate the awkwardness in the capability to enable both

cons:
- breaks cleanliness of hooks interfaces (beforeSwap would have a different returnVal)

Gas savings and cleanliness are determining factors?

### Describe alternatives.

...leave as is :)

snreynolds:
One other alternative is that we could cache the fee and not have to call out to getFee each time?

paco0x:
> One other alternative is that we could cache the fee and not have to call out to getFee each time?

Can we use this mechanism to prevent the frontrunning issue somehow?

For now, the contract calls `hook.getFee()` to get the fee ratio before every swap.

If there's a malicious hook contract, setting the fee ratio to `0` to deceive the router or other off-chain quoting services. Once users send tx to swap through this pool, the hook admin can front-run the tx by setting the fee ratio to a high value and causing users' loss.

hensha256:
Now that fees are push not pull #501 this is no longer needed

PR #360, opened 2023-09-26
Cache dynamic fee in slot0

snreynolds:
## Related Issue
https://github.com/Uniswap/v4-core/issues/347

## Description of changes
Caches uint24 dynamicFee in slot0 which should save gas in the cases where you don't need to update the fee on every swap (ie update once per block).


Imebeez (approved):

contracts/interfaces/IDynamicFeeManager.sol
```diff
@@ -7,7 +7,5 @@ import {IPoolManager} from "./IPoolManager.sol";
 /// @notice The dynamic fee manager determines fees for pools
 /// @dev note that this pool is only called if the PoolKey fee value is equal to the DYNAMIC_FEE magic value
 interface IDynamicFeeManager {
-    function getFee(address sender, PoolKey calldata key, IPoolManager.SwapParams calldata params, bytes calldata data)
-        external
-        returns (uint24);
+    function getFee(address sender, PoolKey calldata key) external view returns (uint24);
```
ewilz, line 10: params seem pretty important here (hooks may change fee based on size or direction of swap). Can't find gas benchmark notion, but guessing the benchmarks assumed beforeSwap gets called and sets the dynamic fee on every swap?? (likely tstore-ing some information about the swap in the hook for the call)

contracts/Fees.sol
```diff
@@ -61,6 +62,13 @@ abstract contract Fees is IFees, Owned {
         }
     }
 
+    function _fetchDynamicFee(PoolKey memory key) internal view returns (uint24 dynamicFee) {
+        if (key.fee.isDynamicFee()) {
+            dynamicFee = IDynamicFeeManager(address(key.hooks)).getFee(msg.sender, key);
+            if (dynamicFee >= 1000000) revert FeeTooLarge();
```
zhongeric, line 68: nit: extract into a constant?

contracts/interfaces/IDynamicFeeManager.sol
zhongeric: agree and thought that the ability to pass through SwapParams to the hook would be a compelling use case for dynamic fees

snreynolds: If you want to change the fee based on the swap you already have them passed through in`beforeSwap`. If you look at the test [here](https://github.com/Uniswap/v4-core/blob/dynamic-fee/contracts/test/DynamicFeesTest.sol#L28) you can see that the DynamicFee hook has all the info it needs already. The `getFee` function is just the call to the contract made in `_fetchDynamicFee` so there are not even params to pass through.

contracts/Fees.sol
```diff
@@ -61,6 +65,13 @@ abstract contract Fees is IFees, Owned {
         }
     }
 
+    function _fetchDynamicFee(PoolKey memory key) internal view returns (uint24 dynamicFee) {
```
marktoda, line 68: nit: specify as swapFee? ie `_fetchDynamicSwapFee` and return `dynamicSwapFee` for clarity given we also have withdrawal fees?

contracts/PoolManager.sol
```diff
@@ -122,7 +122,9 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
         PoolId id = key.toId();
         uint24 protocolFees = _fetchProtocolFees(key);
         uint24 hookFees = _fetchHookFees(key);
-        tick = pools[id].initialize(sqrtPriceX96, protocolFees, hookFees);
+        uint24 dynamicFee = _fetchDynamicFee(key);
```
marktoda, line 125: why do we put these into locals when we never need them again? slightly cleaner imo to fetch them inside the initialize call below as it better expresses that they're only intended to be used there. also probably saves some stack operations

contracts/PoolManager.sol
```diff
@@ -391,6 +392,13 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
         emit HookFeeUpdated(id, newHookFees);
     }
 
+    function setDynamicFee(PoolKey memory key) external {
```
marktoda, line 395: I know you're following pattern of protocol/hook fees, but `set*` feels like the wrong verb here as generally that implies the params include the new value. I'd lean towards `update*`

contracts/PoolManager.sol
```diff
@@ -251,11 +253,11 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
...
+        PoolId id = key.toId();
         // Set the total swap fee, either through the hook or as the static fee set an initialization.
         uint24 totalSwapFee;
         if (key.fee.isDynamicFee()) {
-            totalSwapFee = IDynamicFeeManager(address(key.hooks)).getFee(msg.sender, key, params, hookData);
-            if (totalSwapFee >= 1000000) revert FeeTooLarge();
+            totalSwapFee = pools[id].slot0.dynamicFee;
         } else {
             // clear the top 4 bits since they may be flagged for hook fees
             totalSwapFee = key.fee.getStaticFee();
```
marktoda, line 263: cleanup idea -- now that the dynamic fee is stored in slot0, can we just also store the static fee in slot0? That would unify this branching and in Pool.initialize we could pass in either the static fee or current dynamic fee

This is kinda nice as then the "source of truth" for the current swap fee is in a single place in slot0 and the dynamism is just from updating that value via hook if allowed

contracts/PoolManager.sol
```diff
@@ -251,11 +253,11 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
             }
         }
 
+        PoolId id = key.toId();
         // Set the total swap fee, either through the hook or as the static fee set an initialization.
         uint24 totalSwapFee;
         if (key.fee.isDynamicFee()) {
-            totalSwapFee = IDynamicFeeManager(address(key.hooks)).getFee(msg.sender, key, params, hookData);
-            if (totalSwapFee >= 1000000) revert FeeTooLarge();
+            totalSwapFee = pools[id].slot0.dynamicFee;
```
marktoda, line 260: wondering how the optimizer handles fetching slot0 here vs fetching it in Pool.swap -- is it two sloads or does it somehow optimize one of them out? If it duplicates sloads I would consider how we could unify them to further benefit from the caching

contracts/PoolManager.sol
marktoda: note this is only worth it if the slot0 reading is 0 cost (which in theory it should be since we must fetch from slot0 for every swap anyways)

contracts/PoolManager.sol
marktoda: Maybe idea is that if we store both fees in slot0, we can just have Pool.swap(..) not even take fee as a swap param but read it from slot0 instead

contracts/interfaces/IDynamicFeeManager.sol
nzhl: What's the purpose of adding "view" for `getFee`. I guess there would be use cases that dynamic update fee state in getFee.

contracts/PoolManager.sol
snreynolds: ah thats a good point. i can just return the swapFee from this func. alternatively we can remove it from event too. (did the first option but lmk what you think)

contracts/interfaces/IDynamicFeeManager.sol
snreynolds: Is there a reason to make it not view?

contracts/libraries/Pool.sol
```diff
@@ -74,8 +74,9 @@ library Pool {
         int24 tick;
         uint24 protocolFees;
         uint24 hookFees;
+        // used for the swap fee, either static at initialize or updateable to the dynamic value
```
marktoda, line 77: "updateable to the dynamic value" a bit confusing - maybe just "or dynamic via hook"

contracts/libraries/Pool.sol
```diff
@@ -138,6 +145,12 @@ library Pool {
         self.slot0.hookFees = hookFees;
     }
 
+    /// @notice Only dynamic fee pools may update the swap fee.
+    function _setSwapFee(State storage self, uint24 swapFee) internal {
```
marktoda, line 149: just convention but think we should omit the underscore here? since it's called from other locations via library `using for`, think we generally use no underscore. Underscore would be for purely private functions i think

contracts/libraries/Pool.sol
```diff
@@ -138,6 +145,12 @@ library Pool {
...
+        if (self.slot0.sqrtPriceX96 == 0) revert PoolNotInitialized();
```
marktoda, line 150: maybe require that `self.key.isDynamicFee()`?

contracts/PoolManager.sol
```diff
@@ -388,6 +374,13 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
         emit HookFeeUpdated(id, newHookFees);
     }
 
+    function updateDynamicSwapFee(PoolKey memory key) external {
+        uint24 newDynamicSwapFee = _fetchDynamicSwapFee(key);
```
marktoda, line 378: if this is not a dynamic swap fee pool, this will return 0 and set the fee to 0 i think :scream:

marktoda: think `_fetchDynamicSwapFee` should revert if not dynamic

marktoda: or maybe just call it `_fetchFee` which fetches from the key if static or from `getFee` if dynamic

test/foundry-tests/DynamicFees.t.sol
```diff
@@ -95,4 +103,45 @@ contract TestDynamicFees is Test, Deployers, GasSnapshot {
...
+        bytes memory data = abi.encode(true, uint24(456));
+        swapRouter.swap(
+            key, IPoolManager.SwapParams(false, 1, SQRT_RATIO_1_1 + 1), PoolSwapTest.TestSettings(false, false), data
+        );
+        snapEnd();
+    }
+
+    function testDynamicFeeAndBeforeSwapHook() public {
+        dynamicFees.setFee(123);
+        manager.updateDynamicSwapFee(key);
```
marktoda, line 122: add test for calling this on a non-dynamic swap fee pool

marktoda (changes requested):
Looks sooo nice and clean, just few thoughts and issues!

contracts/libraries/Pool.sol
snreynolds: ah ok priv funcs makes sense

contracts/libraries/Pool.sol
snreynolds: Re this comment [here](https://github.com/Uniswap/v4-core/pull/360#discussion_r1378177399), the change I made actually moves the check up a level (so its directly inside updateDynamicSwapFee). I did this because we have a duplicate check in initialize if the check is actually inside _fetchDynamicSwapFee. 

With this change, then by the time we call `setSwapFee` we have already done that check.

So I think just this change fixes both this comment and the one below. But lmk what you think.

contracts/PoolManager.sol
snreynolds: Very good catchhh thanks. Moved it up a level

marktoda (approved):

marktoda:
Nice!

PR #362, opened 2023-09-26
Prevent ProtocolFeeController from bricking pool initialization on revert

### LINKED ISSUE #221: [Bug]: Bad Owner Can Brick `initialize` By Setting A Bad `protocolFeeController`
### Describe the bug

A malicious/compromised `owner` of the `PoolManager` contract could set a `protocolFeeController` via the `setProtocolFeeController` method such that `PoolManager.initialize` always reverts, making it impossible to create new pools. This could compromise integrating protocols which may expect the `initialize` method to always be successfully callable (assuming that other pre-conditions are met).

A bad protocol fee controller can cause `initialize` to revert in 2 ways:
1. Return a fee parameter that's in an invalid range
2. Cause the `try { ... } catch { ... }` block to revert by returning data that doesn't decode into a `(uint8, uint8)` pair as expected.

### Expected Behavior

Expected behavior would be that the `PoolManager` gracefully falls back to a default fee of `0` as the `try { ... } catch { ... }` block in `_fetchProtocolFees` implies. Alternatively, it should be made more explicit that the `owner` can globally shut down the creation of new pools via the `initialize` function.

### To Reproduce

1. Copy this PoC (https://gist.github.com/Philogy/6a7e4360a677cb38b477542d94222bc5) into `test/foundry-tests/BrickInitialize.t.sol`
2. Run `forge test --mp test/foundry-tests/BrickInitialize.t.sol -vvvv`
3. See that the `testCanBrick()` test passes with initialize reverting in 2 separate cases solely due to a bad protocol fee controller.

Note: The PoC showcases both methods of causing `initialize` to revert.

### Additional context

_No response_

zhongeric:
## Related Issue
Fix #221 

## Description of changes
As outlined in the linked issue, a malicious / incorrectly implemented ProtocolFeeController that reverts on calls to `protocolFeesForPool` would brick the PoolManager, preventing new pools from being able to be initialized. 

While the protocol fee decoding has changed since the issue was created and protocol fees cannot be incorrectly encoded, the fee controller can still revert, so added tests for that scenario.

It's important to note that calling `setProtocolFee` for a given pool with an invalid protocol fee controller will still revert.


test/foundry-tests/PoolManager.t.sol
```diff
@@ -227,6 +230,65 @@ contract PoolManagerTest is Test, Deployers, TokenFixture, GasSnapshot, IERC1155
         assertEq(slot0.sqrtPriceX96, sqrtPriceX96);
     }
 
+    function testPoolManagerInitializeSucceedsWithRevertingFeeController(uint160 sqrtPriceX96) public {
```
zhongeric, line 233: not super sure we need this test but wanted to make sure that hook fees were not affected

contracts/Fees.sol
```diff
@@ -43,7 +43,12 @@ abstract contract Fees is IFees, Owned {
                 protocolWithdrawFee = uint16(updatedProtocolFees & 0xFFF);
```
zhongeric, line 43: can this line revert?

snreynolds: it shouldnt, you could write some fuzz tests though on different protocolFee values returned though

snreynolds (changes requested):

contracts/Fees.sol
```diff
@@ -43,7 +43,12 @@ abstract contract Fees is IFees, Owned {
...
 
                 protocolFees = updatedProtocolFees;
-            } catch {}
+            } catch {
+                // note if the ProtocolFeeController reverts we default all protocol fees to 0
+                protocolSwapFee = 0;
+                protocolWithdrawFee = 0;
+                protocolFees = 0;
```
snreynolds, line 50: I dont think this actually fixes anything.. if the call reverts these values will already be 0 because they wouldn't have gotten reassigned above

zhongeric: If the call reverts it will revert `initialize` though which is what we wanted to prevent right

src/Fees.sol
```diff
@@ -32,24 +32,29 @@ abstract contract Fees is IFees, Owned {
...
             if (gasleft() < controllerGasLimit) revert ProtocolFeeCannotBeFetched();
-            try protocolFeeController.protocolFeesForPool{gas: controllerGasLimit}(key) returns (
-                uint24 updatedProtocolFees
-            ) {
-                protocolSwapFee = uint16(updatedProtocolFees >> 12);
-                protocolWithdrawFee = uint16(updatedProtocolFees & 0xFFF);
+            (bool _success, bytes memory data) = address(protocolFeeController).call{gas: controllerGasLimit}(
+                abi.encodeWithSelector(IProtocolFeeController.protocolFeesForPool.selector, key)
+            );
+            bytes32 _data;
```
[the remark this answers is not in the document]
snreynolds: Also just noting a few things cleaned up in the above snippet:

- can probably have less instantiated variables (ie. protocolFees can just be assigned once)
- data and _data is confusing and I'd use _data and _success as return vars from the low level call and then returnData or some other distinguishing variable when loading it from memory
- why do you assign _data as a bytes32. is there any reason to not just use uint256?
- when checking for overflow you dont need to mask you can just do` returnData == uint24(returnData) `

src/Fees.sol
```diff
@@ -32,24 +32,29 @@ abstract contract Fees is IFees, Owned {
         controllerGasLimit = _controllerGasLimit;
     }
 
-    function _fetchProtocolFees(PoolKey memory key) internal view returns (uint24 protocolFees) {
-        uint16 protocolSwapFee;
-        uint16 protocolWithdrawFee;
+    /// @notice Fetch the protocol fees for a given pool, returning false if the call fails or the returned fees are invalid.
+    /// @dev the success of this function must be checked when called in setProtocolFees
```
snreynolds, line 36: Could you add comment that we DONT check the success value in initialize because even if the call is unsuccessful we need to allow pools to initialize. New pools will just default initialize then to 0 protocolFee

src/Fees.sol
```diff
@@ -70,18 +75,34 @@ abstract contract Fees is IFees, Owned {
...
-        if (fee != 0) {
-            uint16 fee0 = fee % 64;
-            uint16 fee1 = fee >> 6;
-            // The fee is specified as a denominator so it cannot be LESS than the MIN_PROTOCOL_FEE_DENOMINATOR (unless it is 0).
-            if (
-                (fee0 != 0 && fee0 < MIN_PROTOCOL_FEE_DENOMINATOR) || (fee1 != 0 && fee1 < MIN_PROTOCOL_FEE_DENOMINATOR)
-            ) {
-                revert FeeTooLarge();
+    function _isValidProtocolFees(uint24 protocolFees) internal pure returns (bool) {
+        if (protocolFees != 0) {
```
snreynolds, line 79: this is all repeated code. Fine to expose _isValidProtocolFees but then we should have another internal function that performs the same checks as _checkProtocolFee since its copied just twice here. Want to reduce repeated code paths bc it opens surface area for bugs

snreynolds (changes requested):

src/Fees.sol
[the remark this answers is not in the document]
zhongeric: Thanks for the suggestions! Implemented

src/Fees.sol
zhongeric: yea agree, got rid of checkProtocolFee initially because it was the old behavior (reverting) but can re-add to reduce duplication

src/Fees.sol
```diff
@@ -71,17 +77,32 @@ abstract contract Fees is IFees, Owned {
...
+    }
+
+    function _isValidProtocolFees(uint24 protocolFees) internal pure returns (bool) {
+        if (protocolFees != 0) {
+            uint16 protocolSwapFee = uint16(protocolFees >> 12);
+            uint16 protocolWithdrawFee = uint16(protocolFees & 0xFFF);
+            if (protocolSwapFee != 0 && !_isFeeWithinBounds(protocolSwapFee)) {
+                return false;
+            }
+            if (protocolWithdrawFee != 0 && !_isFeeWithinBounds(protocolWithdrawFee)) {
```
hensha256, line 101: it checks for non-0 here and then again inside `_isFeeWithinBounds`. I think it can just check once - i'd probably just keep the one in `_isFeeWithinBounds`

hensha256: actually i feel like you can just do 
```
return _isFeeWithinBounds(protocolSwapFee) && _isFeeWithinBounds(protocolWithdrawFee)
```

hensha256 (changes requested):
little suggestion but mostly looking nice

test/PoolManager.t.sol
```diff
@@ -617,6 +617,29 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
         manager.setProtocolFees(key);
     }
 
+    function test_setProtocolFee_failsWithInvalidProtocolFeeControllers() public {
```
hensha256, line 620: i feel like this test could be kind of perfect for forge fuzzing
if forge sets a `uint256` input you can just do a bunch of if statements to check its behaving like you would expect
[see this test ](https://github.com/Uniswap/v4-core/blob/6b8ec7c6dff9a2bef4ff408f886b4cc1737179ae/test/PoolManagerInitialize.t.sol#L54)for an example

src/Fees.sol
snreynolds: Note that this will change when we have a PR for the changed fee model

snreynolds (commented):
Changes look good! Agree with some of Alice's comments but then can approve!

test/PoolManager.t.sol
zhongeric: Discussed offline and decided to leave as is for now, and after the fee refactor add a fuzz test for protocol swap fee updating

snreynolds: Are we tracking this in alice's PR? I think I would like to see some fuzz tests for setting the protocol fee and just on the unit tests around _isValidProtocolFee

lgtm though

snreynolds (approved):

hensha256 (approved):

PR #369, opened 2023-10-15
Set tick spacing range as constants

hyunchel:
## Related Issue
#371 

## Description of changes
While tick and tick spacing both use int24 as their type, each has a different range. Tick spacing has a range of [1, 32767]. This commit updates Tick test cases to use proper tick spacing range instead that of tick.


hyunchel:
Updated the commit to tag the related issue.

hyunchel:
Added and squashed missed commit.

test/foundry-tests/Tick.t.sol
```diff
@@ -105,18 +105,18 @@ contract TickTest is Test, GasSnapshot {
...
-        checkCantOverflow(1, maxLiquidityPerTick);
+        checkCantOverflow(Constants.MIN_TICK_SPACING, maxLiquidityPerTick);
     }
 
-    function testTick_tickSpacingToMaxLiquidityPerTick_returnsTheCorrectValueForEntireRange() public {
-        uint128 maxLiquidityPerTick = tickSpacingToMaxLiquidityPerTick(887272);
+    function testTick_tickSpacingToMaxLiquidityPerTick_returnsTheCorrectValueForMaxTickSpacing() public {
+        uint128 maxLiquidityPerTick = tickSpacingToMaxLiquidityPerTick(Constants.MAX_TICK_SPACING);
 
-        assertEq(maxLiquidityPerTick, Constants.MAX_UINT128 / 3); // 126 bits
```
zhongeric, line 118: Do you mind adding the additional test you mention in #371 to this test file as well? 
```
function testTick_tickSpacingToMaxLiquidityPerTick_returnsTheCorrectValueForEntireRange() public {
    uint128 maxLiquidityPerTick = tickSpacingToMaxLiquidityPerTick(Constants.MAX_TICK);

    assertEq(maxLiquidityPerTick, type(uint128).max / 3);
    checkCantOverflow(Constants.MAX_TICK, maxLiquidityPerTick);
}
```

zhongeric (approved):

zhongeric (changes requested):
Thanks for the contribution! Overall looks good to me, left one comment referencing something you outlined in your issue. Can you also merge in the latest main after addressing?

hyunchel:
Thanks for the review! I've addressed your request, and have merged main into this branch as well.

zhongeric:
Great thanks! 

On second review, I realized that we don't have `MIN/MAX_TICK_SPACING` in the TickMath library, do you mind adding them in there? I'd prefer to assign the constant to the fully resolved number with a comment saying that it's max int16. We should also probably replace all instances of `MIN/MAX_TICK` in the test file to read from TickMath instead of the test Constants file.

hyunchel:
Sure thing! I have replaced all the references of `MIN/MAX_TICK` to use TickMath file.

There is another declaration of `MIN/MAX_TICK_SPACING` in PoolManager which is required by an interface.
I tried to minimize the change here, but if you think otherwise, please let me know!

contracts/PoolManager.sol
```diff
@@ -34,10 +35,10 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
     using FeeLibrary for uint24;
 
     /// @inheritdoc IPoolManager
-    int24 public constant override MAX_TICK_SPACING = type(int16).max;
+    int24 public constant override MAX_TICK_SPACING = TickMath.MAX_TICK_SPACING;
```
snreynolds, line 38: I know you didn't introduce this change but I think that the `override` keyword is unnecessary.

snreynolds (approved):
nice. good change

contracts/PoolManager.sol
hyunchel: Thanks! Removed both overrides.

zhongeric (dismissed):
Looks good thanks for the contribution! Oops seems like there are some merge conflicts 😅 if you can fix them I'll approve again and merge for you.

hyunchel:
Merge conflicts are resolved now. Thanks for the reviews!

zhongeric (dismissed):

zhongeric:
Oh shoot looks like you have to run `forge snapshot` since we are checking in the snapshots file now

hyunchel:
Oops, sorry about that. Just pushed this PR with snapshot updates!

zhongeric (approved):

zhongeric:
Yay just merged! Thanks again for the contribution :)

Issue #216, opened 2023-06-13
[Bug]: Local snapshots do not match CI snapshots for fuzz tests


snreynolds:
### Describe the bug

There is an ongoing issue where `forge snapshot` locally does not match the CI snapshots. It is failing on on fuzz tests. I've checked the following which I thought could be related to this issue but are not:
- Local and CI using same fuzz runs. 
- No fuzz seed is being set in either env, which defaults to none. But, forge snapshot should already be using a deterministic seed, so this shouldn't be a problem.

### Expected Behavior

After updating the snapshot file locally, and pushing to your branch `forge snapshot --check` should pass in the linter.

### Additional context

PR reverted [here](https://github.com/Uniswap/v4-core/pull/215).

zhongeric:
Fixed by #401

PR #373, opened 2023-10-16
Compare gas snapshots in CI with a small tolerance

hyunchel:
## Related Issue
Resolves issue #216

## Description of changes
The previous attempt to assert identical gas usage in CI has been reverted because a snapshot generated in CI was different from a snapshot generated locally.

The diff still exist as of today, but amounts are minuscule. The amounts range from 1 to tens. And these diffs exist in fuzz tests only. This is a known bug in Foundry: https://github.com/foundry-rs/foundry/issues/5689

This commit adds back gas snapshot, and check for a small diff in CI. By allowing just one percent of diff, forge snapshot diff tool can be utilized while waiting for the Foundry issue to be fixed.


marktoda:
closing in favour of https://github.com/Uniswap/v4-core/pull/331

PR #383, opened 2023-11-01
fix: add gas snaps for swaps with 1155 as input/output

zhongeric:
## Related Issue
Add two small gas snaps for testing swapping into/out of ERC1155s

## Description of changes


snreynolds (changes requested):
Lgtm can you just change the name of the tests to include "Gas". just to be consistent w the other gas snap tests

snreynolds (approved):

PR #384, opened 2023-11-02
ETH WETH fungible

### LINKED ISSUE #292: Treat WETH/ETH Interchangeably
### Component

_No response_

 In the take() and settle() (as well as any ETH pools) WETH/ETH become interchangeable and users may withdraw or deposit either token.
**Pros** 
- Shared liquidity for pools with eth/weth.
- Integrations don’t need to worry about wrap/unwrap

**Con:** 
- Must always check balance before withdrawal for solvency, must unwrap/wrap when balance is too low for desired currency.
- Core becomes opinionated about the canonical WETH.
- Muddying core with more logic that could be dealt with elsewhere
- Griefing poolManager is trivial (griefer can set poolManager to have only ETH for minimal gas)
- Solving liquidity fragmentation seems futile since it's already running rampant in v4 anyway

_No response_

### Describe alternatives.

_No response_

### Additional context.

_No response_

hensha256:
## Related Issue
Closes #292 

## Description of changes
- Pools can only be created with ETH, not WETH
- All accounting is done in ETH
- However you can take/settle ETH deltas using WETH by just passing in the WETH address


.forge-snapshots/donate gas with 1 token.snap
```diff
@@ -1 +1 @@
-95961
\ No newline at end of file
+97672
```
hensha256, line 1: these look large, but our gas tests arent actually just checking the call to the contract... theyre including all the logic in the `PoolDonateTest` or `PoolSwapTest` contracts.

.gitignore
```diff
@@ -4,4 +4,5 @@ crytic-export/
 node_modules/
 typechain/
 foundry-out/
-.vscode/
\ No newline at end of file
+.vscode/
+solmate
```
marktoda, line 8: why do we have solmate top level anyways? it should be inside lib/ no?

marktoda: also it should be a submodule anyways so we don't want to gitignore it?

contracts/PoolManager.sol
```diff
@@ -318,8 +329,13 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
...
     /// @inheritdoc IPoolManager
     function take(Currency currency, address to, uint256 amount) external override noDelegateCall onlyByLocker {
-        _accountDelta(currency, amount.toInt128());
-        reservesOf[currency] -= amount;
+        Currency accountCurrency = (currency == WRAPPED_NATIVE) ? CurrencyLibrary.NATIVE : currency;
+
+        _accountDelta(accountCurrency, amount.toInt128());
+        reservesOf[accountCurrency] -= amount;
+
+        if (currency == WRAPPED_NATIVE) WETH_CONTRACT.deposit{value: amount}();
```
marktoda, line 337: can we instead abstract this into `CurrencyLibrary`? ie have `Currency.transfer(to, amount)` do wrap + transfer if WETH given as input token

contracts/PoolManager.sol
```diff
@@ -331,11 +347,22 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
 
     /// @inheritdoc IPoolManager
     function settle(Currency currency) external payable override noDelegateCall onlyByLocker returns (uint256 paid) {
-        uint256 reservesBefore = reservesOf[currency];
-        reservesOf[currency] = currency.balanceOfSelf();
-        paid = reservesOf[currency] - reservesBefore;
+        Currency accountCurrency = (currency == WRAPPED_NATIVE) ? CurrencyLibrary.NATIVE : currency;
+        uint256 balance = currency.balanceOfSelf();
+
+        if (currency == WRAPPED_NATIVE) {
```
marktoda, line 353: same with moving into CurrencyLibrary

Mostly concerned with adding so much non-generic logic into these very simple generic functions

marktoda: also thinking now that we have custom accounting, would we be better off just having a CA ETH/WETH pool? and enforcing WETH pairs only against ETH

contracts/PoolManager.sol
```diff
@@ -49,9 +50,17 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC1155, IERC1155Rec
...
+    WETH public immutable WETH_CONTRACT;
+
+    Currency public immutable WRAPPED_NATIVE;
+
     mapping(PoolId id => Pool.State) public pools;
 
-    constructor(uint256 controllerGasLimit) Fees(controllerGasLimit) ERC1155("") {}
+    constructor(uint256 controllerGasLimit, WETH wrappedNative) Fees(controllerGasLimit) ERC1155("") {
+        WETH_CONTRACT = wrappedNative;
+        WRAPPED_NATIVE = Currency.wrap(address(WETH_CONTRACT));
```
snreynolds, line 62: Are you saving WRAPPED_NATIVE as an immutable bc there is some gas cost overhead to wrap? Or another reason?

test/foundry-tests/PoolManager.t.sol
```diff
@@ -258,6 +267,27 @@ contract PoolManagerTest is Test, Deployers, TokenFixture, GasSnapshot, IERC1155
...
+
+        // currency1 just a placeholder until the currencies are set below
+        PoolKey memory key =
+            PoolKey({currency0: currency1, currency1: currency1, fee: 3000, hooks: IHooks(address(0)), tickSpacing: 60});
+
+        (key.currency0, key.currency1) =
+            (currencyAdd < wrappedNative) ? (currency, wrappedCurrency) : (wrappedCurrency, currency);
+
+        if (currencyAdd == wrappedNative) {
+            vm.expectRevert(IPoolManager.CurrenciesInitializedOutOfOrder.selector);
```
snreynolds, line 284: A bit random but reading this test made me realize maybe there should be a different error for when you try to initialize the pool with the same currencies?

contracts/PoolManager.sol
snreynolds: Would we even need CA to do this?

marktoda: I was thinking a hook that noops swap and instead just takes ETH and wraps it into WETH.. could avoid CA by just adding a wrapETH command but this feels nicer since it fits nicely into swap paths and  were adding CA anywas

contracts/PoolManager.sol
snreynolds: Only thought would be then to make Currency not a library and just a contract.. then we could pass it through to the inherited contract. Not great but can't think of anything else

contracts/PoolManager.sol
snreynolds: I see.. but then dont you also have to enforce that the eth/weth pool is a CA pool if you want to avoid having weth inside the contract. And then Im not sure how you enforce that

.gitignore
hensha256: ohh it was top level bc when I started this PR we still had hardhat (and for various reasons hardhat needs it top level lol). Deleting!

hensha256: When its top level its still a submodule in `lib` but then you do a symlink at the top level so hardhat can find it

test/foundry-tests/PoolManager.t.sol
hensha256: yeeeeah i did it just bc its cheaper gas-wise for `initialize` to just have 1 if statement, and itll still cause people to look at the right line. But it is a bit odd
I could change `CurrenciesInitializedOutOfOrder` to like `CurrenciesOutOfOrderOrIdentical` or something?

PR #388, opened 2023-11-06
New locker lib: unpacks length and nonzeroDeltaCount

snreynolds:
## Related Issue

## Description of changes


contracts/libraries/Lockers.sol
```diff
@@ -0,0 +1,94 @@
...
+            count := add(count, 1)
+            tstore(slot, count)
+        }
+    }
+
+    function decrementNonzeroDeltaCount() internal {
+        uint256 slot = NONZERO_DELTA_COUNT;
+        assembly {
+            let count := tload(slot)
+            count := sub(count, 1)
```
zhongeric, line 90: Do you think its worth catching the underflow here?

contracts/libraries/Lockers.sol
```diff
@@ -0,0 +1,94 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.20;
+
+/// @notice This is a temporary library that allows us to use transient storage (tstore/tload)
+/// for the lockers array and nonzero delta count.
+library Lockers {
+    // The starting slot for an array of lockers, stored transiently.
+    uint256 constant LOCKERS_SLOT = uint256(keccak256("Lockers"));
```
hensha256, line 8: nitttt: the best-practice is to subtract 1 from this number, so that the preimage isnt known
see EIP1967!

contracts/libraries/Lockers.sol
```diff
@@ -0,0 +1,94 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.20;
+
+/// @notice This is a temporary library that allows us to use transient storage (tstore/tload)
+/// for the lockers array and nonzero delta count.
+library Lockers {
```
hensha256, line 6: Would be great to see some tests for this library!

hensha256 (commented):
Couple of things but i love this change with them separated!

contracts/libraries/Lockers.sol
snreynolds: No because we only call this if we've incremented prior so I think its ok and its an internal function

PR #392, opened 2023-11-07
Accumulate protocol fees as 6909s

zhongeric:
## Related Issue
#203 

## Description of changes
Accrue protocol fees to the protocol fee controller as 6909 balance, removes the ability for the owner to claim protocol fees. Could optionally allow the owner to claim if desired.

Hook fees untouched in this PR because they will be removed in favor of CA 

Gas report
```
Diff w/ 6909 (old fee accumulation logic)
| src/PoolManager.sol:PoolManager contract |                 |        |        |         |         |
|------------------------------------------|-----------------|--------|--------|---------|---------|
| Deployment Cost                          | Deployment Size |        |        |         |         |
| -19427                                   | -97             |        |        |         |         |
| Function Name                            | min             | avg    | median | max     | # calls |
| burn                                     | -8              | -8     | -8     | -9      | 0       |
| collectProtocolFees                      | -19965          | -7841  | -4096  | 540     | -1      |
| donate                                   | 22              | 23     | 22     | 22      | 0       |
| mint                                     | 0               | 8      | 9      | 11      | 0       |
| swap                                     | 0               | -398   | -181   | 22      | -1      |
```


src/PoolManager.sol
```diff
@@ -347,6 +347,17 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
         emit ProtocolFeeUpdated(id, newProtocolFees);
     }
 
+    function collectProtocolFees(address recipient, Currency currency, uint256 amount)
```
hensha256, line 350: This function is basically a function that allows users to collect the underlying collateral for their claim - yet its locked down to something only 1 address is allowed to do.
I think it me it would make more sense to make the `protocolFeeController` have to do whatever all other users have to do to collect their claims.
So either:

- get a lock
- call `burn`
- call `take`

Or open this function up to be a function anyone can call, on the Claim contract

test/PoolManager.t.sol
```diff
@@ -641,39 +641,18 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
-        assertEq(currency0.balanceOf(address(1)), 0);
         manager.collectProtocolFees(address(1), currency0, 0);
+        snapEnd();
         assertEq(currency0.balanceOf(address(1)), expectedFees);
-        assertEq(manager.protocolFeesAccrued(currency0), 0);
+        assertEq(manager.balanceOf(address(feeController), currency0), 0);
     }
 
-    function test_collectProtocolFees_nativeToken_allowsOwnerToAccumulateFees_gas() public {
+    function test_collectProtocolFees_nativeToken_returnsAllFeesIf0IsProvidedAsParameter_gas() public {
```
hensha256, line 655: Would love to see tests for:

- Collecting more fees than they have accrued, should revert
- Invalid caller (although i think we should be removing that, as stated in my previous comment)
- Collecting an amount of fees that is < whole balance
- Collecting a fee where 0 have been accrued (so passing in 0, and the balance is 0)

This could actually be a perfect use for a simple fuzz test:

- argument of `uint256`
- if == 0 expect they claim their whole balance
- if <= their balance expect they claim the amount specified
- if > their balance expect revert

hensha256 (changes requested):

test/PoolManager.t.sol
hensha256: lol sorry im a super picky test reviewer as of this week 😆

src/PoolManager.sol
zhongeric: Yep so initially I was planning on having protocol fee controller have to lock to collect fees like a normal user of Claims but I opted for this approach to keep the most backward compatibility with the current implementation.

I don't think we should open up an additional function in core to burn Claims for ERC20s since we already require lock/burn/take. I'm down to enforce that feeController locks to collect fees

test/PoolManager.t.sol
zhongeric: Lol all good yeah I think there def needs some more coverage here. I'll add that fuzz test!

src/PoolManager.sol
snreynolds: I like exposing this outside of lock to keep backwards compatibility

PR #395, opened 2023-11-09
Add solc binaries and transient storage lock library

snreynolds:
## Related Issue

## Description of changes
changes the locker library to use transient storage with tstore/tload and adds the appropriate solc binaries
`Lockers.sol` will be removed once we have the `transient` keyword


foundry.toml
```diff
@@ -4,9 +4,12 @@ optimizer_runs = 800
 via_ir = false
 ffi = true
 fs_permissions = [{ access = "read-write", path = ".forge-snapshots/"}, { access = "read", path = "./out"}]
-evm_version = 'paris'
+cancun = true
+solc = "./bin/solc"
 
 [profile.ci]
 fuzz_runs = 100000
+solc = "./bin/solc-static-linux"
```
marktoda, line 12: wait so does the other one only work on mac? thats awk for people who run non-mac locally

foundry.toml
```diff
@@ -4,9 +4,12 @@ optimizer_runs = 800
...
+cancun = true
```
marktoda, line 13: the default one should override this no?

marktoda: also why is it cancun = true rather than `evm_version = cancun` which is the standard foundry config?

src/libraries/Lockers.sol
```diff
@@ -0,0 +1,94 @@
...
+            count := add(count, 1)
+            tstore(slot, count)
+        }
+    }
+
+    function decrementNonzeroDeltaCount() internal {
+        uint256 slot = NONZERO_DELTA_COUNT;
+        assembly {
+            let count := tload(slot)
+            count := sub(count, 1)
```
marktoda, line 90: underflow concern here?

src/libraries/Lockers.sol
```diff
@@ -0,0 +1,94 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.20;
+
+/// @notice This is a temporary library that allows us to use transient storage (tstore/tload)
+/// for the lockers array and nonzero delta count.
```
marktoda, line 5: would help to make explicit TODO that we should remove this once solidity supports transience natively

marktoda (approved):

foundry.toml
snreynolds: Yeah there are different binaries for diff machines :/

foundry.toml
snreynolds: I think its just not supported yet? This was a custom config they did for us 

https://github.com/foundry-rs/foundry/issues/5782

src/libraries/Lockers.sol
snreynolds: Hm.. you're saying we should throw? or make it checked math? In PoolManager we will never get underflow but if other people are using this library that is a concern I guess

marktoda: not suggesting anything, just seems easy to accidentally underflow it

>  PoolManager we will never get underflow

ie true in current implementation but not directly enforced, only by multi step logical assumptions about how the deltas are zeroed out

maybe just a comment about the above would be useful

foundry.toml
marktoda: what if we made .env set the solc binary environment variable depending on os?
something like
```
if [[ "$OSTYPE" == "linux-gnu"* ]]; then
        FOUNDRY_SOLC = ...
elif [[ "$OSTYPE" == "darwin"* ]]; then
        FOUNDRY_SOLC = ...
else
        # Unknown.
fi
```

zsp00: windows users have questions?

marktoda: if youre a windows user can you `echo $OSTYPE` so we know what to put? I _think_ its "win32" but not sure

marktoda (dismissed):

.env
```diff
@@ -1 +1,7 @@
 FOUNDRY_FUZZ_SEED=0x4444
+
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+        FOUNDRY_SOLC="./bin/solc-static-linux"
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+        FOUNDRY_SOLC="./bin/solc"
```
marktoda, line 6: nit: rename to solc-mac for clarity

marktoda (approved):

marktoda (approved):

PR #402, opened 2023-11-13
Add {before/after}mint hooks, remove all modifyPosition hooks

### LINKED ISSUE #350: Split {before,after}ModifyPosition into {before,after}Mint and {before,after}Burn
### Component

Hooks

Right now, `beforeModifyPosition` is a risky permission. Liquidity providers need to audit and trust the hook code or risk never being able to withdraw their money. That makes it harder to fearlessly integrate pools that have this permission.

However, the vast majority of that risk is from the ability of the hook to interrupt _burns_, rather than mints. `beforeMint` would generally be a more harmless hook, since it would never prevent existing liquidity providers from withdrawing their money. If we split out the `modifyPosition` hooks into ones that affect only mints and burns respectively, we allow hooks that only need to be called on mints to avoid asking for dangerous permissions that they don't need.

Replace the hook calls in https://github.com/Uniswap/v4-core/blob/main/contracts/PoolManager.sol#L205 and https://github.com/Uniswap/v4-core/blob/main/contracts/PoolManager.sol#L243C1-L251C1 with switch statements that call the appropriate `beforeMint`/`beforeBurn`/`afterMint`/`afterBurn` based on whether `params.liquidityDelta` is positive or negative.

### Describe alternatives.

Hook contracts could use a switch statement at the beginning of their callback implementation to make it easy to see that they will never interfere with burn transactions. 

But this generally undermines the point of hook permissions. It still requires reading the code of the contract and can't be verified programmatically the way hook permissions can. 

And it would be trickier for hook contracts that are upgradeable (which would have to put that check in the proxy contract somehow).

### Additional context.

_No response_

zhongeric:
## Related Issue
Fixes #350 

## Description of changes

TODO:
- Add before/after burn


src/interfaces/IPoolManager.sol
```diff
@@ -151,8 +151,13 @@ interface IPoolManager is IFees, IERC1155 {
...
 
-    /// @notice Modify the position for the given pool
-    function modifyPosition(PoolKey memory key, ModifyPositionParams memory params, bytes calldata hookData)
+    /// @notice Mint a new position in a pool
+    /// @dev Poke by calling with a zero liquidityDelta
+    /// @param key The pool to mint a position in
+    /// @param params The parameters of pass into modifyPosition
+    /// @param hookData Any data to pass to the callback, via `ILockCallback(msg.sender).lockAcquired(data)`
+    /// @return delta The balance delta of the position minted
+    function mintPosition(PoolKey memory key, ModifyPositionParams memory params, bytes calldata hookData)
```
zhongeric, line 160: consider taking in different params than the underlying modifyPosition, i.e using uint256 for mint and burnPosition but casting to int for burn

snreynolds:
Closing - full PR https://github.com/Uniswap/v4-core/pull/434

Issue #366, opened 2023-10-12
Allow lockers to resolve deltas on behalf of other lockers


snreynolds:
### Component

Lock and Call

### Describe the suggested feature and problem it solves.

To add more flexibility for lockers and open up new use cases, this idea would be to introduce a new way to pay off deltas by adding the ability for a locker to resolve open deltas on behalf of a different locker.

### Describe the desired implementation.

Add a function `resolve` that nets out two lockers' deltas. Additionally, add the option for a locker to specify a "resolver" address. An initial proof of concept was explored [here](https://github.com/Uniswap/v4-core/compare/resolve), along with a hook that could NoOp swaps/modifyPositions to show the Custom Accounting use case.

ewilz:
the balanceDelta's returned by all functions basically become moot with this, correct? Bc accountDeltas could be completely different the the balanceDeltas returned to callers?

PR #404, opened 2023-11-14
Access lock

snreynolds:
## Related Issue
This is related to https://github.com/Uniswap/v4-core/issues/366. While this doesn't allow ANY locker to resolve deltas on behalf of others it allows hooks to have this behavior.

Hooks that elect an `ACCESS_LOCK_FLAG` will be able to call functions `swap`, `modifyPosition`, `donate`, `initialize` `take`, `settle`, and `mint` without having to acquire a new lock, meaning the resulting deltas are applied to the original locker.

## Description of changes
- Add a new hook permission flag `ACCESS_LOCK`.
- Update `onlyByLocker` to allow the locker or the allowed hook to call into PoolManager
- Updates some hooks test to account for new flags
- Adds test suite for access look hooks.
- Uses transient storage to save the current hook address on donate, modifyPosition, and swap.

Test file `AccessLock.t.sol` overview:

It tests that a hook with the `ACCESS_LOCK_FLAG` can call every function gated by the `onlyByLocker` modifier. We call each of these functions "LockActions" in the tests.
     *  - Mint
     *  - Take
     *  - Swap
     *  - ModifyPosition
     *  - Donate
We then make sure that each of these actions are callable from the appropriate hook callbacks (beforeModifyPosition, beforeSwap, and beforeDonate).


src/PoolManager.sol
```diff
@@ -174,8 +177,13 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
     }
 
     modifier onlyByLocker() {
+        // todo fix stack too deep :/
         address locker = Lockers.getCurrentLocker();
-        if (msg.sender != locker) revert LockedBy(locker);
+        if (msg.sender != locker) {
+            if (msg.sender != CurrentHookAddress.get() || !Hooks.shouldAccessLock(IHooks(CurrentHookAddress.get()))) {
```
hensha256, line 183: Cache `CurrentHookAddress.get()`?

src/PoolManager.sol
```diff
@@ -49,6 +50,9 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
 
     constructor(uint256 controllerGasLimit) Fees(controllerGasLimit) {}
 
+    // todo change to transient storage
+    address currentHook;
```
hensha256, line 54: maybe it makes sense to remove this until we have the transient keyword - given youre not using it

hensha256: or instead actually i think you could remove the library, and wherever you want to write

```
currentHook = hook
```

just write
```
assembly { tstore(currentHook.slot, hook) }
```

and similarly you can use
```
address hook;
asembly {hook := tload(currentHook.slot)}
```

hensha256: could put those in helper functions for now if you want to but theyre fairly small

src/libraries/Hooks.sol
```diff
@@ -19,8 +19,9 @@ library Hooks {
     uint256 internal constant AFTER_SWAP_FLAG = 1 << 154;
     uint256 internal constant BEFORE_DONATE_FLAG = 1 << 153;
     uint256 internal constant AFTER_DONATE_FLAG = 1 << 152;
+    uint256 internal constant ACCESS_LOCK_FLAG = 1 << 151;
 
-    struct Calls {
+    struct Permissions {
```
hensha256, line 24: nice

src/libraries/Hooks.sol
```diff
@@ -98,4 +103,8 @@ library Hooks {
     function shouldCallAfterDonate(IHooks self) internal pure returns (bool) {
         return uint256(uint160(address(self))) & AFTER_DONATE_FLAG != 0;
     }
+
+    function shouldAccessLock(IHooks self) internal pure returns (bool) {
```
hensha256, line 107: As much as its named in line with the others, I think this name doesnt feel right. Its not that the hook "should" access the lock, its that it "can" access the lock.

I think for this and for NoOp, we should either go with `canAccessLock` and `canNoOp` or `shouldAllowAccessLock` and `shouldAllowNoOp`. I actually think "can" makes the most sense but "should allow" is in line with the shoulds haha

src/test/AccessLockHook.sol
```diff
@@ -0,0 +1,116 @@
...
+
+        return selector;
+    }
+
+    function onERC1155Received(address, address, uint256, uint256, bytes memory) public pure returns (bytes4) {
+        return this.onERC1155Received.selector;
+    }
+}
+
+contract NoAccessLockHook is BaseTestHooks {
```
hensha256, line 99: personally would prefer this in a separate file

hensha256 (commented):
Coming together nicely!

src/test/PoolDonateTest.sol
```diff
@@ -52,13 +55,17 @@ contract PoolDonateTest is PoolTestBase, Test {
...
-        assertEq(reserveBefore0, reserveAfter0);
-        assertEq(reserveBefore1, reserveAfter1);
-        assertEq(deltaAfter0, int256(data.amount0));
-        assertEq(deltaAfter1, int256(data.amount1));
+        if (!data.key.hooks.shouldAccessLock()) {
+            assertEq(reserveBefore0, reserveAfter0);
+            assertEq(reserveBefore1, reserveAfter1);
+            assertEq(deltaAfter0, int256(data.amount0));
+            assertEq(deltaAfter1, int256(data.amount1));
+        }
```
zhongeric, line 63: So we're not checking reserves and deltas for access lock hooks since the hook will be modifying the original sender's lock, are we pushing the assertions into the test file instead?

src/test/PoolSwapTest.sol
```diff
@@ -59,35 +62,47 @@ contract PoolSwapTest is Test, PoolTestBase {
...
-
-        assertEq(reserveBefore0, reserveAfter0);
-        assertEq(reserveBefore1, reserveAfter1);
-
-        if (data.params.zeroForOne) {
-            if (data.params.amountSpecified > 0) {
-                // exact input, 0 for 1
-                assertEq(deltaAfter0, data.params.amountSpecified);
-                assert(deltaAfter1 < 0);
+        if (BalanceDelta.unwrap(delta) == 0) revert NoSwapOccurred(); // Note: Delta actually can end up being 0 even if there was liquidity originally intialized. This happens when liquidity gets pushed entirely into 1 token.
```
zhongeric, line 65: is there an action item on this note?

test/Hooks.t.sol
```diff
@@ -19,22 +19,31 @@ import {Deployers} from "./utils/Deployers.sol";
...
 
     address payable ALL_HOOKS_ADDRESS = payable(0xfF00000000000000000000000000000000000000);
     MockHooks mockHooks;
 
+    // Update this value when you add a new hook flag. And then update all appropriate asserts.
+    uint256 hookPermissionCount = 9;
+    uint256 clearAllHookPermisssionsMask;
+
     function setUp() public {
+        clearAllHookPermisssionsMask = uint256(~uint160(0) >> (hookPermissionCount));
```
zhongeric, line 37: nice great add

test/AccessLock.t.sol
```diff
@@ -0,0 +1,494 @@
...
+        // Make the swap amount small (like a NoOp).
+        donateRouter.donate(key, 1 * 10 ** 18, 1 * 10 ** 18, abi.encode(amount, AccessLockHook.LockAction.Donate));
+
+        uint256 balanceOfAfter0 = MockERC20(Currency.unwrap(currency0)).balanceOf(address(this));
+        uint256 balanceOfAfter1 = MockERC20(Currency.unwrap(currency1)).balanceOf(address(this));
+
+        // Should have less balance in both currencies.
+        assertLt(balanceOfAfter0, balanceOfBefore0);
+        assertLt(balanceOfAfter1, balanceOfBefore1);
+    }
```
marktoda, line 493: what about some edge casier tests like:
- doing one hook operation, then another - the first hook should no longer have access rights but the new one should
- doing one hook operation, then an operation on a pool with _no hook_ - the first hook should no longer have access rights
- .. im sure we can think of more

marktoda (commented):
Looks awesome! Some thoughts -

src/PoolManager.sol
snreynolds: I get stack too deep :/ Any ideas? I left a comment above to trim but I couldn't think of anything

src/libraries/Hooks.sol
snreynolds: Alice and I talked about using `hasPermissionToAccessLock` and `hasPermissionToNoOp`.. Thoughts?

src/test/PoolDonateTest.sol
snreynolds: Yeah the assertions I was testing for AccessLock are after the PoolManager calls 

like here https://github.com/Uniswap/v4-core/blob/access-lock/test/AccessLock.t.sol#L132 

I don't check reserves but I think these assertions make sense? Lmk if you think theres something else I should check though

src/test/PoolSwapTest.sol
snreynolds: Hm no.. I can remove the note and honestly just remove this check though. We already remove it in Alice's NoOp Pr

src/PoolManager.sol
```diff
@@ -185,6 +193,8 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
         IPoolManager.ModifyPositionParams memory params,
         bytes calldata hookData
     ) external override noDelegateCall onlyByLocker returns (BalanceDelta delta) {
+        CurrentHookAddress.set(address(key.hooks));
```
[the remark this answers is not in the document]
snreynolds: Hm I thought it wasn't necessary because you don't need a lock to call initialize, but maybe we should?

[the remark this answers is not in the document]
snreynolds: Hm this just made me realize that there is actually strange behavior (not a security concern bc I think things will revert) but if a hook calls into any of these functions and there is NO current locker then the deltas accrue to the 0 address. Maybe we want to restrict that there is at least one locker? 

Secondly there are two scenarios Im thinking of 

1. If we don't set current hook in initialize and someone calls initialize with no lock -  The hook will have to acquire a lock to call any of the functions.
2. If we don't set current hook in initialize and someone has a lock already but calls initialize.  The hook can't operate on the outside lock.
3. If we do set current hook in initialize and someone calls initialize with no lock.  the hook can operate on the 0 address lock meaning any deltas applied will auto revert.
4. 3. If we do set current hook in initialize and someone calls initialize in their current lock. the hook can operate on the current lock.

Maybe 4 is the most desired? But we need to prevent 3 I think.. unless we're ok with that strange revert behavior

src/PoolManager.sol
```diff
@@ -185,6 +193,8 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
+
```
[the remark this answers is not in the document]
snreynolds: https://github.com/Uniswap/v4-core/issues/415

test/Hooks.t.sol
```diff
@@ -528,15 +596,16 @@ contract HooksTest is Test, Deployers, GasSnapshot {
...
                 beforeInitialize: true,
                 afterInitialize: true,
                 beforeModifyPosition: true,
                 afterModifyPosition: true,
                 beforeSwap: true,
                 afterSwap: true,
                 beforeDonate: true,
-                afterDonate: true
+                afterDonate: true,
+                accessLock: false
```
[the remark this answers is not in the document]
snreynolds: oh ya.. good catch

src/PoolManager.sol
```diff
@@ -174,8 +177,13 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
     }
 
     modifier onlyByLocker() {
+        // todo fix stack too deep :/
         address locker = Lockers.getCurrentLocker();
-        if (msg.sender != locker) revert LockedBy(locker);
+        if (msg.sender != locker) {
+            if (msg.sender != CurrentHookAddress.get() || !Hooks.shouldAccessLock(IHooks(CurrentHookAddress.get()))) {
+                revert LockedBy(locker);
```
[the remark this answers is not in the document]
snreynolds: can do.. but also lmk if you have any ideas for stack too deep so I can cache the hookAddress :P

test/AccessLock.t.sol
snreynolds: I'll add a hook accessing a lock and applying deltas when there is no outside locker

src/PoolManager.sol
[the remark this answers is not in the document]
snreynolds: Actually, this is a critical bug.

In the current code, a hook that has an access lock permission that has successfully set the LAST currentHook address to itself can call modifyPosition/swap/donate/mint/take successfully without settling deltas because it can call these functions OUTSIDE the lock. 

I've reproduced the vulnerability with this test here.

I think there are two things we can do
a) require there must be a lock ie in onlyByLocker we can check that if msg.sender is not the current locker and if the current locker == addr(0) we revert

b) at the end of a lock clear the currentHook address.

My gut is to actually do both to be safe but I think a) alone would solve the issue.

B) by itself I think actually does not solve the bug if we were to add `_setCurrentLocker` in initialize. This is because we would have set the current hook OUTSIDE of a lock so we would never actually clear the current hook because that path would never be hit.

I also wonder if any of this warrants `initialize` needing a lock?

src/PoolManager.sol
snreynolds: btw was able to fix stack too deep by doing this

https://github.com/Uniswap/v4-core/compare/access-lock...stack-too-deep?expand=1 

but it adds gas :/

src/PoolManager.sol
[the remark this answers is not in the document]
hensha256: Really nice find sara! Definitely agree that (a) is the way to go.

And I think I agree that we should add a lock to `initialize`, and do the exact same currentHook flow for that one too.

If we think that hooks might want to do things in before/afterInitialize (which seems like a reasonable assumption)... it feels right that those actions might add deltas to pools and we should add locking too.

src/PoolManager.sol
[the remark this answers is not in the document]
hensha256: > Hm actually hesistant bc we might log a hook address that doesn't have access lock permissions which is confusing?

arguably its help - maybe the hook not having permissions is the problem as to why someone is getting a revert theyre confused about. I think no harm in logging it!

src/PoolManager.sol
```diff
@@ -175,7 +176,14 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
 
     modifier onlyByLocker() {
         address locker = Lockers.getCurrentLocker();
-        if (msg.sender != locker) revert LockedBy(locker);
+        if (msg.sender != locker) {
+            if (
+                locker == address(0) || msg.sender != CurrentHookAddress.get()
+                    || !Hooks.hasPermissionToAccessLock(IHooks(CurrentHookAddress.get()))
```
hensha256, line 182: yessss this function name reads so clearly love it

src/PoolManager.sol
```diff
@@ -243,6 +253,8 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
         onlyByLocker
         returns (BalanceDelta delta)
     {
+        CurrentHookAddress.set(address(key.hooks));
```
hensha256, line 256: I do think we should be clearing this after each call? It feels wrong that at the end of a call to `swap`, the control passes back to the locker, and `CurrentHookAddress` is still set to some hook that is _not_ the current hook?

This contract does pass control out to other contracts at times (like the locker, tokens), so i just want to lower the risk of when contracts can take control of the lock again?

src/libraries/CurrentHookAddress.sol
```diff
@@ -0,0 +1,23 @@
...
+        assembly {
+            _address := tload(slot)
+        }
+    }
+
+    function set(address _address) internal {
+        if (_address != get()) {
+            uint256 slot = HOOK_ADDRESS_SLOT;
+            assembly {
+                tstore(slot, _address)
```
hensha256, line 19: they really should to support accessing constants in assembly 🙄

src/test/AccessLockHook.sol
```diff
@@ -0,0 +1,142 @@
...
+                new bytes(0)
+            );
+        } else {
+            revert("Invalid action");
+        }
+
+        return selector;
+    }
+
+    function onERC1155Received(address, address, uint256, uint256, bytes memory) public pure returns (bytes4) {
```
hensha256, line 95: Is this a remnant from 1155 days? Can remove now?

src/test/AccessLockHook.sol
```diff
@@ -0,0 +1,142 @@
...
+            if (manager.getCurrentHook() != address(key2.hooks)) {
+                revert IncorrectHookSet();
+            }
+            // Should revert since currentHook is the other pools hook
+            manager.mint(key.currency1, address(this), 10);
+        }
+        return IHooks.beforeModifyPosition.selector;
+    }
+
+    function onERC1155Received(address, address, uint256, uint256, bytes memory) public pure returns (bytes4) {
```
hensha256, line 139: same here

src/libraries/Lockers.sol
```diff
@@ -8,6 +8,9 @@ library Lockers {
     // The starting slot for an array of lockers, stored transiently.
     uint256 constant LOCKERS_SLOT = uint256(keccak256("Lockers")) - 1;
```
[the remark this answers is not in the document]
hensha256: e.g. i'd like to be sure that:
- you start with a nested lock, length 2
- the second locker gets a `currentHook` of X
- second locker releases (and is popped, length = 1)
- a new second locker takes out a lock (length = 2) and in that moment `currentHook == address(0)` and there isnt some remnant currentHook from before the decrement of `length`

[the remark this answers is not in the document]
snreynolds: Interesting.. so it is only true OUTSIDE of the library. (ie because we manually clear it any time a hookAddr is pushed, not because the library enforces this). 

This means we can only test that this holds true outside of the library unit tests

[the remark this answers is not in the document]
snreynolds: added this test here! https://github.com/Uniswap/v4-core/pull/404/commits/f70443aabe3308cadcd6130b79b0a4da827fef06

src/PoolManager.sol
```diff
@@ -175,25 +179,38 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
     }
 
     modifier onlyByLocker() {
-        address locker = Lockers.getCurrentLocker();
-        if (msg.sender != locker) revert LockedBy(locker);
+        _checkLocker(msg.sender, Lockers.getCurrentLocker(), Lockers.getCurrentHook());
         _;
     }
 
+    function _checkLocker(address caller, address locker, IHooks hook) internal pure {
```
marktoda, line 186: since the modifier is the only calsite, why have as a separate internal fn?

src/PoolManager.sol
```diff
@@ -175,25 +179,38 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
     }
 
     modifier onlyByLocker() {
```
marktoda, line 181: now that the complexity of this modifier has increased, its probably worth having natspec for it

src/PoolManager.sol
```diff
@@ -175,25 +179,38 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
+        _checkLocker(msg.sender, Lockers.getCurrentLocker(), Lockers.getCurrentHook());
         _;
     }
 
+    function _checkLocker(address caller, address locker, IHooks hook) internal pure {
+        if (caller != locker) {
+            if (locker == address(0) || caller != address(hook) || !hook.hasPermissionToAccessLock()) {
+                revert LockedBy(locker, address(hook));
+            }
+        }
```
marktoda, line 191: nit: can flatten this for readability, maybe like:
```
if (caller == locker) return;
if (caller = hook && hook.hasPermissionToAccessLock()) return;
revert
```

src/PoolManager.sol
```diff
@@ -175,25 +179,38 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
     modifier onlyByLocker() {
-        address locker = Lockers.getCurrentLocker();
-        if (msg.sender != locker) revert LockedBy(locker);
+        _checkLocker(msg.sender, Lockers.getCurrentLocker(), Lockers.getCurrentHook());
         _;
     }
 
+    function _checkLocker(address caller, address locker, IHooks hook) internal pure {
+        if (caller != locker) {
+            if (locker == address(0) || caller != address(hook) || !hook.hasPermissionToAccessLock()) {
```
marktoda, line 188: under what circumstance is locker == address(0)?

marktoda (commented):
Nice, looking good!

src/PoolManager.sol
snreynolds: stack too deep :/ I could remove it but then I can't save msg.sender or Lockers.getCurrentLocker or Lockers.getCurrentHook

src/PoolManager.sol
```diff
@@ -105,6 +105,8 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
         if (key.currency0 >= key.currency1) revert CurrenciesOutOfOrderOrEqual();
         if (!key.hooks.isValidHookAddress(key.fee)) revert Hooks.HookAddressNotValid(address(key.hooks));
 
+        (bool set) = Lockers.setCurrentHook(key.hooks);
```
hensha256, line 108: should this be inside the `if` statement?
```
bool set;
if (key.hooks.shouldCallBeforeInitialize()) {
    set = Lockers.setCurrentHook(key.hooks);
    if (key.hooks.beforeInitialize ... {}
}
```

hensha256: ohhh i guess then youd have to do it inside both hooks if statements
just feels weird to set it if we arent even calling for a hook

src/PoolManager.sol
```diff
@@ -174,26 +178,38 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
         if (pools[id].isNotInitialized()) revert PoolNotInitialized();
     }
 
+    /// @notice This will revert if a function is called by any address other than the current locker OR the most recently called, pre-permissioned hook.
     modifier onlyByLocker() {
```
hensha256, line 182: nit: now that `onlyByLocker` is on `initialize` I think `onlyByLocker` and `_checkLocker` should be higher in the contact

src/PoolManager.sol
```diff
@@ -233,6 +249,9 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
             }
         }
 
+        // We only want to clear the current hook if it is the first time setting the hook address.
```
hensha256, line 252:
```suggestion
        // We only want to clear the current hook if it was set in setCurrentHook in this execution frame
```

hensha256: and maybe add the same comment every time we call this function

src/libraries/Lockers.sol
```diff
@@ -94,4 +99,35 @@ library Lockers {
...
+        if (address(getCurrentHook()) == address(0)) {
+            uint256 slot = HOOK_ADDRESS_SLOT + length();
+            assembly {
+                tstore(slot, currentHook)
+            }
+            return true;
+        }
+    }
+
+    // We only want to clear the current hook if the msg.sender is the original locker.
```
hensha256, line 126: this comment feels odd because this check doesn't happen here? Is this a warning to devs?

snreynolds: Oh yeah old comment I think

src/PoolManager.sol
snreynolds: I think we can make this change if we want in the hook callsites refactor

src/test/AccessLockHook.sol
```diff
@@ -0,0 +1,196 @@
...
+
+    function beforeModifyPosition(
+        address sender,
+        PoolKey calldata key,
+        IPoolManager.ModifyPositionParams calldata params,
+        bytes calldata hookData
+    ) external override returns (bytes4) {
+        if (address(manager.getCurrentHook()) != address(this)) {
+            revert IncorrectHookSet();
+        }
```
zhongeric, line 130: This seems like its testing a well behaved acess lock hook. Should we add a test hook that actively tries to access locks even though it's not the currentHook? There might already be tests for this that I might be missing too

src/test/PoolModifyPositionTest.sol
```diff
@@ -11,6 +11,8 @@ import {Hooks} from "../libraries/Hooks.sol";
 import {Test} from "forge-std/Test.sol";
 import {FeeLibrary} from "../libraries/FeeLibrary.sol";
 
+import "forge-std/console2.sol";
```
zhongeric, line 14: nit: console2

src/PoolManager.sol
```diff
@@ -233,6 +251,9 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
```
hensha256, line 254: nit: you updated this comment everywhere else except this line
```
// We only want to clear the current hook if it was set in setCurrentHook in this execution frame.
```

test/AccessLock.t.sol
```diff
@@ -0,0 +1,726 @@
...
+     *
+     * The following test suite tests that appropriate hooks can call
+     *  every function gated by the `onlyByLocker` modifier.
+     *  We call these "LockActions".
+     *  LockActions:
+     *  - Mint
+     *  - Take
+     *  - Swap
+     *  - ModifyPosition
+     *  - Donate
```
zhongeric, line 105: why is Burn / Settle not tested? Is it because they are purely returning money to core?

src/test/AccessLockHook.sol
```diff
@@ -0,0 +1,196 @@
...
+            manager.modifyPosition(
+                key,
+                IPoolManager.ModifyPositionParams({tickLower: -60, tickUpper: 60, liquidityDelta: int256(amount)}),
+                new bytes(0)
+            );
+        } else if (action == LockAction.NoOp) {
+            assertEq(address(manager.getCurrentHook()), address(this));
+            return Hooks.NO_OP_SELECTOR;
+        } else {
+            revert("Invalid action");
```
hensha256, line 104: nit: custom error

test/AccessLock.t.sol
```diff
@@ -0,0 +1,726 @@
...
+        modifyPositionRouter.modifyPosition(
+            key,
+            IPoolManager.ModifyPositionParams({tickLower: -120, tickUpper: 120, liquidityDelta: 100 * 10e18}),
+            ZERO_BYTES
+        );
+
+        uint256 balanceOfBefore1 = MockERC20(Currency.unwrap(currency1)).balanceOf(address(this));
+        uint256 balanceOfBefore0 = MockERC20(Currency.unwrap(currency0)).balanceOf(address(this));
+
+        // Just no-op.
```
zhongeric, line 179: nit: comment could be more descriptive since it's hard to know what logic is run on LockAction.Swap, it was a bit hard to understand what this test was doing. Like `// No-op the modifyPosition call and swap before hand, applying the deltas to the locker`

test/AccessLock.t.sol
```diff
@@ -0,0 +1,726 @@
...
+            ZERO_BYTES
+        );
+
+        uint256 balanceOfBefore1 = MockERC20(Currency.unwrap(currency1)).balanceOf(address(this));
+        uint256 balanceOfBefore0 = MockERC20(Currency.unwrap(currency0)).balanceOf(address(this));
+
+        // Small amount to swap (like NoOp). This way we can expect balances to just be from the hook applied delta.
+        BalanceDelta delta = swapRouter.swap(
+            key,
+            IPoolManager.SwapParams(true, 1, TickMath.MIN_SQRT_RATIO + 1),
```
zhongeric, line 258: nit: why not NoOp the swap?

test/AccessLock.t.sol
```diff
@@ -0,0 +1,726 @@
...
+        vm.prank(address(key.hooks));
+        manager.modifyPosition(key, IPoolManager.ModifyPositionParams(0, 60, 1 * 10 ** 18), ZERO_BYTES);
+    }
+
+    function test_getCurrentHook_isClearedAfterNestedLock() public {
+        // Create pool for AccessLockHook3.
+        (PoolKey memory keyAccessLockHook3,) =
+            initPool(currency0, currency1, IHooks(accessLockHook3), Constants.FEE_MEDIUM, SQRT_RATIO_1_1, ZERO_BYTES);
+        // Fund AccessLockHook3 with currency0.
+        MockERC20(Currency.unwrap(currency0)).transfer(address(accessLockHook3), 10);
```
zhongeric, line 686: should we do a larger number than 10 for precision purposes?

zhongeric (commented):
looks good just some questions!

src/PoolManager.sol
```diff
@@ -298,14 +329,21 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
         _checkPoolInitialized(id);
 
         if (key.hooks.shouldCallBeforeDonate()) {
             bytes4 selector = key.hooks.beforeDonate(msg.sender, key, amount0, amount1, hookData);
             // Sentinel return value used to signify that a NoOp occurred.
-            if (key.hooks.isValidNoOpCall(selector)) return BalanceDeltaLibrary.MAXIMUM_DELTA;
-            else if (selector != IHooks.beforeDonate.selector) revert Hooks.InvalidHookResponse();
+            if (key.hooks.isValidNoOpCall(selector)) {
+                // We only want to clear the current hook if it was set in setCurrentHook in this execution frame.
+                if (set) Lockers.clearCurrentHook();
```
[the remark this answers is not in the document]
zhongeric: Like, if for some reason we add a new return path for these methods we have to copy this code to all of those as well.

test/CurrentHookAddress.t.sol
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    function test_setCurrentHook() public {
+        Lockers.setCurrentHook(IHooks(address(1)));
+        assertEq(address(Lockers.getCurrentHook()), address(1));
+    }
+
+    function test_setCurrentHook_TwiceDoesNotSucceed() public {
+        Lockers.setCurrentHook(IHooks(address(1)));
+        Lockers.setCurrentHook(IHooks(address(2)));
```
hensha256, line 20: i think this should be checking the return values?
like if it changed -> true, if it didnt change -> false

src/PoolManager.sol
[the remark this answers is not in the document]
snreynolds: Yep Mark and I talked about this on the hook refactor PR to take this into consideration. It would be nice to try to compile one return path.

src/libraries/Hooks.sol
```diff
@@ -44,17 +46,19 @@ library Hooks {
...
-                || calls.beforeDonate != shouldCallBeforeDonate(self) || calls.afterDonate != shouldCallAfterDonate(self)
-                || calls.noOp != hasPermissionToNoOp(self)
+            permissions.beforeInitialize != shouldCallBeforeInitialize(self)
+                || permissions.afterInitialize != shouldCallAfterInitialize(self)
+                || permissions.beforeModifyPosition != shouldCallBeforeModifyPosition(self)
+                || permissions.afterModifyPosition != shouldCallAfterModifyPosition(self)
+                || permissions.beforeSwap != shouldCallBeforeSwap(self)
+                || permissions.afterSwap != shouldCallAfterSwap(self)
+                || permissions.beforeDonate != shouldCallBeforeDonate(self)
+                || permissions.afterDonate != shouldCallAfterDonate(self) || permissions.noOp != hasPermissionToNoOp(self)
```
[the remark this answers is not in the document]
snreynolds: auto formatted like this :/

test/AccessLock.t.sol
snreynolds: Ah yeah I can add tests for those

test/AccessLock.t.sol
snreynolds: oo I can change to use NoOp now that its merged!

snreynolds: Hm actually i'm going to leave as is 0 would prefer these tests don't go through the no-op code

hensha256 (dismissed):

marktoda (dismissed):

marktoda (approved):

PR #410, opened 2023-11-17
Added delta overflow checks

aadams:
## Related Issue

When fuzzing modify position, sometimes we would get too large of a modify position and it would overflow the cast to int128. Because this was dependent on tick-spacing, ticks, and liquidity, it was hard to track down


snreynolds (dismissed):
nice

snreynolds (approved):

PR #411, opened 2023-11-17
feat: hooks refactor POC

marktoda:
The per-hook block is increasing in size over time and now is roughly
8 lines of code duplicated 8 times after #404 and #324. This commit
  attempts to refactor some of this copied logic into the Hooks library
  to improve readability of the core PoolManager code while also
  removing duplicated code

This is a proof of concept, I'll wait until the above PRs are merged to
include them in this model and can improve testing a lot with this new
approach as well

## Related Issue

## Description of changes


marktoda:
oof gas quite a bit worse due to extra jumps, will think about how to improve that

src/libraries/Hooks.sol
```diff
@@ -67,35 +73,79 @@ library Hooks {
             );
     }
 
-    function shouldCallBeforeInitialize(IHooks self) internal pure returns (bool) {
-        return uint256(uint160(address(self))) & BEFORE_INITIALIZE_FLAG != 0;
+    function validateHooksResponse(bytes4 selector, bytes4 expectedSelector) internal pure {
+        if (selector != expectedSelector) {
+            revert InvalidHookResponse();
+        }
```
marktoda, line 79: to add: noop handling here

src/libraries/Hooks.sol
```diff
@@ -67,35 +73,79 @@ library Hooks {
...
     }
 
-    function shouldCallBeforeModifyPosition(IHooks self) internal pure returns (bool) {
-        return uint256(uint160(address(self))) & BEFORE_MODIFY_POSITION_FLAG != 0;
+    function beforeModifyPosition(
+        PoolKey memory key,
+        IPoolManager.ModifyPositionParams memory params,
+        bytes memory hookData
+    ) internal {
+        if (!key.hooks.shouldCall(BEFORE_MODIFY_POSITION_FLAG)) return;
```
marktoda, line 105: to add: CA handling

src/libraries/Hooks.sol
snreynolds: this refactor will be strange with NoOp because we have to run the clear currentHook logic before returning

snreynolds:
One comment.. if we can find a nice way to use a helper for the callsite calls, I think it would be neater to do the `if (set) Lockers.clearCurrentHook()` check INSIDE the hook checks so that we aren't setting a current hook in the no hook case.

PR #412, opened 2023-11-17
chore: add fee collection signatures to the IFees interface

saucepoint:
## Related Issue
`IFees.sol` interface did not match the available signatures defined in `Fees.sol` (inherited by PoolManager)

## Description of changes

Added the `collectHookFees` and `collectProtocolFees` to the interface

---

Note: I'm happy to update the `Fees.t.sol` to use `IPoolManager` instead, so that we can catch missing signatures in the future

It would require some changes with how the tests are unpacking slot0:
```solidity
(Pool.Slot0 memory slot0,,,) = manager.pools(key0.toId());

changed to:

// now that manager is IPoolManager:
(uint160 sqrtPriceX96, int24 tick, uint24 protocolFees, uint24 hookFees) = manager.getSlot0(key0.toId());
Pool.Slot0 memory slot0 = Pool.Slot0(sqrtPriceX96, tick, protocolFees, hookFees);

```

### DIFF
--- a/src/interfaces/IFees.sol
+++ b/src/interfaces/IFees.sol
@@ -21,4 +21,14 @@ interface IFees {
 
     /// @notice Given a hook and a currency address, returns the fees accrued
     function hookFeesAccrued(address, Currency) external view returns (uint256);
+
+    /// @notice Collect the protocol accrued fees, sent to a recipient
+    function collectProtocolFees(address recipient, Currency currency, uint256 amount)
+        external
+        returns (uint256 amountCollected);
+
+    /// @notice Collect a hook's accrued fees, sent to a recipient
+    function collectHookFees(address recipient, Currency currency, uint256 amount)
+        external
+        returns (uint256 amountCollected);
 }


zhongeric (approved):
Do you mind making an issue for moving tests over to using `IPoolManager` in the future? This PR looks good to me though!

snreynolds:
Hey! I'd actually say that we don't want to merge this atm. We are actively working on changing how protocol fees are accrued. IE this PR actually puts it into PoolManager  https://github.com/Uniswap/v4-core/pull/392/files#diff-acdce9da6003b3fe3f557e3a9a32c4cc13c484a8abdd96429721098b152cf3b4

While we probably won't go with that exact PR, we do want protocolFees to accrue in accordance with our internal accounting (either with MinimalClaims or 1155-like or 6909-like) and so I think we should not merge this while that is still in development.

snreynolds:
If we wanted collectProtocolFees to exist in the IFees interface that means the Fees contract would need to be aware of the MinimalBalance/ERC1155/ERC6909-like contract and so we might have to do some weird abstract wrappers or something but I'm thinking we should just make an open issue to consolidate the fees interface/functions after/while we make those changes to underlying protocol fee collection logic

saucepoint:
makes sense, happy to hold off until everything shakes out

re: the draft PR and collectProtocolFees, I think adding it directly to `IPoolManager` makes sense if the function is defined in PoolManager. because collectHookFees is defined in Fees, it makes sense to put it in IFees imo

snreynolds:
> makes sense, happy to hold off until everything shakes out
> 
> re: the draft PR and collectProtocolFees, I think adding it directly to `IPoolManager` makes sense if the function is defined in PoolManager. because collectHookFees is defined in Fees, it makes sense to put it in IFees imo

Yep the thinking totally makes sense!!

PR #417, opened 2023-11-20
remove forge snapshot --check

snreynolds:
## Related Issue

It is with a heavy heart I wish to sunset snapshot checking. Tolerance already had to be increased to 25%+ so I don't think its that useful.

giphy


.forge-snapshots/simple swap native.snap
```diff
@@ -1 +0,0 @@
-132772
```
hensha256, line 1: why is this one going?

marktoda (approved):

.forge-snapshots/simple swap native.snap
snreynolds: I ran rm -rf .forge-snapshots and then re-ran snapshots so it probably just didn't get deleted from a different PR

snreynolds: seems like it is called "simple swap with native" now

PR #418, opened 2023-11-20
feat: update justfile with custom solc

marktoda:
For folks who dont want to update their global env they can use `just
test` or `just build` which sets solc using cli arg

## Related Issue

## Description of changes

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,15 +1,13 @@
+solc_file := if os() == "macos" { "./bin/solc-mac" } else { "./bin/solc-static-linux" }
+
 test: test-forge
 prep: fix snapshots
-snapshots: snapshots-forge
 
 test-forge: install-forge build-forge
-    forge test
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -1,15 +1,17 @@
...
 
 test-forge: install-forge build-forge
-    forge test
+    forge test --use {{ solc_file }}
 
 build-forge: install-forge
-    forge build
+    forge build --use {{ solc_file }}
 
 snapshots-forge: install-forge test-forge
```
snreynolds, line 13: we can remove this?

snreynolds (changes requested):

justfile
snreynolds: just while we're in here modifying thingsss

snreynolds (approved):

PR #358, opened 2023-09-24
Replace ERC-1155 With ERC-6909 For Multi-Token Accounting

jtriley-eth:
Closes #357 

## Description of changes

- Replace ERC-1155 with ERC-6909
- Replace self-transfer and callback based burn mechanism with burn function
- Update tests
- Remove ERC11155Receiver interface

## Gas Difference

Computed as `cost(ERC6909) - cost(ERC1155)`.

| deploy cost | codesize |
| ----------- | -------- |
| -799_822    | -4395    |

| fn name                        | min     | avg     | med     | max     |
| ------------------------------ | ------- | ------- | ------- | ------- |
| MAX_TICK_SPACING               | +22     | +22     | +22     | +22     |
| balanceOf                      | -78     | -78     | -78     | -78     |
| [burn*](#burn)                 | -22697  | -22697  | -22697  | -22697  |
| collectHookFees                | 0       | 0       | 0       | 0       |
| collectProtocolFees            | -34     | -34     | -34     | -34     |
| currencyDelta                  | 0       | 0       | 0       | 0       |
| donate                         | 0       | 0       | 0       | 0       |
| getLock                        | +22     | +22     | +22     | +22     |
| getPosition                    | 0       | 0       | 0       | 0       |
| getSlot0                       | -23     | -23     | -23     | -23     |
| initialize                     | +7       | -35     | -45     | +53      |
| lock                           | -79     | +353     | -2692   | -461    |
| lockData                       | -22     | -22     | -22     | -22     |
| mint                           | +15_601 | +15_601 | +15_601 | +15_601 |
| modifyPosition                 | +20     | +754    | +20     | +5535   |
| pools                          | 0       | 0       | 0       | 0       |
| protocolFeeController          | 0       | 0       | 0       | 0       |
| protocolFeesAccrued            | +22     | +22     | +22     | +22     |
| [transferFrom*](#transferfrom) | -26_675 | -12_609 | -8_374  | -5_935  |
| [setOperator*](#setoperator)    | -33     | -33     | -33     | -33     |
| setHookFees                    | 0       | +254    | 0       | 0       |
| setProtocolFeeController       | 0       | 0       | 0       | 0       |
| setProtocolFees                | 0       | -66     | 0       | 0       |
| settle                         | +17     | +17     | +17     | +22     |
| swap                           | +21     | +21     | +16     | +21     |
| take                           | -35     | -35     | -35     | -35     |

### burn

[...]

PR #422, opened 2023-11-21
Implement ERC6909 for balance accounting

zhongeric:
## Related Issue
We've decided to merge in ERC6909 support for now to unblock integrators and further development on core. However, we have certain optimizations that we want to make to the underlying spec, and are working closely with @jtriley-eth and others. 

We hope to move towards a token standard in the future for these types of user balances so this PR and any changes made in it should not be considered final and will very likely be changed in the future.

## Description of changes
- Uses #358 , removes claims logic
- Uses current implementation of ERC6909 in solmate with totalSupply removed
- Adds some gas tests


src/PoolManager.sol
```diff
@@ -17,13 +17,13 @@ import {IHookFeeManager} from "./interfaces/IHookFeeManager.sol";
 import {IPoolManager} from "./interfaces/IPoolManager.sol";
 import {ILockCallback} from "./interfaces/callback/ILockCallback.sol";
 import {Fees} from "./Fees.sol";
-import {Claims} from "./Claims.sol";
+import {ERC6909} from "solmate/tokens/ERC6909.sol";
```
hensha256, line 20: I'd rather we copy and paste this contract in, and rewrite the `_mint` and `_burn` functions to not use `totalSupply`, end remove the `totalSupply` mapping entirely. It feels wrong that we still have that mapping but its always 0. And I think it'll be neater if the PoolManager can just call `_mint` and _`burn` instead of having the logic locally

src/PoolManager.sol
```diff
@@ -320,24 +320,37 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
+        }
+        balanceOf[from][id] -= amount;
+        emit Transfer(msg.sender, from, address(0), id, amount);
     }
 
     /// @inheritdoc IPoolManager
-    function burn(Currency currency, uint256 amount) external noDelegateCall onlyByLocker {
-        _accountDelta(currency, -(amount.toInt128()));
-        _burn(currency, amount);
+    function settle(Currency currency) external payable override noDelegateCall onlyByLocker returns (uint256 paid) {
```
hensha256, line 348: can you put `settle` back up before `mint` and `burn` please? Unless theres a reason to change the order? The diff is harder to review

test/PoolManager.t.sol
```diff
@@ -399,50 +400,53 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
-        // swap from currency1 to currency0 again, using Claims as input tokens
+        // swap from currency1 to currency0 again, using 6909s as input tokens
         params = IPoolManager.SwapParams({zeroForOne: false, amountSpecified: -25, sqrtPriceLimitX96: SQRT_RATIO_4_1});
 
         testSettings = PoolSwapTest.TestSettings({withdrawTokens: true, settleUsingTransfer: false});
 
         vm.expectEmit(true, true, true, false);
-        emit Burn(address(swapRouter), currency1, 27);
-        snapStart("swap burn claim for input");
+        emit Transfer(address(swapRouter), address(this), address(0), CurrencyLibrary.toId(currency1), 27);
```
hensha256, line 443: mmm maybe we should ask the EIP to change this to Mint and Burn. Its such a waste of gas of burn to always emit address(0), and same with burn

hensha256: thoughts?

hensha256 (changes requested):

src/PoolManager.sol
hensha256: oh i see @jtriley-eth is doing a PR to do this within solmate itself... i guess we will hold this PR til thats done

snreynolds: Kind of agree. So either we hold this PR until its completed in solmate or copy/paste/credit but modify totalSupply and then later use the solmate package when its ready. Kind of in favor of the latter so we can unblock other PRs for core.

test/PoolManager.t.sol
snreynolds: I agree this is wasteful :/

test/Claims.t.sol
```diff
@@ -1,99 +0,0 @@
...
-pragma solidity ^0.8.20;
-
-import {Test} from "forge-std/Test.sol";
-import {Claims} from "../src/Claims.sol";
-import {IClaims} from "../src/interfaces/IClaims.sol";
-import {CurrencyLibrary, Currency} from "../src/types/Currency.sol";
-import {MockClaims} from "../src/test/MockClaims.sol";
-import {Deployers} from "./utils/Deployers.sol";
-
-contract ClaimsTest is Test, Deployers {
```
snreynolds, line 11: Can we keep some of these same tests/modify them for 6909?

snreynolds (changes requested):

test/Claims.t.sol
zhongeric: We could if we're rolling our own implementation but I don't really think there's a need to have these tests if they're already in the 6909 repo?

src/PoolManager.sol
zhongeric: Sure I can copy and paste the solmate impl into here

src/PoolManager.sol
zhongeric: Sure!

test/PoolManager.t.sol
zhongeric: Resolved offline that we're ok with Transfer

test/Claims.t.sol
zhongeric: Using solmate's impl

src/PoolManager.sol
zhongeric: Using solmate impl

src/PoolManager.sol
```diff
@@ -336,15 +336,22 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
-    function mint(Currency currency, address to, uint256 amount) external noDelegateCall onlyByLocker {
+    function mint(Currency currency, address to, uint256 amount) external override noDelegateCall onlyByLocker {
         _accountDelta(currency, amount.toInt128());
-        _mint(to, currency, amount);
+        _mint(to, currency.toId(), amount);
     }
 
     /// @inheritdoc IPoolManager
-    function burn(Currency currency, uint256 amount) external noDelegateCall onlyByLocker {
+    function burn(Currency currency, address from, uint256 amount) external override noDelegateCall onlyByLocker {
```
marktoda, line 345: now thaat we are 6909, wonder if we should go back to `uint256 tokenId` as the first param here? Bit weird to be able to do `mint(currency, to, amount)` but not be able to do `transfer(currency, to, amount)`

marktoda (commented):
Nice, EZ and looks great! 

Seems we probably need some tests for standard 6909 functionality as well though, to ensure no interactions and functions as expected. Also good to have full core functionality tested locally to ensure i.e. upstream changes to solmate don't affect our understandings and invariants in core. i.e. claim transfers, operators, approvals etc

src/PoolManager.sol
zhongeric: Yea I think it is a nice benefit that all funcs on core use Currency

zhongeric: Oh hmm, dealing with merge conflicts and it definitely is annoying that `balanceOf` requires uint while mint/burn use Currency.

src/PoolManager.sol
```diff
@@ -377,15 +377,23 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
         _accountDelta(currency, -(amount.toInt128()));
-        _burn(currency, amount);
+        uint256 id = currency.toId();
+        address sender = msg.sender;
+        if (from != sender && !isOperator[from][sender]) {
```
hensha256, line 390: i think id rather this be in a helper contract - feels weird that PM is having to deal with allowances and operators when thats a token-specific thing
maybe we could have an intermediate contract like `V4Core6909` or something that is `6909` but adds `burnFrom` logic?

src/PoolManager.sol
hensha256: i kind of love a bit of overloading lol, but i know we're keeping v4 core minimal

snreynolds: I do think an ordering change would be nice though. So (receiver, id/currency, amount) to mirror 6909?

snreynolds (commented):
Looks great. I think also just updating CurrencyLib so that we have balanceOf and transfer with native Currency type for underlying erc20 and 6909 makes sense. But lmk thoughts.

src/PoolManager.sol
zhongeric: Yea agree, was kinda thinking that too. Will implement

src/V46909.sol
```diff
@@ -0,0 +1,21 @@
...
+pragma solidity ^0.8.19;
+
+import {ERC6909} from "solmate/tokens/ERC6909.sol";
+
+contract V46909 is ERC6909 {
+    /// @notice Burn `amount` tokens of token type `id` from `from`.
+    /// @param from The address to burn tokens from.
+    /// @param id The currency to burn.
+    /// @param amount The amount to burn.
+    function burnFrom(address from, uint256 id, uint256 amount) public {
```
zhongeric, line 11: internal

src/PoolManager.sol
zhongeric: discussed offline: reordered parameters and using Currency for mint/burn + uint256 for other 6909 specific functions

test/PoolManager.t.sol
```diff
@@ -500,52 +501,103 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
 
-    function test_swap_GasUseClaimAsInput() public {
+    function test_swap_GasMint6909IfNativeOutputNotTaken() public {
+        IPoolManager.SwapParams memory params =
+            IPoolManager.SwapParams({zeroForOne: false, amountSpecified: 100, sqrtPriceLimitX96: SQRT_RATIO_2_1});
+
+        PoolSwapTest.TestSettings memory testSettings =
+            PoolSwapTest.TestSettings({withdrawTokens: false, settleUsingTransfer: true, currencyAlreadySent: false});
+
+        vm.expectEmit(true, true, true, false);
```
snreynolds, line 528: this should be vm.expectEmit() which does true for everything

test/PoolManager.t.sol
```diff
@@ -500,52 +501,103 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
+        // give permission for swapRouter to burn the 6909s
+        manager.setOperator(address(swapRouter), true);
 
+        // swap from currency1 to currency0 again, using 6909s as input tokens
+        params = IPoolManager.SwapParams({zeroForOne: false, amountSpecified: -25, sqrtPriceLimitX96: SQRT_RATIO_4_1});
         testSettings =
             PoolSwapTest.TestSettings({withdrawTokens: true, settleUsingTransfer: false, currencyAlreadySent: false});
-        manager.transfer(address(swapRouter), currency1, claimsBalance);
 
         vm.expectEmit(true, true, true, false);
```
snreynolds, line 560: same here I think we want vm.expectEmit()

test/PoolManager.t.sol
```diff
@@ -500,52 +501,103 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
+    }
+
+    function test_swap_GasUse6909Input() public {
         IPoolManager.SwapParams memory params =
             IPoolManager.SwapParams({zeroForOne: true, amountSpecified: 100, sqrtPriceLimitX96: SQRT_RATIO_1_2});
 
         PoolSwapTest.TestSettings memory testSettings =
             PoolSwapTest.TestSettings({withdrawTokens: false, settleUsingTransfer: true, currencyAlreadySent: false});
 
         vm.expectEmit(true, true, true, false);
```
snreynolds, line 545: same

test/PoolManager.t.sol
```diff
@@ -500,52 +501,103 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
+    }
+
+    function test_swap_GasUseNative6909Input() public {
+        IPoolManager.SwapParams memory params =
+            IPoolManager.SwapParams({zeroForOne: false, amountSpecified: 100, sqrtPriceLimitX96: SQRT_RATIO_2_1});
+
+        PoolSwapTest.TestSettings memory testSettings =
+            PoolSwapTest.TestSettings({withdrawTokens: false, settleUsingTransfer: true, currencyAlreadySent: false});
+
+        vm.expectEmit(true, true, true, false);
```
snreynolds, line 577: same

test/PoolManager.t.sol
```diff
@@ -500,52 +501,103 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
+
+        // give permission for swapRouter to burn the 6909s
+        manager.setOperator(address(swapRouter), true);
+
+        // swap from currency0 to currency1, using 6909s as input tokens
+        params = IPoolManager.SwapParams({zeroForOne: true, amountSpecified: -25, sqrtPriceLimitX96: SQRT_RATIO_1_4});
+        testSettings =
+            PoolSwapTest.TestSettings({withdrawTokens: true, settleUsingTransfer: false, currencyAlreadySent: false});
+
+        vm.expectEmit(true, true, true, false);
```
snreynolds, line 592: and again :)

snreynolds (commented):
One small change.. I may have missed calling out some of the vm.expectEmit() s but any event check before the transfer event should check all the topics and the data field so I think you can do vm.expectEmit()

snreynolds (dismissed):

src/V46909.sol
```diff
@@ -0,0 +1,21 @@
...
+    /// @notice Burn `amount` tokens of token type `id` from `from`.
+    /// @param from The address to burn tokens from.
+    /// @param id The currency to burn.
+    /// @param amount The amount to burn.
+    function _burnFrom(address from, uint256 id, uint256 amount) internal {
+        address sender = msg.sender;
+        if (from != sender && !isOperator[from][sender]) {
+            uint256 senderAllowance = allowance[from][sender][id];
+            if (senderAllowance != type(uint256).max) {
+                allowance[from][sender][id] = senderAllowance - amount;
```
hensha256, line 16: wondering if it would be nicer to have a custom error here?
like `if (amount > senderAllowance) revert X()` ?

i guess solmate doesnt have one on transferFrom though

snreynolds: non standard for transferFrom flows i think ..

test/V46909.t.sol
```diff
@@ -0,0 +1,42 @@
...
+    using CurrencyLibrary for Currency;
+
+    MockV46909 v46909;
+
+    function setUp() public {
+        (currency0, currency1) = deployMintAndApprove2Currencies();
+        v46909 = new MockV46909();
+    }
+
+    function testBurnFromFromWithApproval(address sender, Currency currency, uint256 mintAmount, uint256 transferAmount)
```
hensha256, line 19:
```suggestion
    function test_burnFrom_withApproval(address sender, Currency currency, uint256 mintAmount, uint256 transferAmount)
```

test/V46909.t.sol
```diff
@@ -0,0 +1,42 @@
...
+        v46909.burnFrom(sender, currency, transferAmount);
+
+        if (transferAmount <= mintAmount) {
+            if (mintAmount == type(uint256).max) {
+                assertEq(v46909.allowance(sender, address(this), id), type(uint256).max);
+            } else {
+                assertEq(v46909.allowance(sender, address(this), id), mintAmount - transferAmount);
+            }
+            assertEq(v46909.balanceOf(sender, id), mintAmount - transferAmount);
+        }
```
hensha256, line 40: please can you add a test that checks that it reverts when it doesnt have approval?

src/V46909.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.19;
+
+import {ERC6909} from "solmate/tokens/ERC6909.sol";
+
+contract V46909 is ERC6909 {
```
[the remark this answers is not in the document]
marktoda: also nit but I think `abstract contract` is more appropriate here as it's not intended to be deployed standalone

hensha256 (commented):
just a last couple of things

src/PoolManager.sol
```diff
@@ -377,15 +377,15 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
     }
 
     /// @inheritdoc IPoolManager
-    function mint(Currency currency, address to, uint256 amount) external noDelegateCall onlyByLocker {
+    function mint(address to, Currency currency, uint256 amount) external override noDelegateCall onlyByLocker {
```
marktoda, line 380: why switch the ordering here?

marktoda (commented):
looks great, just some minor suggestions

src/PoolManager.sol
snreynolds: this matches the underlying 6909 mint/burn functions

src/V46909.sol
zhongeric: yeah not sure about this one, seems like solmate / ERC6909 generally relies on underflow / overflow checks instead of explicitly throwing

test/V46909.t.sol
zhongeric: this is likely caught by the fuzz test right? i.e. when `mintAmount` is 0 the approval won't go through and thus all calls would revert?

src/V46909.sol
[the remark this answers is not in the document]
zhongeric: yeah not opposed, I do prefer `ERC6909Claims` over V46909 lol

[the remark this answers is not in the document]
zhongeric: Changed to ERC6909Claims

src/interfaces/IPoolManager.sol
```diff
@@ -6,12 +6,11 @@ import {PoolKey} from "../types/PoolKey.sol";
 import {Pool} from "../libraries/Pool.sol";
 import {IHooks} from "./IHooks.sol";
 import {IFees} from "./IFees.sol";
-import {IClaims} from "./IClaims.sol";
 import {BalanceDelta} from "../types/BalanceDelta.sol";
 import {PoolId} from "../types/PoolId.sol";
 import {Position} from "../libraries/Position.sol";
 
-interface IPoolManager is IFees, IClaims {
+interface IPoolManager is IFees {
```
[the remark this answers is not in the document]
zhongeric: Hmm yeah so it's a bit weird because the solmate implementation doesn't use the interface either, so getting a ton of compile errors due to conflicting implementations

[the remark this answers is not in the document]
zhongeric: Resolved offline - going to copy in the solmate implementation locally and inherit the interface

test/V46909.t.sol
zhongeric: Added solmate tests which catch this

src/interfaces/IPoolManager.sol
```diff
@@ -180,11 +177,11 @@ interface IPoolManager is IFees, IClaims {
...
 
-    /// @notice Called by the user to move value into Claims balance
-    function mint(Currency token, address to, uint256 amount) external;
+    /// @notice Called by the user to move value into ERC6909 balance
+    function mint(address to, uint256 id, uint256 amount) external;
 
-    /// @notice Called by the user to redeem their Claims balance
-    function burn(Currency token, uint256 amount) external;
+    /// @notice Called by the user to move value from ERC6909 balance
+    function burn(address from, uint256 id, uint256 amount) external;
```
snreynolds, line 184: Hmm what about placing these in the IERC6909Claims interface? Just thinking now that especially we've changed the interface to match the underlying 6909 one maybe it should? 

Could go either way tho..

test/utils/Deployers.sol
```diff
@@ -7,6 +7,7 @@ import {Currency, CurrencyLibrary} from "../../src/types/Currency.sol";
 import {IHooks} from "../../src/interfaces/IHooks.sol";
 import {IPoolManager} from "../../src/interfaces/IPoolManager.sol";
 import {PoolManager} from "../../src/PoolManager.sol";
+import {ERC6909Claims} from "../../src/ERC6909Claims.sol";
```
snreynolds, line 10: unnecessary import?

test/ERC6909Claims.t.sol
```diff
@@ -0,0 +1,364 @@
...
+contract ERC6909ClaimsTest is Test {
+    using CurrencyLibrary for Currency;
+
+    MockERC6909Claims token;
+
+    function setUp() public {
+        token = new MockERC6909Claims();
+    }
+
+    function test_burnFrom_withApproval(address sender, uint256 id, uint256 mintAmount, uint256 transferAmount)
```
snreynolds, line 17: maybe just add one with no approval on burn from? make sure it reverts

src/interfaces/IPoolManager.sol
zhongeric: Eh I think they should be in PoolManager since they are wrappers around the internal _mint and _burn and since they have the delta accounting logic in them

test/PoolManager.t.sol
```diff
@@ -496,52 +497,103 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
         swapRouter.swap(key, params, testSettings, ZERO_BYTES);
         snapEnd();
 
-        uint256 claimsBalance = manager.balanceOf(address(this), currency1);
-        assertEq(claimsBalance, 98);
+        uint256 erc6909Balance = manager.balanceOf(address(this), CurrencyLibrary.toId(currency1));
+        assertEq(erc6909Balance, 98);
+    }
+
+    function test_swap_GasMint6909IfNativeOutputNotTaken() public {
```
hensha256, line 517: `test_swap_mint6909IfNativeOutputNotTaken_gas`

hensha256 (dismissed):
lgtm!

snreynolds (approved):
looks good!

zhongeric:
Closing in favor of #447

PR #424, opened 2023-11-21
Lock on initialize

hensha256:
## Related Issue
Now that we are adding the ability of hooks to Access Lock (#404), we need to prevent external contacts from being able to do anything malicious by calling `initialize` with a hook address that can then access the lock on other functions.


zhongeric (approved):
Not sure if there are any additional edge cases created by this but implementation wise this looks good!

test/PoolManagerInitialize.t.sol
```diff
@@ -61,23 +61,23 @@ contract PoolManagerInitializeTest is Test, Deployers, GasSnapshot {
...
         } else if (key0.tickSpacing > manager.MAX_TICK_SPACING()) {
             vm.expectRevert(abi.encodeWithSelector(IPoolManager.TickSpacingTooLarge.selector));
-            manager.initialize(key0, sqrtPriceX96, ZERO_BYTES);
+            initializeRouter.initialize(key0, sqrtPriceX96, ZERO_BYTES);
         } else if (key0.tickSpacing < manager.MIN_TICK_SPACING()) {
             vm.expectRevert(abi.encodeWithSelector(IPoolManager.TickSpacingTooSmall.selector));
-            manager.initialize(key0, sqrtPriceX96, ZERO_BYTES);
-        } else if (key0.currency0 > key0.currency1) {
+            initializeRouter.initialize(key0, sqrtPriceX96, ZERO_BYTES);
+        } else if (key0.currency0 >= key0.currency1) {
```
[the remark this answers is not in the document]
hensha256: It got combined to have 1 fewer `if` statements. Kind of sad that our fuzz tests didnt catch that this test wasnt updated 😭
Might rename the error message to make it clearer

src/test/PoolModifyPositionTest.sol
```diff
@@ -43,13 +46,13 @@ contract PoolModifyPositionTest is PoolTestBase {
...
-            assert(delta0 > 0 || delta1 > 0);
-            assert(!(delta0 < 0 || delta1 < 0));
+            assertTrue(delta0 > 0 || delta1 > 0, "No positive delta");
+            assertTrue(!(delta0 < 0 || delta1 < 0), "Negative delta on deposit");
             if (delta0 > 0) _settle(data.key.currency0, data.sender, delta.amount0(), true);
             if (delta1 > 0) _settle(data.key.currency1, data.sender, delta.amount1(), true);
         } else {
-            assert(delta0 < 0 || delta1 < 0);
-            assert(!(delta0 > 0 || delta1 > 0));
+            assertTrue(delta0 < 0 || delta1 < 0 || data.key.fee.hasHookWithdrawFee(), "No negative delta");
```
hensha256, line 54: CI fuzzing randomly caught this edge case that it hasnt caught before 🤷‍♀️

.forge-snapshots/initialize.snap
```diff
@@ -1 +1 @@
-53843
\ No newline at end of file
+74927
```
snreynolds, line 1: 💀

snreynolds (approved):

.forge-snapshots/initialize.snap
hensha256: This does include the asserts in the test - this isnt the complete initialize cost

snreynolds: ah true

PR #432, opened 2023-12-05
Remove hook fees and protocol fee on withdrawal

hensha256:
## Description of changes

- Removes all hook fees (hooks can still take fees using the ACCESS_LOCK_FLAG)
- Removes protocol fees on liquidity removal
- Updates tests to reflect this
- Adds tests to show hooks taking fees using access lock


src/test/AccessLockHook.sol
```diff
@@ -13,6 +13,8 @@ import {ILockCallback} from "../interfaces/callback/ILockCallback.sol";
 import {MockERC20} from "solmate/test/utils/mocks/MockERC20.sol";
 import {Constants} from "../../test/utils/Constants.sol";
 import {PoolIdLibrary} from "../types/PoolId.sol";
+import {BalanceDelta} from "../types/BalanceDelta.sol";
+import {console2} from "lib/forge-std/src/console2.sol";
```
marktoda, line 17: nit rm console

marktoda (approved):
nice lgtm

marktoda (dismissed):

marktoda (approved):

src/Fees.sol
```diff
@@ -52,44 +50,22 @@ abstract contract Fees is IFees, Owned {
...
-        }
-        return true;
-    }
-
-    /// @dev Only the lower 12 bits are used here to encode the fee denominator.
     function _isFeeWithinBounds(uint16 fee) internal pure returns (bool) {
         if (fee != 0) {
-            uint16 fee0 = fee % 64;
-            uint16 fee1 = fee >> 6;
+            uint16 fee0 = fee % 256;
```
snreynolds, line 67: Why change this? If we want to increase protocol fee granularity I think thats a separate discussion. I think this means the fee can be a lot smaller and I thought we intentionally had it bound before (both from a min and a max).

In any case it's worth a comment on what the min fee value is now that there are 8 bits that can be used instead of 6.

snreynolds (changes requested):

src/Fees.sol
snreynolds: Yeah I think that's fine, can you add a comment? I just wasn't expecting that to change, not that its a bad change but just deserves some reasoning :)

snreynolds (approved):

PR #434, opened 2023-12-06
Before after add/remove liquidity, remove all modifyPosition hooks

### LINKED ISSUE #350: Split {before,after}ModifyPosition into {before,after}Mint and {before,after}Burn

rileydcampbell:
## Which issue does this pull request resolve?
This PR fixes #350. The main idea is:
> Right now, beforeModifyPosition is a risky permission. Liquidity providers need to audit and trust the hook code or risk never being able to withdraw their money. That makes it harder to fearlessly integrate pools that have this permission.

## Description of changes
We broke up the modify position hook into an `addLiquidity` and `removeLiquidity` hook. This allows for a better separation of permissions. The following changes were made to implement these hooks.
- Add before/after addLiquidity 
- Add before/after removeLiquidity 
- Remove modifyPosition hooks
- Update existing tests and added new ones


test/AccessLock.t.sol
```diff
@@ -139,9 +140,28 @@ contract AccessLockTest is Test, Deployers {
...
+        modifyPositionRouter.addLiquidity(key, LIQ_PARAMS, ZERO_BYTES);
+        modifyPositionRouter.removeLiquidity(key, LIQ_PARAMS, abi.encode(amount, AccessLockHook.LockAction.Mint));
+
+        uint256 balanceOfAfter0 = MockERC20(Currency.unwrap(currency0)).balanceOf(address(this));
+        uint256 balanceOfAfter1 = MockERC20(Currency.unwrap(currency1)).balanceOf(address(this));
+
+        // The balance of our contract should be equal to our original balance because we added and removed our liquidity.
+        // TODO: Maybe due to a rounding issue, but balanceOfBefore0 == balanceOfAfter0 ?
+        assertTrue(balanceOfBefore0 - balanceOfAfter0 <= 1);
+        assertTrue(balanceOfBefore1 - balanceOfAfter1 - amount <= 1);
```
rileydcampbell, line 157: Not sure why I had an off by one issue here?

hensha256 (commented):
I feel like theres improvements in efficiency of the internal code that can be made now that we have separation of add/remove liquidity functions.

For example, in the Pool library in `modifyPosition`, it has to do checks
```
if (params.liquidityDelta > 0) {
   /// aka if this is an ADD liquidity then do X
} 
if (params.liquidityDelta < 0) {
   /// aka if this is a REMOVE liquidity then do Y
}
```

this PR gives the nice separation of entry points, but then still leads to the same internal flow of checking whether liquidityDelta is positive or negative, even though that is actually known in advance

I think we should maybe investigate splitting pool logic into internal `addLiquidity` and `removeLiquidity`, where `removeLiquidity` never has to do any of the `params.liquidityDelta > 0` checks and vice versa?

rileydcampbell:
> I feel like theres improvements in efficiency of the internal code that can be made now that we have separation of add/remove liquidity functions.
> 
> For example, in the Pool library in `modifyPosition`, it has to do checks
> 
> ```
> if (params.liquidityDelta > 0) {
>    /// aka if this is an ADD liquidity then do X
> } 
> if (params.liquidityDelta < 0) {
>    /// aka if this is a REMOVE liquidity then do Y
> }
> ```
> 
> this PR gives the nice separation of entry points, but then still leads to the same internal flow of checking whether liquidityDelta is positive or negative, even though that is actually known in advance
> 
> I think we should maybe investigate splitting pool logic into internal `addLiquidity` and `removeLiquidity`, where `removeLiquidity` never has to do any of the `params.liquidityDelta > 0` checks and vice versa?

Definitely agree! I can take a closer look at the Pool logic. This would be as a follow up yeah? Not blocking

snreynolds (commented):
seems like a good structure although hard to give a full review knowing a lot of this will change with Alice's and Mark's changes... could you pull those in? 😄

src/PoolManager.sol
```diff
@@ -205,18 +252,45 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
+            ) {
                 revert Hooks.InvalidHookResponse();
             }
         }
 
+        // We only want to clear the current hook if it was set in setCurrentHook in this execution frame.
+        if (set) Lockers.clearCurrentHook();
+    }
+
+    function _modifyPosition(PoolKey memory key, IPoolManager.PoolModifyPositionParams memory params)
```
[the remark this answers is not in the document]
rileydcampbell: I think this will just go away entirely after Alice's change.

src/PoolManager.sol
```diff
@@ -205,18 +252,45 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
+        internal
+        returns (BalanceDelta delta)
+    {
+        PoolId id = key.toId();
         Pool.FeeAmounts memory feeAmounts;
         (delta, feeAmounts) = pools[id].modifyPosition(
             Pool.ModifyPositionParams({
```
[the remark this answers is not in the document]
rileydcampbell: Yep yep, sounds good 👍

src/PoolManager.sol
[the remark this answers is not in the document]
rileydcampbell: yeah ++ [this](https://github.com/Uniswap/v4-core/pull/434#discussion_r1420981199)

rileydcampbell:
> seems like a good structure although hard to give a full review knowing a lot of this will change with Alice's and Mark's changes... could you pull those in? 😄

Yeah will do! Might take me a sec to do the refactor/fix all the merge conflicts

hensha256:
>This would be as a follow up yeah? Not blocking

Yeah we can make it a followup, please can you open a github issue so we dont forget?

src/PoolManager.sol
```diff
@@ -205,18 +252,45 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
+                revert Hooks.InvalidHookResponse();
+            }
+        }
+
+        delta = _modifyPosition(
+            key,
+            IPoolManager.PoolModifyPositionParams({
+                tickLower: params.tickLower,
+                tickUpper: params.tickUpper,
+                liquidityDelta: -params.liquidityDelta.toInt128()
```
marktoda, line 272: IMO this is no longer `liquidityDelta` since you're always accepting positive value and flipping it negative in the remove case, so is more like `liquidityToRemove`

marktoda: also -- you don't sanitize `liquidityDelta` to ensure positive here? so I can pass in negative value to `removeLiquidity` and it'll get negated so we'll actually end up removing liquidity :D

marktoda: and vice versa for addLiquidity, if you pass negative value it will actually remove hahah

marktoda: oh nvm.. you made liquidityDelta an unsigned int

rileydcampbell: Pulled in your changes, so no longer relevant

src/libraries/Hooks.sol
```diff
@@ -161,12 +168,22 @@ library Hooks {
...
+            }
         } else {
-            return true;
+            if (key.hooks.hasPermission(BEFORE_REMOVE_LIQUIDITY_FLAG)) {
+                shouldExecute = self.callHookNoopable(
+                    abi.encodeWithSelector(IHooks.beforeRemoveLiquidity.selector, msg.sender, key, params, hookData)
+                );
+            } else {
+                shouldExecute = true;
+            }
```
rileydcampbell, line 186: Feel like this looks a little ugly, could also change to something like:
```
    /// @notice calls beforeModifyPosition hook if permissioned and validates return value
    function beforeModifyPosition(
        IHooks self,
        PoolKey memory key,
        bytes calldata hookData
    ) internal returns (bool shouldExecute) {
        if (params.liquidityDelta >= 0 && key.hooks.hasPermission(BEFORE_ADD_LIQUIDITY_FLAG)) {
            shouldExecute = self.callHookNoopable(
                abi.encodeWithSelector(IHooks.beforeAddLiquidity.selector, msg.sender, key, params, hookData)
            );
        } else if (params.liquidityDelta < 0 && key.hooks.hasPermission(BEFORE_REMOVE_LIQUIDITY_FLAG)) {
            shouldExecute = self.callHookNoopable(
                abi.encodeWithSelector(IHooks.beforeRemoveLiquidity.selector, msg.sender, key, params, hookData)
            );
        } else {
            shouldExecute = true;
        }
    }
 ```
It's a little more gas ~100 but maybe that's ok?

Comment (zhongeric) [reply src/libraries/Hooks.sol:186]: I think we should expose separate getters for querying beforeAdd/RemoveLiquidity and deprecating beforeModifiyPosition + afterModifyPosition so you don't need this conditional

File: src/test/NoOpTestHooks.sol [hunk 6]
```diff
@@ -14,8 +14,10 @@ contract NoOpTestHooks is BaseTestHooks {
             Hooks.Permissions({
                 beforeInitialize: false,
                 afterInitialize: false,
-                beforeModifyPosition: true,
-                afterModifyPosition: false,
+                beforeAddLiquidity: true,
+                afterAddLiquidity: false,
+                beforeRemoveLiquidity: false,
```
zhongeric, line 19: nit: beforeRemoveLiquidity should be true as well

test/AccessLock.t.sol
zhongeric: Ya should equal, that's werid

test/AccessLock.t.sol
```diff
@@ -172,7 +194,30 @@ contract AccessLockTest is Test, Deployers {
         assertEq(MockERC20(Currency.unwrap(currency1)).balanceOf(address(accessLockHook)), amount);
     }
 
-    function test_beforeModifyPosition_swap_succeedsWithAccessLock() public {
+    function test_beforeRemoveLiquidity_take_succeedsWithAccessLock() public {
+        // Add liquidity so there is something to take.
+        delta = modifyPositionRouter.modifyPosition(key, LIQ_PARAMS, ZERO_BYTES);
+        // Can't take more than the manager has.
+        uint128 takeAmount = uint128(key.currency1.balanceOf(address(manager)) - 1);
```
zhongeric, line 201: shouldn't need the - 1 here right?

test/AccessLock.t.sol
```diff
@@ -172,7 +194,30 @@ contract AccessLockTest is Test, Deployers {
...
+        // Hook only takes currency 1 rn.
+        modifyPositionRouter.modifyPosition(
+            key, REMOVE_LIQ_PARAMS, abi.encode(takeAmount, AccessLockHook.LockAction.Take)
+        );
+        uint256 balanceOfAfter0 = MockERC20(Currency.unwrap(currency0)).balanceOf(address(this));
+        uint256 balanceOfAfter1 = MockERC20(Currency.unwrap(currency1)).balanceOf(address(this));
+
+        // The balance of our contract should be equal to our original balance because we added and removed our liquidity.
+        // TODO: Maybe due to a rounding issue, but balanceOfBefore0 == balanceOfAfter0 ?
+        assertTrue(balanceOfBefore0 + uint256(uint128(delta.amount0())) - balanceOfAfter0 <= 1);
```
zhongeric, line 215: I'd write these tests to test equality and just let them fail for now until we figure out why its off by one

src/libraries/Hooks.sol
snreynolds: I think I prefer this branching. And @zhongeric I think Mark and I talked about this and we kinda liked having the beforeModifyPos/afterModifyPos as the entrypoints in the Hooks library so that we are removing the liquidityDelta logic from core codepaths

snreynolds (commented):
One last overall

Should we be hitting the beforeAddLiquidity logic for liquidityParams of 0? I think in the worst case someone could revert on just collecting fees but not wanting to change their position. Either we don't call out to hooks or maybe have the 0 liquidityParams branch out to the afterAddLiquidity branch?

Also should we change name to modifyLiquidity()?

rileydcampbell:
I removed any of the add/remove liquidity hook calls for a `liquidityDelta == 0`.

test/AccessLock.t.sol
rileydcampbell: Turns out this is a known issue see [here](https://github.com/Uniswap/v3-core/issues/570).

src/interfaces/IPoolManager.sol
```diff
@@ -155,8 +155,13 @@ interface IPoolManager is IFees, IClaims {
...
 
-    /// @notice Modify the position for the given pool
-    function modifyPosition(PoolKey memory key, ModifyPositionParams memory params, bytes calldata hookData)
+    /// @notice Modify the liquidity for the given pool
+    /// @dev Poke by calling with a zero liquidityDelta
+    /// @param key The pool to modify liquidity in
+    /// @param params The parameters for modifying the liquidity
+    /// @param hookData Any data to pass to the callback, via `ILockCallback(msg.sender).lockAcquired(data)`
+    /// @return delta The balance delta of the liquidity
+    function modifyLiquidity(PoolKey memory key, ModifyPositionParams memory params, bytes calldata hookData)
```
snreynolds, line 164: laaaast lil nit: ModifyLiquidityParams ?

src/libraries/Hooks.sol
```diff
@@ -161,12 +168,16 @@ library Hooks {
         IPoolManager.ModifyPositionParams memory params,
         bytes calldata hookData
     ) internal returns (bool shouldExecute) {
```
snreynolds, line 170: and same name change on these functions? Just thinking so that they are all standardized

beforeModifyPosition -> beforeModifyLiquidity 
afterModifyPosition -> afterModifyLiquidity

rileydcampbell: fixed 👍

snreynolds (approved):
🚢

PR #437, opened 2023-12-07
A few cleanup tasks

hensha256:
- fix compiler warnings from unused vars
- increase fuzzing runs on CI and locally. This caused a bunch of tests to fail - fixed those tests and also changed lots of `vm.assume` to use `bound`
- remove unused fuzzing config

My fix on the fuzzing tests of access lock werent very good fixes because they broke a lot, and we're about to remove them, so felt like a waste of time lol


src/test/PoolModifyPositionTest.sol
```diff
@@ -53,7 +53,7 @@ contract PoolModifyPositionTest is Test, PoolTestBase {
         (,,, int256 delta1) = _fetchBalances(data.key.currency1, data.sender);
 
         // These assertions only apply in non lock-accessing pools.
-        if (!data.key.hooks.hasPermissionToAccessLock()) {
+        if (!data.key.hooks.hasPermissionToAccessLock() && !data.key.fee.hasHookWithdrawFee()) {
```
snreynolds, line 56: Why does this change in this PR? Are we hitting a fuzz case here?

snreynolds (approved):
lgtm just 1 q

src/test/PoolModifyPositionTest.sol
hensha256: Yeah fuzz case - a hook withdrawal fee changes the output amounts expected of a modify position.
This addition will be removed again in #432

snreynolds (approved):

PR #439, opened 2023-12-08
refactor: hooks callsites

marktoda:
This PR cleans up the hooks callsites to remove duplicated code and standardize validation of hooks responses

It also changes the CurrentHook behavior to always set the current hook directly before making a hook call, and unset it directly after. This means it isn't set if a hook call is not required, but also is unset and re-set if both before* and after* are used


src/libraries/Hooks.sol
```diff
@@ -83,6 +91,147 @@ library Hooks {
...
+        internal
+    {
+        if (key.hooks.shouldCallAfterDonate()) {
+            self.callHook(
+                abi.encodeWithSelector(IHooks.afterDonate.selector, msg.sender, key, amount0, amount1, hookData)
+            );
+        }
+    }
+
     function shouldCallBeforeInitialize(IHooks self) internal pure returns (bool) {
```
marktoda, line 235: maybe should remove these `shouldCall*` things and just inline them in the hook functions themselves

snreynolds (commented):
cleannnnnn

src/libraries/Hooks.sol
marktoda: agreed done

src/libraries/Hooks.sol
```diff
@@ -83,6 +91,147 @@ library Hooks {
...
+        if (!success) _revert(result);
+
+        selector = abi.decode(result, (bytes4));
+
+        // We only want to clear the current hook if it was set in setCurrentHook in this execution frame.
+        if (set) Lockers.clearCurrentHook();
+    }
+
+    /// @notice performs a hook call using the given calldata on the given hook
+    function callHook(IHooks self, bytes memory data) internal {
```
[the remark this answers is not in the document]
marktoda: oh because noopable calls _callHook but does its own validation of the response

snreynolds (approved):

test/Hooks.t.sol
```diff
@@ -154,16 +154,16 @@ contract HooksTest is Test, Deployers, GasSnapshot {
...
-        assertFalse(Hooks.shouldCallAfterInitialize(hookAddr));
-        assertFalse(Hooks.shouldCallBeforeModifyPosition(hookAddr));
-        assertFalse(Hooks.shouldCallAfterModifyPosition(hookAddr));
-        assertFalse(Hooks.shouldCallBeforeSwap(hookAddr));
-        assertFalse(Hooks.shouldCallAfterSwap(hookAddr));
-        assertFalse(Hooks.shouldCallBeforeDonate(hookAddr));
-        assertFalse(Hooks.shouldCallAfterDonate(hookAddr));
-        assertFalse(Hooks.hasPermissionToNoOp(hookAddr));
-        assertFalse(Hooks.hasPermissionToAccessLock(hookAddr));
+        assertFalse(Hooks.hasPermission(hookAddr, Hooks.BEFORE_INITIALIZE_FLAG));
```
hensha256, line 157: nit: definitely prefer the `hookAddr.hasPermission(X)` syntax

hensha256 (dismissed):
this is a beautiful pull request

src/libraries/Hooks.sol
```diff
@@ -83,47 +92,165 @@ library Hooks {
             );
     }
 
-    function shouldCallBeforeInitialize(IHooks self) internal pure returns (bool) {
-        return uint256(uint160(address(self))) & BEFORE_INITIALIZE_FLAG != 0;
+    /// @notice performs a hook call using the given calldat aon the given hook
```
rileydcampbell, line 95:
```suggestion
    /// @notice performs a hook call using the given calldata on the given hook
```

PR #442, opened 2023-12-11
Remove nested locking, access lock, and lockCaller

hensha256:
- Remove access lock hook flag (ready for a different custom accounting implementation in #450)
- Remove nested locking - other addresses can call functions as long as _someone_ has a lock. The deltas that those addresses change are then added to their own address's entry in the mapping. Essentially calling `lock` and then `lockAcquired` in sub-locks had become somewhat redundant
- Remove `lockCaller`. Currently an EOA can still call to make the lock, but we will no longer expose lockCaller due to it not making sense after removing nested locking


src/PoolManager.sol
```diff
@@ -88,27 +88,21 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
...
-        if (caller == address(hook) && hook.hasPermission(Hooks.ACCESS_LOCK_FLAG)) return;
-        revert LockedBy(locker, address(hook));
-    }
-
     /// @inheritdoc IPoolManager
     function initialize(PoolKey memory key, uint160 sqrtPriceX96, bytes calldata hookData)
         external
         override
-        onlyByLocker
+        isLocked
```
hensha256, line 105: I feel like this can be removed from initialize now that we dont have access lock?

src/PoolManager.sol
```diff
@@ -135,35 +129,33 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
 
     /// @inheritdoc IPoolManager
     function lock(address lockTarget, bytes calldata data) external payable override returns (bytes memory result) {
-        Lockers.push(lockTarget, msg.sender);
+        // Get the lock caller because thats an EOA and is not user-controlable
+        if (Locker.isLocked()) revert LockedBy(Locker.getLocker());
+
+        Locker.setLockerAndCaller(lockTarget, msg.sender);
```
marktoda, line 135: are these even neceessary anymore? feel like now they're kinda like `tx.origin` so maybe not that useful to expose?

src/PoolManager.sol
```diff
@@ -135,35 +129,33 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, Claims {
 
     /// @inheritdoc IPoolManager
     function lock(address lockTarget, bytes calldata data) external payable override returns (bytes memory result) {
-        Lockers.push(lockTarget, msg.sender);
+        // Get the lock caller because thats an EOA and is not user-controlable
+        if (Locker.isLocked()) revert LockedBy(Locker.getLocker());
```
marktoda, line 133: maybe `AlreadyLocked` is more accurate error now?

src/libraries/Hooks.sol
```diff
@@ -87,15 +84,13 @@ library Hooks {
...
         return address(hook) == address(0)
             ? !fee.isDynamicFee()
-            : (uint160(address(hook)) >= ACCESS_LOCK_FLAG || fee.isDynamicFee());
+            : (uint160(address(hook)) >= NO_OP_FLAG || fee.isDynamicFee());
     }
 
     /// @notice performs a hook call using the given calldata on the given hook
     /// @return expectedSelector The selector that the hook is expected to return
     /// @return selector The selector that the hook actually returned
     function _callHook(IHooks self, bytes memory data) private returns (bytes4 expectedSelector, bytes4 selector) {
```
marktoda, line 93: think we can even do a nicer refactor here now where we just call the hook in each callsite function and change this shared function to `validateHookResponse()` 
i can do in a followup tho

src/PoolManager.sol
hensha256: if we dont have it, how will our router know who to pull tokens from?
like our `router.lockAcquired` is called, they do some swap, and whose tokens do they use to settle? In my head they have to use `lockCaller`

marktoda: True - but what about a situation where like:
- I call router to do swaps
- router initiates lock
- I'm lockcaller
- ... 
- some hook or token or something calls router 
- I'm still lockcaller

marktoda: Maybe lockCaller / lockTarget isn't actually useful anymore?

hensha256: Yeah I think we still want it out? Idk if we decided 100%? Happy to pull it out in this PR if you 2 agree, otherwise we can make it a future issue to discuss

src/PoolManager.sol
```diff
@@ -135,35 +129,33 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
...
 
-        address locker = Lockers.getCurrentLocker();
-        int256 current = currencyDelta[locker][currency];
+        int256 current = currencyDelta[msg.sender][currency];
         int256 next = current + delta;
 
         unchecked {
             if (next == 0) {
-                Lockers.decrementNonzeroDeltaCount();
+                Locker.decrementNonzeroDeltaCount();
```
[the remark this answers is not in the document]
hensha256: yah ive thought they should be in different libraries for quite a while... its a library that handles 2 separate things

src/PoolManager.sol
hensha256: i think this should be removed in another PR. THis PR is already so big and getting hard to review - and this change would require updating how all our tests initialize pools 🤪

hensha256: https://github.com/Uniswap/v4-core/issues/477

src/PoolManager.sol
```diff
@@ -135,35 +130,33 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
 
     /// @inheritdoc IPoolManager
     function lock(address lockTarget, bytes calldata data) external payable override returns (bytes memory result) {
-        Lockers.push(lockTarget, msg.sender);
+        // Get the lock caller because thats an EOA and is not user-controlable
+        if (Locker.isLocked()) revert AlreadyLocked();
```
ewilz, line 134: Will disallowing multiple locks break composability?? i.e. router has to open a lock, as well as any other integrating contract that's not a hook. What if they get called after another non-hook contract has already locked it, they now have 2 ways they'd need to integrate?

ewilz: building on this ---^ the word `lock` seems like a misnomer now, in fact it's like the opposite, now it's open for business (right?? 😄 ) Wondering if we need to change our mental model if this is the direction we're going.....

ewilz: if the gas savings are for hooks specifically, wonder if we can allow multiple locks, but free up hooks from needing to use them? (really just spitballing/food-for-thought, but just making sure we've thought through this new mental model of locks)

hensha256: >Will disallowing multiple locks break composability??

Yeah I think that integrating contracts who want to be able to be a sub-lock too will need to write something more like this to just jump straight to their execution logic if someone else has already called lock at an upper level
```
if (!poolManager.isLocked) {
   poolManager.lock(lockData);
} else {
   lockAcquired(lockData)
}
```

Thoughts? Do you think thats too janky?

hensha256: And yes I agree the word `lock` feels completely wrong haha

ewilz: yah, it's annoying boilerplate but ultimately saves gas, and I'm having trouble finding a good reason to keep lockers nested or stack/recursed like it was way back.

 Although the lock vernacular doesn't seem to make much sense anymore..... but naming is so hAaRrdddDD......

```
if (!poolManager.ledgerOpened?) {
   poolManager.openLedger(ledgerData);
} else {
   ledgerCallback(ledgerData)
}
```

??

```
if (!poolManager.settlementInitiated) {
   poolManager.openSettlement(settlementData);
} else {
   settlementCallback(settlementData)
}```

File: src/libraries/Unlocker.sol [hunk 7]
```diff
@@ -0,0 +1,43 @@
...
+/// TODO: This library can be deleted when we have the transient keyword support in solidity.
+library Unlocker {
+    // The slot holding the unlocker, transiently
+    uint256 constant UNLOCKER_SLOT = uint256(keccak256("Unlocker")) - 1;
+
+    /// @notice Thrown when trying to set the unlock address as address(0)
+    /// we use unlocker==address(0) to signal that the pool is not unlocked
+    error InvalidUnlocker();
+
+    function setUnlocked(address unlocker) internal {
```
ewilz, line 16: wondering if we only need to keep track of whether it's unlocked? what's the point of tracking the very first address that opened the lock?

src/PoolManager.sol
```diff
@@ -88,27 +89,21 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
...
+    function getUnlocker() external view override returns (address) {
+        return Unlocker.getUnlocker();
     }
 
-    /// @notice This will revert if a function is called by any address other than the current locker OR the most recently called, pre-permissioned hook.
-    modifier onlyByLocker() {
-        _checkLocker(msg.sender, Lockers.getCurrentLocker(), Lockers.getCurrentHook());
+    /// @notice This will revert if a function is called by any address other than the current unlocker OR the most recently called, pre-permissioned hook.
+    modifier isUnlocked() {
+        if (!Unlocker.isUnlocked()) revert ManagerNotUnlocked();
```
snreynolds, line 98: I would change this to `ManagerLocked()` instead of double negative

src/libraries/Unlocker.sol
snreynolds: i agree dont see a point, but it costs no extra overhead also

src/libraries/Unlocker.sol
```diff
@@ -0,0 +1,43 @@
...
+import {IHooks} from "../interfaces/IHooks.sol";
+
+/// @notice This is a temporary library that allows us to use transient storage (tstore/tload)
+/// TODO: This library can be deleted when we have the transient keyword support in solidity.
+library Unlocker {
+    // The slot holding the unlocker, transiently
+    uint256 constant UNLOCKER_SLOT = uint256(keccak256("Unlocker")) - 1;
+
+    /// @notice Thrown when trying to set the unlock address as address(0)
+    /// we use unlocker==address(0) to signal that the pool is not unlocked
```
snreynolds, line 13: that the pool is `locked`. i find the double negatives confusing

src/libraries/NonZeroDeltaCount.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.20;
+
+import {IHooks} from "../interfaces/IHooks.sol";
```
snreynolds, line 4: was there a reason to split it into two files now? just bc nonzerodelta count is not per locker anymore?

snreynolds (commented):
looks good

I honestly find all the `unlock` nomenclature pretty hard to read..
unlockCallback / Unlocker.isUnlocked/Unlocker.setUnlocked... its just pretty clunky reading it.
and like having things like `Unlocker.clearUnlock` to me is really just  lock()

snreynolds:
> looks good
> 
> I honestly find all the `unlock` nomenclature pretty hard to read.. unlockCallback / Unlocker.isUnlocked/Unlocker.setUnlocked... its just pretty clunky reading it. and like having things like `Unlocker.clearUnlock` to me is really just lock()

I dont want naming to be blocker on merge but maybe we can reassess? I just find it way more unreadable than before

src/libraries/Unlocker.sol
hensha256: I made #487 to handle this in a new PR as this one is so big already

src/PoolManager.sol
```diff
@@ -135,35 +130,33 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
...
     /// @inheritdoc IPoolManager
     function lock(address lockTarget, bytes calldata data) external payable override returns (bytes memory result) {
-        Lockers.push(lockTarget, msg.sender);
+        // Get the lock caller because thats an EOA and is not user-controlable
+        if (Locker.isLocked()) revert AlreadyLocked();
+
+        Locker.setLocker(lockTarget);
 
         // the caller does everything in this callback, including paying what they owe via calls to settle
         result = ILockCallback(lockTarget).lockAcquired(msg.sender, data);
```
snreynolds, line 139: interesting. one thing I just thought of is that there isn't really a `result` subsequent callers on the pool manager can get now. I don't think its bad, just a breaking integration change.

i guess new integrators would just ingest the return from any function they call directly

src/libraries/NonZeroDeltaCount.sol
```diff
@@ -0,0 +1,39 @@
...
+        uint256 slot = NONZERO_DELTA_COUNT;
+        assembly {
+            let count := tload(slot)
+            count := add(count, 1)
+            tstore(slot, count)
+        }
+    }
+
+    /// @notice Potential to underflow.
+    /// Current usage ensures this will not happen because we call decrememnt with known boundaries (only up to the number of times we call increment).
```
snreynolds, line 30:
```suggestion
    /// Current usage ensures this will not happen because we call decrement with known boundaries (only up to the number of times we call increment).
```

test/Locker.t.sol
```diff
@@ -0,0 +1,46 @@
...
+pragma solidity ^0.8.20;
+
+import {Test} from "forge-std/Test.sol";
+import {Locker} from "../src/libraries/Locker.sol";
+
+contract LockerTest is Test {
+    address constant ADDRESS_AS = 0xaAaAaAaaAaAaAaaAaAAAAAAAAaaaAaAaAaaAaaAa;
+    address constant ADDRESS_BS = 0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB;
+
+    function test_setLocker(address locker) public {
```
snreynolds, line 11: for fuzz test I would add the _fuzz prefix.

so its test_fuzz

good way to standardize and be able to use regex to filter out tests if we need in CI or anywhere else

test/NonZeroDeltaCount.t.sol
```diff
@@ -0,0 +1,41 @@
...
+
+    function test_decrementNonzeroDeltaCount() public {
+        NonZeroDeltaCount.increment();
+        NonZeroDeltaCount.decrement();
+        assertEq(NonZeroDeltaCount.read(), 0);
+    }
+
+    // Reading from right to left. Bit of 0: call increase. Bit of 1: call decrease.
+    // The library allows over over/underflow so we dont check for that here
+    function test_nonZeroDeltaCount_fuzz(uint256 instructions) public {
```
snreynolds, line 24:
```suggestion
    function test_fuzz_nonZeroDeltaCount(uint256 instructions) public {
```

snreynolds (approved):
some small last changes!

src/PoolManager.sol
hensha256: Right but also the address who returns the result from `lockAcuired` is also the same address that called `lock` anyway (once we remove EOA calling)... so they could also just use transient storage to "return" that value to themselves. Arguably the return value isnt needed at all right?

src/libraries/NonZeroDeltaCount.sol
hensha256: oops just seen this sorry - yeah just because it feels like very separate things they feel unrelated

src/PoolManager.sol
snreynolds: that is true if we remove EOA calling

PR #443, opened 2023-12-12
Swap helper in Deployers

saucepoint:
## Related Issue
I'm planning on leveraging v4-core's `Deployers` to streamline best practices around hook-testing

also want to drop [this file from v4-template](https://github.com/uniswapfoundation/v4-template/blob/main/test/utils/HookTest.sol) to reduce maintenance/versioning

## Description of changes

* Added a simple swap helper where we assume unlimited impact + ERC20 transfers
* Set the global `currency0, currency1` when tokens are deployed + approved

---

edit: Also open to organizing this function elsewhere?


test/utils/Deployers.sol
```diff
@@ -159,6 +165,23 @@ contract Deployers {
         uninitializedNativeKey.fee = 100;
     }
 
+    /// @notice Helper function for a simple swap that allows for unlimited price impact
+    function swap(PoolKey memory _key, bool zeroForOne, int256 amountSpecified, bytes memory hookData)
```
hensha256, line 169: wondering if it would be helpful to add native input support too? either a different function `swapNativeInput` or generalise this one a bit more.
Something like `bool isNativeInput = (zeroForOne && key.currency0 == NATIVE) ? true : false`
and then i think youd only have to change 2 things:
- `currencyAlreadySent: isNativeInput`
- `{ value: isNativeInput ? amountSpecified : 0 }`

hensha256: hm i guess that `value` formula only works if this is exact input not exact output... maybe it becomes too annoying to do this, what do you think?

saucepoint: ah yeah. there's no reasonable way to cleanly abstract exact output with native token input

leaning towards a `swapNativeInput` vs overloading the simple one with an "optional" msgValue parameter.

hensha256: sgtm!

saucepoint:
I modified `swap` to accept native tokens i.f.f. `zeroForOne and exactInput` (or oneForZero). `swapNativeInput` can handle both exactInput or exactOutput for native token input

tests are a bit exhaustive, but i suppose thats not a bad thing

also: looks like currencyAlreadySent is for EOA-initiated locks and isnt a flag for native token swaps

hensha256 (approved):

PR #447, opened 2023-12-14
fix: Add ERC6909 support

zhongeric:
## Related Issue
Rebased now signed commits into the PR for attribution. Functionaly same as #422 

## Description of changes


src/ERC6909Claims.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import {ERC6909} from "./ERC6909.sol";
+
+/// @notice ERC6909Claims inherits ERC6909 and implements an internal burnFrom function
+abstract contract ERC6909Claims is ERC6909 {
+    /// @notice Burn `amount` tokens of token type `id` from `from`.
```
marktoda, line 8: think its important to note that if from !== msg.sender they must be operator

marktoda (dismissed):
lgtm! would be good to have sara or alice look again too tho

snreynolds (approved):
lgtm

PR #449, opened 2023-12-18
Accumulate protocol fees as 6909s 

zhongeric:
Squashed #392 to clean up history


src/PoolManager.sol
```diff
@@ -307,6 +311,17 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
         emit ProtocolFeeUpdated(id, newProtocolFee);
     }
 
+    function collectProtocolFees(address recipient, uint256 id, uint256 amount)
```
zhongeric, line 314: There was a comment on the previous PR on whether or not we should expose this entrypoint for the feeController. I think governance would appreciate just being able to call this func on core, and not having to lock + handle the callback. What do ppl think?

hensha256: yeah i think worth keeping it as this function so its as simple as possible without needing a custom lockAcquired function

hensha256: though do you think its worth putting a `burn` function on our 6909 that anyone can call (without a lock) that just burn their 6909s and get the tokens? `burn` doesnt feel like it would need a lock as it wouldnt change any of the reserves

test/PoolManager.t.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.20;
 
+import {console2} from "forge-std/console2.sol";
```
hensha256, line 4: remove :)

test/PoolManager.t.sol
```diff
@@ -982,19 +1002,48 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
+
+        uint256 feesAccrued = manager.balanceOf(address(feeController), currency1.toId());
+        console2.log(feesAccrued);
+
+        vm.prank(address(feeController));
+        if (balanceToClaim == 0 || balanceToClaim == feesAccrued) {
+            // Expect that they claimed their entire balance
+            manager.collectProtocolFees(protocolFeeRecipient, currency1.toId(), balanceToClaim);
+            assertEq(currency1.balanceOf(protocolFeeRecipient), feesAccrued);
+            assertEq(manager.balanceOf(address(feeController), currency0.toId()), 0);
```
hensha256, line 1032:
```suggestion
            assertEq(manager.balanceOf(address(feeController), currency1.toId()), 0);
```

i think this needs to be for currency1 to check that their 6909 balance is 0 (because fees were accrued in currency1)

test/PoolManager.t.sol
```diff
@@ -982,19 +1002,48 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
+        (Pool.Slot0 memory slot0,,,) = manager.pools(key.toId());
         assertEq(slot0.protocolFee, protocolFee);
+
+        // token 1 for 0
+        IPoolManager.SwapParams memory swapParams =
+            IPoolManager.SwapParams(false, SafeCast.toInt256(swapAmount), TickMath.MAX_SQRT_RATIO - 1);
+        swapRouter.swap(key, swapParams, PoolSwapTest.TestSettings(true, true, false), ZERO_BYTES);
+
+        uint256 feesAccrued = manager.balanceOf(address(feeController), currency1.toId());
+        console2.log(feesAccrued);
```
hensha256, line 1025: remove :)

test/PoolManager.t.sol
```diff
@@ -982,19 +1002,48 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
+        manager.setProtocolFee(key);
+        (Pool.Slot0 memory slot0,,,) = manager.pools(key.toId());
         assertEq(slot0.protocolFee, protocolFee);
+
+        // token 1 for 0
+        IPoolManager.SwapParams memory swapParams =
+            IPoolManager.SwapParams(false, SafeCast.toInt256(swapAmount), TickMath.MAX_SQRT_RATIO - 1);
+        swapRouter.swap(key, swapParams, PoolSwapTest.TestSettings(true, true, false), ZERO_BYTES);
+
+        uint256 feesAccrued = manager.balanceOf(address(feeController), currency1.toId());
```
hensha256, line 1024: maybe assert that currency0 balance is 0 here

hensha256 (commented):
Mostly looking great! Just a few things

test/PoolManager.t.sol
```diff
@@ -982,19 +1002,48 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
+        console2.log(feesAccrued);
+
+        vm.prank(address(feeController));
+        if (balanceToClaim == 0 || balanceToClaim == feesAccrued) {
+            // Expect that they claimed their entire balance
+            manager.collectProtocolFees(protocolFeeRecipient, currency1.toId(), balanceToClaim);
```
snreynolds, line 1030: I think you can move this line up. Same logic regardless of claim amount.

test/PoolManager.t.sol
```diff
@@ -982,19 +1002,48 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
+            assertEq(currency1.balanceOf(protocolFeeRecipient), feesAccrued);
+            assertEq(manager.balanceOf(address(feeController), currency0.toId()), 0);
+        } else if (balanceToClaim < feesAccrued) {
+            // Expect that they claimed some of their balance
+            manager.collectProtocolFees(protocolFeeRecipient, currency1.toId(), balanceToClaim);
+            assertEq(currency1.balanceOf(protocolFeeRecipient), balanceToClaim);
+            assertEq(manager.balanceOf(address(feeController), currency1.toId()), feesAccrued - balanceToClaim);
```
snreynolds, line 1037: could add currency0 balance check here too

snreynolds (changes requested):
small change but lgtm

.gitmodules
```diff
@@ -11,3 +8,6 @@
 [submodule "lib/solmate"]
 	path = lib/solmate
 	url = https://github.com/transmissions11/solmate
+[submodule "lib/forge-std"]
```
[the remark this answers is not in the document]
zhongeric: Mmm version may have changed let me double check

test/PoolManager.t.sol
zhongeric: Ah yes great catch

test/PoolManager.t.sol
zhongeric: yep sounds good

test/PoolManager.t.sol
zhongeric: Wait actually no we can't move it up because we need to expect revert in the case where balance is > claimable

.gitmodules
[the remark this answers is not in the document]
zhongeric: doesn't look like we bumped the submodule, I'll reorder manually

src/PoolManager.sol
zhongeric: Yea I like just having the collect fees func and not exposing burn outside of a lock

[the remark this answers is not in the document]
hensha256: The more i think about it the more that makes sense... the fee controller doesnt care that theyre accruing as 6909s behind the scenes right, they just want to collect the current?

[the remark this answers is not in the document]
zhongeric: Hmm good point. However I do think this relates back to the discussions we had a long time ago about 6909s in general, how we didn't want to write wrappers over all of the functions to enable calling them with `Currency` instead of `uint256` as per the standard. Right now in core we have a clear distinction where ERC20s use Currency (take and settle), and 6906s use the id (mint/burn). Since fees are explicitly accumulated as 6909s instead of ERC20s (they are claims to a balance left in core), I think it makes sense to have this be `id` instead of `currency`. Wdyt?

hensha256 (approved):

src/PoolManager.sol
```diff
@@ -307,6 +311,17 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
...
+        external
+        returns (uint256 amountCollected)
+    {
+        if (msg.sender != address(protocolFeeController)) revert InvalidCaller();
```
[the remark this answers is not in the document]
zhongeric: Yea so because we accrue to the fee controller the owner can't claim without approval. Like feels weird to have that now compared to before when it was in a separate mapping

src/PoolManager.sol
snreynolds: I think we should expose just collect fees then?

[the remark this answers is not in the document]
snreynolds: Hmm right I remember this discussion and we left `mint` and `burn` to have `id`, which would then make sense to make this param an `id` unless we changed all of them to be Currency. I think I don't have a strong preferance either way.. I am forgetting why we felt strongly that mint and burn should be id?

src/PoolManager.sol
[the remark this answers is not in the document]
snreynolds: Technically I think we could allow the owner to pull fees from the protocolFeeController with no approval but that is odd behavior and unexpected/outside the standard... That also would mean the `owner` could always transfer fees from a bricked protocolFeeController to a new one though

src/PoolManager.sol
ewilz: Not a fan of this function if we accumulate in 6909. Let the owner collect fees just like any other 6909 owner. This is basically how V2 works, but theres a utility contract so it's easy to make a governance call.

zhongeric: Oh interesting, like push it to periphery? I think the main rationale was to not make governance handle lock etc. but yea that would work

test/PoolManager.t.sol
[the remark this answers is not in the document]
zhongeric: this is outdated, fixed on latest commit

PR #450, opened 2023-12-18
`payOnBehalf` function

hensha256:
This PR builds on top of #442 

This function allows someone to call `payOnBehalf` to move debt from someone else's deltas to their own. This can be used to
- pay off another address' debt e.g. to pay a fee to a hook
- effectively "send" money to another address by crediting to their delta so that they can call `take`

2 examples are provided in the tests:
- A hook that takes a fee with `take` on the output of a swap. The router can then call `payOnBehalf(hook)` to move that delta to the router's account. The router then can see their net "out" on the swap and `take` it.
- A hook that NoOps swaps, and instead swaps on another pool. This leaves the hook with a debt of input tokens, and a credit of output tokens both of which need to be sent to the user. To credit the output tokens to the user the hook calls `payOnBehalf(router, outputToken)` which moves the tokens owed to the router's delta. The router then calls `payOnBehalf(hook, inputToken)` to pay the input of the swap, and `take(outputToken)` to take out their credited output tokens.


src/test/PoolSwapWithFeeTest.sol
```diff
@@ -0,0 +1,78 @@
...
+        sender = (sender == address(this)) ? data.sender : sender;
+
+        BalanceDelta delta = manager.swap(data.key, data.params, "");
+
+        if (data.params.zeroForOne) {
+            // settle input
+            _settle(data.key.currency0, sender, delta.amount0(), true);
+
+            // settle output and fee
+            manager.acceptDebt(data.key.currency1, address(data.key.hooks), TOTAL_DEBT);
```
saucepoint, line 61: i can get behind this pattern. almost all debt creation will originate from the hook address (hook fees, custom curves)

also it creates a "best practice" where debt _should_ originate from the hook address and not some external contract

---

an alternative would be implementing a way to pay down all debt (regardless of address) but I dont think this is feasible because of the mapping-based accounting

hensha256: yeah thats a nice point about it starting a best practice - i hadnt thought of that.

This function also makes it possible for the hook to "transfer" funds back to the router by having the hook call `manager.acceptDebt(router, x)`. The router will then have a balance of x tokens they can withdraw and give to the user. You can imagine this for a case where the hook performs a swap or withdraws liquidity and wants to give the output tokens to the user. The second use-case `acceptDebt` is called _before_ the debt is "created" so i think we're going to rename the function to `settleFor` or `payOnBehalf` or something like that.

src/PoolManager.sol
```diff
@@ -39,10 +39,12 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
     /// @inheritdoc IPoolManager
     int24 public constant MIN_TICK_SPACING = TickMath.MIN_TICK_SPACING;
 
+    uint256 public constant TOTAL_DEBT = uint256(0);
```
zhongeric, line 42: nit: would appreciate a comment here saying that this is a sentinel value for target's entire delta balance

test/PoolManager.t.sol
```diff
@@ -770,6 +775,38 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
+        vm.etch(hookAddr, address(impl).code);
+
+        (key,) = initPoolAndAddLiquidity(currency0, currency1, IHooks(hookAddr), 100, SQRT_RATIO_1_1, ZERO_BYTES);
+
+        PoolSwapWithFeeTest router = new PoolSwapWithFeeTest(manager);
+        MockERC20(Currency.unwrap(currency0)).approve(address(router), Constants.MAX_UINT256);
+
+        IPoolManager.SwapParams memory params =
+            IPoolManager.SwapParams({zeroForOne: true, amountSpecified: 100, sqrtPriceLimitX96: SQRT_RATIO_1_2});
+        router.swapExactInput(key, params, 95);
```
zhongeric, line 790: can we also add a test where the hook taking a fee causes the amountOut (after paying on behalf) to drop below minAmountOut and make sure that it reverts?

src/test/PoolSwapWithNoOpTest.sol
```diff
@@ -0,0 +1,74 @@
...
+
+        // check that the hook has an input token debt, and this address is owed output tokens
+        require(inputDelta > 0 && outputDelta < 0);
+
+        // move the hook's input debt to this address and then settle it
+        manager.payOnBehalf(currencyIn, address(data.key.hooks), uint256(inputDelta));
+        _settle(currencyIn, sender, inputDelta.toInt128(), true);
+
+        // take the input tokens
+        _take(currencyOut, sender, outputDelta.toInt128(), true);
```
zhongeric, line 70: no check for minAmountOut here?

src/test/PoolSwapWithNoOpTest.sol
```diff
@@ -0,0 +1,74 @@
...
+        int256 outputDelta = manager.currencyDelta(address(this), currencyOut);
+
+        // check that the hook has an input token debt, and this address is owed output tokens
+        require(inputDelta > 0 && outputDelta < 0);
+
+        // move the hook's input debt to this address and then settle it
+        manager.payOnBehalf(currencyIn, address(data.key.hooks), uint256(inputDelta));
+        _settle(currencyIn, sender, inputDelta.toInt128(), true);
+
+        // take the input tokens
```
zhongeric, line 69: nit: output tokens

src/test/PoolSwapWithNoOpTest.sol
```diff
@@ -0,0 +1,74 @@
...
+        Currency currencyOut = data.params.zeroForOne ? data.key.currency1 : data.key.currency0;
+
+        int256 inputDelta = manager.currencyDelta(address(data.key.hooks), currencyIn);
+        int256 outputDelta = manager.currencyDelta(address(this), currencyOut);
+
+        // check that the hook has an input token debt, and this address is owed output tokens
+        require(inputDelta > 0 && outputDelta < 0);
+
+        // move the hook's input debt to this address and then settle it
+        manager.payOnBehalf(currencyIn, address(data.key.hooks), uint256(inputDelta));
```
zhongeric, line 66: do you think a sample implementation would also have a `maxAmountIn` param to protect users from hooks taking more input than required?

zhongeric (commented):
looking good!

src/PoolManager.sol
```diff
@@ -291,6 +297,13 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
         _burnFrom(from, id, amount);
     }
 
+    // Allows one caller to resolve a debt on behalf of another address
+    function payOnBehalf(Currency currency, address caller, uint256 amount) external isLocked {
```
marktoda, line 301: what about `transferDelta`?

src/PoolManager.sol
```diff
@@ -142,9 +144,13 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
     }
 
     function _accountDelta(Currency currency, int128 delta) internal {
+        _accountDeltaForTarget(currency, delta, msg.sender);
+    }
+
+    function _accountDeltaForTarget(Currency currency, int128 delta, address target) internal {
```
marktoda, line 150: what about just `_accountDeltaFor`? think target in name is maybe duplicative

src/PoolManager.sol
marktoda: would max uint be better here?

hensha256: 0 is much cheaper calldata so its what we use [in UR](https://github.com/Uniswap/universal-router/blob/41183d6eb154f0ab0e74a0e911a5ef9ea51fc4bd/contracts/libraries/Constants.sol#L11), and [SwapRouter02](https://github.com/Uniswap/swap-router-contracts/blob/550c0f20373a487996fcc957075377b67af9df07/contracts/libraries/Constants.sol#L8) to signal "the entire value".
However happy to discuss if you have a strong reason for changing!

src/PoolManager.sol
ewilz, line 301: with this function can't the user's `accountDelta` and `balanceDelta` return from a function be out of whack?? seems unideal

PR #462, opened 2024-01-21
move Swap event before afterSwap hook

danrobinson:
## Related Issue

It is possible for afterSwap to do a reentrant call into swap on the same pool (and will be even easier after some potential changes to the locking pattern). If it does, then Swap events could fire out of order (because the second swap, which happens during the afterSwap hook, would finish and fire its event before the event from the first swap), which be confusing for data analysis.

## Description of changes

Moves the Swap event before the afterSwap hook, to prevent misordered Swap events.

### DIFF
--- a/src/PoolManager.sol
+++ b/src/PoolManager.sol
@@ -247,7 +247,6 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
         );
 
         key.hooks.afterSwap(key, params, delta, hookData);
-
     }
 
     /// @inheritdoc IPoolManager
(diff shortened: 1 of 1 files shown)


hensha256 (dismissed):
nice find

hensha256:
oh 2 things
- please can you sign your commit! this repo cant merge unsigned commits
- please can you run the linter (`forge fmt` or `just fix` depending which setup youre using`)

saucepoint:
do we want to do the same for modifyLiquidity?

danrobinson:
Good catch—done!

hensha256:
initialize too! Also weird that we dont have an event emitted for Donate?...want to add one?

rondales (approved):

Issue #346, opened 2023-09-01
Allow donate() to specify tick


danrobinson:
### Component

PoolManager/Pool

### Describe the suggested feature and problem it solves.

Right now, donate() can donate additional fees to currently in-range liquidity provides. But there is no way to donate fees to out-of-range liquidity providers—specifically liquidity providers who would be in range if liquidity was at a different tick.

The most common use cases would be for hooks that want to distribute some kind of rent to liquidity providers who have not been traded against (which is currently impossible), or who want to distribute fees or rent to liquidity providers after a trade in some way other than proportional to volume (which would require splitting the swap up and donating after each step).

Some examples:

* In its current implementation, TWAMM charges fees on net volume rather than gross. But if TWAMM wanted to charge gross fees and distribute those to the liquidity providers who would have been in range when those swaps were executed, it would need something like this.
* Designs for ex ante LVR-capturing auctions (like mcAMM) may want to distribute rents not proportional to actual volume traded, but proportional to the option value given up by liquidity providers. That may even involve distributing rent to liquidity providers who were not swapped against.
* Some other LVR-capturing designs might want to disproportionately pay fees to the first LPs swapped against, since otherwise they suffer much more LVR than the later LPs (because they trade at a worse price).

### Describe the desired implementation.

Change `donate` signature to:

```
function donate(PoolKey memory key, uint256 amount0[], uint256 amount1[], int24[] ticks, bytes calldata hookData)
```

`amount0`, `amount1`, and `ticks` would all have to be arrays of the same length. `ticks` would indicate the buckets into which the fees should be donated (identified by the tick immediately below them). `ticks` must be sorted from low to high (reverts if it is not).

#### Algorithm

* First, compute the liquidity that would be in range at `ticks[0]` by walking down to it.
  * Start with `liquidityAtTick = liquidityCurrent`.
  * Starting at the current tick, walk down all initialized ticks until reaching `ticks[0]`, applying `liquidityNet` to `liquidityAtTick`. The last tick applied should be the last one to the right of `ticks[0]`.
* Next, walk back over all initialized ticks to the current tick, tracking liquidity and cumulative fee growth and updating the feeGrowthOutside values on each tick as you go. 
  * Once you've passed the current tick, stop. Add the `cumulativeFeeGrowth`s to the respective `feeGrowthGlobal`s in the current state.
[...]

PR #465, opened 2024-01-27
Add Simplified Multi-tick donate

Philogy:
## Related Issue

#346

## Description of changes

Implements a simplified multi-tick donate function. This is still a prototype therefore the DRAFT status. Many TODOs left for productionization:
- [ ] Remove console logs
- [ ] More tests
- [ ] End tick validation?
- [ ] Split donate into donate above/below? separate hooks?


snreynolds (commented):
whats the bytecode change?

PR #468, opened 2024-01-31
Add Tests for maxUsableTick and minUsableTick Functions

partylikeits1983:
This commit introduces tests for the newly added utility functions `maxUsableTick` and `minUsableTick` in the TickMath library.

Added two tests: 
1. `test_maxUsableTick`: Validates the maximum usable tick calculation for different tick spacings. It ensures the function returns the correct maximum tick value based on the given tick spacing.

2. `test_minUsableTick`: Similar to the max test, this checks the accuracy of the minimum usable tick calculation.

After adding these two tests, code coverage for functions in src/libraries/TickMath.sol is now 100%, before it was 50%.

Please let me know if I need to open a new PR without the changes to .forge-snapshots directory.


test/TickMath.t.sol
```diff
@@ -44,6 +44,20 @@ contract TickMathTestTest is Test, JavascriptFfi {
         assertEq(maxTick, MAX_TICK);
     }
 
+    function test_maxUsableTick() public {
+        assertEq(tickMath.maxUsableTick(1), tickMath.MAX_TICK());
```
aadams, line 48: these numbers feel a bit random. these are the fees in v3, but not the tick-spacings.

maybe we could add a few more?

test/TickMath.t.sol
```diff
@@ -44,6 +44,20 @@ contract TickMathTestTest is Test, JavascriptFfi {
...
+        assertEq(tickMath.maxUsableTick(500), 887000);
+        assertEq(tickMath.maxUsableTick(3000), 885000);
+        assertEq(tickMath.maxUsableTick(10000), 880000);
+    }
+
+    function test_minUsableTick() public {
+        assertEq(tickMath.minUsableTick(1), tickMath.MIN_TICK());
+        assertEq(tickMath.minUsableTick(500), -887000);
```
aadams, line 56: would be nice to add something that minUsableTick() = - maxUsableTick() as well

hensha256:
hey @partylikeits1983 ! any chance of responding to the comments above please? 🙏

PR #480, opened 2024-02-13
Remove EOA calling capability

hensha256:
Removing the ability to call `lock` from an EOA. After #442 the gas benefits of this are only ~1000, and there is an additional overhead for contract-callers of 300-500 gas.
The main reason for removing is that this feature removes the ability for the contract implementing `lockAcquired` to have guarantees that they passed in the lock data themselves. This means that extra data validation will be needed, and its more likely for lockers with bugs to be coded.

Built on top of #442 - that needs to be merged first


src/libraries/Pool.sol
```diff
@@ -204,9 +205,20 @@ library Pool {
                 }
             }
         }
+        console2.log("0");
```
ewilz, line 208: consoles

ewilz:
looking good to me, but obviously a lot of merge conflicts, can approve after those are addressed. I think another reason this pattern is no longer great is bc the first caller has special privileges to use an outside callback contract, whereas any subsequent caller wouldn't. Feels weird. + assuming most calls come through an aggregator or router anyway...

ewilz (approved):

Issue #467, opened 2024-01-29
Flip the sign on currencyDeltas and accountDelta


danrobinson:
### Component

Delta accounting

### Describe the suggested feature and problem it solves.

This is a change to conventions to make singleton accounting more consistent and intuitive.

The singleton tracks transient balances for users in a [currencyDelta](https://github.com/Uniswap/v4-core/blob/8109ec3c2f9db321ba48fff44ed429e6c1bd3eb3/src/PoolManager.sol#L45) mapping. It tracks changes to those transient balances using the [BalanceDelta](https://github.com/Uniswap/v4-core/blob/8109ec3c2f9db321ba48fff44ed429e6c1bd3eb3/src/types/BalanceDelta.sol) type.

Right now, these are all signed integers, reflecting the fact that unlike ERC-20 or ERC-6909 balances, transient balances and deltas can be temporarily negative as long as they are resolved at the end of the transaction.

But separately from this, `currencyDelta` also flips the usual sign for tracking balances. If the user deposits more money into the singleton, or if some pool or other account transfers money to the user, the user's transient balance on the singleton goes __down_. When the user owes the pool money, they have a positive `currencyDelta` on the pool.

This is inconsistent with how the singleton accounts for ERC-6909 balances (resulting in the odd interaction that when a user converts transient balance into ERC-6909 balance using [mint](https://github.com/Uniswap/v4-core/blob/8109ec3c2f9db321ba48fff44ed429e6c1bd3eb3/src/PoolManager.sol#L291), both of its balances go up). It is also inconsistent with how essentially all ERC-20 contracts track balances of users (where a positive balance means the user has money rather than owes money). Finally, it is counterintuitive.

Since `BalanceDeltas` are directly applied to `currencyDelta`, they also follow these flipped conventions—functions return negative numbers for BalanceDeltas when a user should receive money as a result of the function, and positive numbers when a user should pay money as a result of the function.

### Describe the desired implementation.

Flip the meaning of `currencyDelta` and `balanceDelta`, and flip the sign on every "delta" at the moment it is first created.

### Additional context.

One reason to consider keeping the current convention is that the convention for positive deltas meaning the user owes money to the pool was set by the callbacks in Uniswap v3 (such as https://github.com/Uniswap/v3-core/blob/main/contracts/UniswapV3Pool.sol#L482). 

However, I think that was a very different context and different functionality. Uniswap v3 did not have a full accounting system for user deltas, so it was freer to make a different choice.

[...]

Issue #390, opened 2023-11-07
Test suite: Improve Deployers/Set up Flow for pools


snreynolds:
### Component

General design optimization (improving efficiency, cleanliness, or developer experience)

### Describe the suggested feature and problem it solves.

Deployers contract that we use in the testing suite has gotten a little bloated. I think we should clean this up and have a standard way of approving tokens and creating a new pool, and then enforce a standard way to perform all operations on top of that: adding liquidity/swapping/donating.

hensha256:
Things I'd like to see improved
- Lots of asserts missing on all `swap` tests - like the amount actually swapped
- Helper function to set up a pool, approve the modifyPositionRouter for tokens, and give the pool initial liquidity

hensha256:
- Add explicit `settle` testing
- Improve dynamic fee tests - currently they dont explicitly make it clear that the fee is changing with asserts

hensha256:
- Update all uses of `vm.expectRevert()` to specify the selector of the error it should be throwing

hensha256:
- Proper tests for modify position!

hensha256:
- Naming the same 
- Separate gas tests

snreynolds:
Added these notes as checkboxes above

zhongeric:
FYI we are splitting up modifyPosition tests with work in #402 so hopefully it'll be more fine grained

hensha256:
Add testing for interactions of NoOp #324 and Access Lock #404 once they are both merged.
For example testing a nested call to swap where the inner call NoOps

hensha256:
Test all functions revert without a lock

hensha256:
Remove `vm.assume` and replace with `bound` _or_ actually test the other values revert !

hensha256:
Make a common `SWAP_PARAMS` in Deployers.sol instead of having 
`IPoolManager.SwapParams({zeroForOne: true, amountSpecified: -100, sqrtPriceLimitX96: SQRT_RATIO_1_2})`
copied everywhere

hensha256:
helper functions `_settle` and `_take` could take a uint instead of an int

PR #481, opened 2024-02-14
Flip the sign of deltas

hensha256:
closes https://github.com/Uniswap/v4-core/issues/467

The current set up is counter intuitive.

After this pull request:
- as positive delta means that the address can withdraw money from the pool manager
- withdrawing money will subtract from the delta.
- a negative delta means that the address owes the pool manager
- sending money into the PM will add to the delta.


src/PoolManager.sol
```diff
@@ -227,16 +227,16 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
...
+        // the fee is on the input currency
         unchecked {
             if (feeForProtocol > 0) {
                 protocolFeesAccrued[params.zeroForOne ? key.currency0 : key.currency1] += feeForProtocol;
             }
         }
 
         emit Swap(
-            id, msg.sender, delta.amount0(), delta.amount1(), state.sqrtPriceX96, state.liquidity, state.tick, swapFee
+            id, msg.sender, -delta.amount0(), -delta.amount1(), state.sqrtPriceX96, state.liquidity, state.tick, swapFee
```
ewilz, line 239: why is the events delta the opposite as the returned delta?

ewilz:
I realize this is potentially a can of worms w swap math + backwards compatibility, but does a negative amountSpecified make more sense to denote amountIn now (negative for the user)?

src/PoolManager.sol
hensha256: was doing this to align it with amoutnSpecified - which we are now going to fix in #491

src/libraries/Pool.sol
```diff
@@ -248,7 +248,7 @@ library Pool {
         }
 
         // Fees earned from LPing are removed from the pool balance.
```
snreynolds, line 250: Maybe we could make this comment more clear now... 
// Fees earned from LPing are added to the user's collectable amount.

src/libraries/Pool.sol
```diff
@@ -210,7 +210,7 @@ library Pool {
                 // current tick is below the passed range; liquidity can only become in range by crossing from left to
                 // right, when we'll need _more_ currency0 (it's becoming more valuable) so user must provide it
                 result = result
```
snreynolds, line 212: Few notes here... From the code prior (not your changed code) I'm wondering why we need to add to result to begin with? result is not changed before this block of code.

Secondly, that made me think its odd to change the sign here and that technically it seems right to have `SqrtPriceMath.getAmount0Delta` and `SqrtPriceMath.getAmount1Delta` return the correctly signed delta especially since it is originally set there. Feels odd to set it incorrectly and then change it again here

src/libraries/Pool.sol
```diff
@@ -444,16 +444,17 @@ library Pool {
             self.feeGrowthGlobal1X128 = state.feeGrowthGlobalX128;
         }
 
+        // in an unchecked block so sign flipping is not safe without a helper function
         unchecked {
             if (params.zeroForOne == exactInput) {
                 result = toBalanceDelta(
-                    (params.amountSpecified - state.amountSpecifiedRemaining).toInt128(),
-                    state.amountCalculated.toInt128()
+                    (params.amountSpecified - state.amountSpecifiedRemaining).toInt128().flipSign(),
```
snreynolds, line 451: Hm.. Since we are flipping amountSpecified this will not be needed correct?

snreynolds: Like imo all the math should be changed to return the correct sign and something like flipSign shoudn't be needed

src/test/PoolTestBase.sol
```diff
@@ -17,25 +18,25 @@ abstract contract PoolTestBase is ILockCallback {
     }
 
     function _take(Currency currency, address recipient, int128 amount, bool withdrawTokens) internal {
-        assert(amount < 0);
+        assertGt(amount, 0);
```
snreynolds, line 21: how come take just doesn't take in `uint128`

src/test/PoolTestBase.sol
```diff
@@ -17,25 +18,25 @@ abstract contract PoolTestBase is ILockCallback {
...
+            manager.take(currency, recipient, uint128(amount));
         } else {
-            manager.mint(recipient, currency.toId(), uint128(-amount));
+            manager.mint(recipient, currency.toId(), uint128(amount));
         }
     }
 
     function _settle(Currency currency, address payer, int128 amount, bool settleUsingTransfer) internal {
-        assert(amount > 0);
+        assertLt(amount, 0);
```
snreynolds, line 30: same here why not just have this be a positive amount?

it will always be negative passed in but then its always flipped again below. i think itd be much easier to read if it was just always positive

test/utils/SwapHelper.t.sol
```diff
@@ -41,31 +41,31 @@ contract SwapHelperTest is Test, Deployers, GasSnapshot {
...
         BalanceDelta result = swap(key, true, amountSpecified, ZERO_BYTES);
-        assertEq(int256(result.amount1()), amountSpecified);
+        assertEq(-int256(result.amount1()), amountSpecified);
     }
 
     function test_swap_helper_oneForZero_exactInput() public {
         int256 amountSpecified = 100;
         BalanceDelta result = swap(key, false, amountSpecified, ZERO_BYTES);
-        assertEq(int256(result.amount1()), amountSpecified);
+        assertEq(-int256(result.amount1()), amountSpecified);
```
snreynolds, line 56: yay these will all be consistent w emilys changes!

snreynolds (changes requested):

src/libraries/Pool.sol
hensha256: Right yeah it gets fixed in emily's PR if you can review that too! 🙌 Then we can merge that into this PR

src/libraries/Pool.sol
hensha256: imo its actually a badly named function... the function returns a `uint256` thats described as
"amount0 Amount of currency0 required to cover a position of size liquidity between the two passed prices"
so its not a delta at all, its just an amount

So maybe it would make more sense to rename it `getAmount0`?

Otherwise I can do as youre saying - but we will need to change the return type of the function to actually make it a delta as it currently isnt one

Thoughts?

hensha256: > From the code prior (not your changed code) I'm wondering why we need to add to result to begin with? result is not changed before this block of code.

Agreed, updating

hensha256: > Secondly, ...
Also done

src/test/PoolTestBase.sol
hensha256: yeah right now it just takes in an `int128` so that the tests can pass them in as like `delta.amount0()` etc without casting

src/PoolManager.sol
```diff
@@ -259,7 +259,8 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
 
     /// @inheritdoc IPoolManager
     function take(Currency currency, address to, uint256 amount) external override noDelegateCall isLocked {
-        _accountDelta(currency, amount.toInt128());
+        // flipping a postive number in an int is always safe
```
ewilz, line 262: I find this comment confusing. I take it as saying:" flipping a uint is implicitly safe", but in fact we're using a safeCast library, so it's not actually implicitly safe and we need to make it safe. I think using the library function here is self explanatory and the comment is causing confusion. ??

src/PoolManager.sol
```diff
@@ -269,19 +270,19 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
...
         paid = reservesOf[currency] - reservesBefore;
-        // subtraction must be safe
-        _accountDelta(currency, -(paid.toInt128()));
+        _accountDelta(currency, paid.toInt128());
     }
 
     /// @inheritdoc IPoolManager
     function mint(address to, uint256 id, uint256 amount) external override noDelegateCall isLocked {
-        _accountDelta(CurrencyLibrary.fromId(id), amount.toInt128());
+        // flipping a postive number in an int is always safe
```
ewilz, line 278: same here. "Must be safe" (the old comment)is less confusing if we do want to keep a comment.

src/test/PoolSwapTest.sol
```diff
@@ -9,6 +9,7 @@ import {IHooks} from "../interfaces/IHooks.sol";
 import {Hooks} from "../libraries/Hooks.sol";
 import {PoolTestBase} from "./PoolTestBase.sol";
 import {Test} from "forge-std/Test.sol";
+import {console2} from "forge-std/console2.sol";
```
ewilz, line 12: console left here 🤷 only a test file though lol.

src/libraries/Pool.sol
```diff
@@ -444,6 +441,7 @@ library Pool {
             self.feeGrowthGlobal1X128 = state.feeGrowthGlobalX128;
         }
 
+        // in an unchecked block so sign flipping is not safe without a helper function
```
ewilz, line 444: again, this feels over-commented to me, using the helper function is self explanatory

src/test/PoolTestBase.sol
ewilz: good call. same for settle.

test/utils/SwapHelper.t.sol
```diff
@@ -39,98 +39,98 @@ contract SwapHelperTest is Test, Deployers, GasSnapshot {
...
-        assertEq(int256(result.amount0()), amountSpecified);
+        assertEq(result.amount0(), amountSpecified);
     }
 
     function test_swap_helper_zeroForOne_exactOutput() public {
-        int256 amountSpecified = -100;
+        int256 amountSpecified = 100;
         BalanceDelta result = swap(key, true, amountSpecified, ZERO_BYTES);
-        assertEq(int256(result.amount1()), amountSpecified);
+        assertEq(result.amount1(), amountSpecified);
```
ewilz, line 50: haha beautiful change

ewilz (dismissed):
approved with some nits about code comments

src/PoolManager.sol
hensha256: Oh what my comment is trying to refer to is the `-` not the `.toInt128`. Its this scenario:
- you have an int128 `x`, and you know that `x>0`
- its completely safe do just do `int128 y = -x`

whereas if you have an int128 `x` and you know that `x<0`, for this case it is NOT save to do `int128 y = -x`. This is because the lowest possible negative number in an int type, does not have a corresponding positive value.

that is to say `type(int128).min * -1 = type(int128).max + 1` so you cannot safely flip in 1 direction, but you can in the other direction

idk how to put that in a single comment though 😂 i'll try again

hensha256: just went with `subtraction must be safe`

src/libraries/SwapMath.sol
```diff
@@ -74,13 +74,13 @@ library SwapMath {
...
-            if (!exactIn && amountOut > uint256(-amountRemaining)) {
-                amountOut = uint256(-amountRemaining);
+            if (!exactIn && amountOut > uint256(amountRemaining)) {
+                amountOut = uint256(amountRemaining);
             }
 
             if (exactIn && sqrtRatioNextX96 != sqrtRatioTargetX96) {
                 // we didn't reach the target, so take the remainder of the maximum input as fee
-                feeAmount = uint256(amountRemaining) - amountIn;
+                feeAmount = uint256(-amountRemaining) - amountIn;
```
snreynolds, line 83: I think I'm not 100% understanding this line. If we have't reached the target the feeAmount is the amount left to swap minus the amount already swapped?

Also why does this case only happen in the exactIn case?

src/test/PoolTestBase.sol
snreynolds: ya agreed. is there an issue open for this? Think this is a strictly good improvement. Want to make sure it doesnt get lost.

test/SqrtPriceMath.t.sol
```diff
@@ -249,7 +249,7 @@ contract SqrtPriceMathTestTest is Test, GasSnapshot {
         assertEq(amount0, 0);
     }
 
-    function test_getAmount0Delta_returns0_1Amount1ForPriceOf1To1_21() public {
+    function test_getAmount0Delta_1Amount1ForPriceOf1To1_21() public {
```
snreynolds, line 252: hm I know not your change but what is 1Amount1? Not really sure what this is testing/what the change to the name means

src/test/PoolTestBase.sol
hensha256: its in the test improvements issue #390

test/SqrtPriceMath.t.sol
hensha256: It tests getAmount0Delta, with an amount1 of 1, and a price of SQRT_RATIO_121_100.
I removed the `return0` because it doesnt test that (i think it was copied from the test name above)

src/libraries/SqrtPriceMath.sol
```diff
@@ -200,8 +200,8 @@ library SqrtPriceMath {
     {
         unchecked {
             return liquidity < 0
```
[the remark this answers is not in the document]
hensha256: Right the argument is that `liquidityDelta` is the user's delta of _liquidity units_ not token units. When they pass in `liquidityDelta = 100` its because they want their liquidity balance to increase by 100 (which means sending tokens into the contract). So `liquidityDelta` is in line with `amountSpecified` now... for both of them:
- positive: the balance of the thing in question goes up
- negative: the balance of the thing in question goes down

src/libraries/SwapMath.sol
hensha256: > If we have't reached the target the feeAmount is the amount left to swap minus the amount already swapped?

If we havent reached the target (which is the essentially maximum slippage where the trade has to stop), that means that all of the `amountRemaining` has successfully be traded into output tokens. The case this happens, the function follows the following flow...
- calculate `amountRemainingLessFee` which is just the amount to be traded, minus the LP fee
- calculates the `sqrtRatioNextX96` that does get reached by this amount of input tokens
- calculate the exact `amountIn` needed to trade to that next sqrtRatio using `getAmount0Delta`. Due to rounding of calculations this could be slightly different to the `amountRemainingLessFee` that it was going to trade before
- now the true LP fee amount needs to be calculated, which is the difference between the total amount to be traded `amountRemaining` (which is now negative so needs to be flipped), and `amountIn` which is the amount thats actually being swapped. This means that the fee is rounded in the LP's favour.

> Also why does this case only happen in the exactIn case?

LP fees are taken on input tokens. For exactIn, the calculation makes sure that the final amountIn including fees is amountRemaining. For non-exactIn the fee amount can just be calculated as `feePips`% of the amountIn, again rounded in the LP's favour. Theres not a need to make sure that the total `feeAmount + amountIn == amountRemaining` because its not exactIn so the "exact" part stops mattering

test/SqrtPriceMath.t.sol
snreynolds: sounds good

snreynolds (approved):

Issue #477, opened 2024-02-12
Remove lock from initialize


hensha256:
### Describe the suggested feature and problem it solves.

After #442 , the lock on initialize is no longer needed. This includes updating how all our tests initialize the pool - PoolInitializeTest.sol is no longer needed

PR #489, opened 2024-02-28
Remove lock from initialize

dianakocsis:
## Related Issue
Remove lock from initialize #477

## Description of changes
- Removes the `isLocked` modifier on the `initialize` function in PoolManager.sol and updates the corresponding tests


snreynolds (approved):
Looks good! Can you add a test initializing within a lock?

Feel free to add that in a followup after #442 gets merged since there is a new nested lock test file

src/test/PoolNestedActionsTest.sol
```diff
@@ -211,6 +216,19 @@ contract NestedActionExecutor is Test, PoolTestBase {
         _settle(key.currency1, user, int128(deltaThisAfter1), true);
     }
 
+    function _initialize() internal {
+        address locker = manager.getLocker();
+        assertTrue(locker != address(this), "Locker wrong");
+        PoolId id = key.toId();
+        (uint256 price,,) = manager.getSlot0(id);
+        assertEq(price, Constants.SQRT_RATIO_1_1);
```
hensha256, line 224: I think you can remove lines 222-224 given they are checking things about another pool. Better to have it check the pool in question before and after
- so before check that the price is 0
- and after check (as you have) that its `SQRT_RATIO_1_2`

PR #491, opened 2024-02-29
POC: amountSpecified matches deltas

ewilz:
## Related Issue

## Description of changes


test/PoolManager.t.sol
```diff
@@ -464,7 +464,7 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
         modifyLiquidityRouter.modifyLiquidity{value: 1 ether}(nativeKey, liqParams, ZERO_BYTES);
 
         IPoolManager.SwapParams memory params =
-            IPoolManager.SwapParams({zeroForOne: true, amountSpecified: 100, sqrtPriceLimitX96: SQRT_RATIO_1_2});
+            IPoolManager.SwapParams({zeroForOne: true, amountSpecified: -100, sqrtPriceLimitX96: SQRT_RATIO_1_2});
```
hensha256, line 467: This identical struct is used all over this file, and DynamicFees.t.sol. I wonder if we should just save it as our default `SWAP_PARAMS` in `Deployers.sol` like we have `LIQ_PARAMS` etc in there. And then al these files can just use that constant defined in 1 place?

hensha256: or `SWAP_EXACT_IN_PARAMS` or something

hensha256 (commented):
Looking great, much simpler than i thought it was going to be lol

test/PoolManager.t.sol
hensha256: I added this to our #390

ewilz: good call

Issue #487, opened 2024-02-27
store boolean for lock


hensha256:
### Describe the suggested feature and problem it solves.

After #442 we can just remove storing the address completely and store it as a boolean?

PR #493, opened 2024-03-01
store boolean instead of address

dianakocsis:
## Related Issue
store boolean for lock #487

## Description of changes
Removes storing the locker address and now stores a boolean instead


src/PoolManager.sol
```diff
@@ -88,11 +88,6 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
         return pools[id].positions.get(_owner, tickLower, tickUpper);
     }
 
-    /// @inheritdoc IPoolManager
-    function getLocker() external view override returns (address locker) {
```
hensha256, line 92: Instead of `getLocker` we should probably expose `isLocked` here so people can see if the contract is locked or not

src/libraries/Locker.sol
```diff
@@ -14,31 +14,25 @@ library Locker {
     /// we use locker==address(0) to signal that the pool is not locked
     error InvalidLocker();
 
-    function setLocker(address locker) internal {
-        if (locker == address(0)) revert InvalidLocker();
+    function lock() internal {
         uint256 slot = LOCKER_SLOT;
```
hensha256, line 18: nit: we should rename `LOCKER_SLOT`... maybe `LOCKED_SLOT` or ... `LOCK_SLOT` ? Open to other ideas

src/test/PoolNestedActionsTest.sol
```diff
@@ -46,14 +46,8 @@ contract PoolNestedActionsTest is Test, ILockCallback {
     }
 
     function _nestedLock() internal {
-        address locker = manager.getLocker();
-        assertEq(locker, address(this));
```
hensha256, line 50: I think we should keep these lines - just checking instead that `manager.isLocked() == true` twice in this test (before and after the revert)

test/Locker.t.sol
```diff
@@ -8,38 +8,14 @@ contract LockerTest is Test {
...
-        assertEq(address(Locker.getLocker()), locker);
-
-        Locker.clearLocker();
-
-        assertEq(address(Locker.getLocker()), address(0));
-    }
-
-    function test_fuzz_isLocked(address locker) public {
-        vm.assume(locker != address(0));
+    function test_fuzz_lock() public {
```
hensha256, line 11: this is no longer a fuzz test - fuzz tests are when the test has a parameter like `address locker` and forge will choose 100,000 different values for that parameter and check that the test passes for all of them. Now that we dont have a parameter this can just be `test_lock`

hensha256 (changes requested):
Looking good!! Just a couple of little things 😄 Nice to see the gas went down a bit with it too

src/interfaces/IPoolManager.sol
```diff
@@ -120,6 +120,9 @@ interface IPoolManager is IFees, IERC6909Claims {
     /// @notice Returns the reserves for a given ERC20 currency
     function reservesOf(Currency currency) external view returns (uint256);
 
+    /// @notice Returns whether the contract is locked
+    function isLockSet() external view returns (bool);
```
hensha256, line 124: mm my instinct is that i prefer `isLocked` for this. Although we'll be renaming it in a later PR anyway so it doesnt really matter 😆

hensha256 (commented):
Just a couple more things! We're nearly there!

src/libraries/Locker.sol
snreynolds: I like LOCK_SLOT

src/interfaces/IPoolManager.sol
snreynolds: this function will be removed when we have extsload also just a note

snreynolds: so its probably nbd rn and since we have isLocked already I think its a fine temp name

hensha256: yeah good point lets just leave it as is

hensha256 (approved):
lgtm!

src/test/PoolNestedActionsTest.sol
```diff
@@ -92,29 +90,26 @@ contract NestedActionExecutor is Test, PoolTestBase {
...
-        assertEq(locker, msg.sender);
+        locked = manager.isLockSet();
+        assertEq(locked, true);
     }
 
-    function _swap() internal {
-        (address locker) = manager.getLocker();
-        assertTrue(locker != address(this), "Locker wrong");
-
+    function _swap(address locker) internal {
```
snreynolds, line 110: super nit but this really is now the caller right? Seems like we renamed (which I like better!)

snreynolds (dismissed):
looks good! one tiny nit

snreynolds (approved):
lgtm!

PR #495, opened 2024-03-05
Make currency delta transient

hensha256:


src/PoolManager.sol
```diff
@@ -144,7 +145,7 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
     function _accountDelta(Currency currency, int128 delta) internal {
         if (delta == 0) return;
 
-        int256 current = currencyDelta[msg.sender][currency];
+        int256 current = msg.sender.getCurrencyDelta(currency);
```
snreynolds, line 148: this syntax is kinda funky but I do like it as a library 🤷‍♀️

snreynolds (approved):

src/libraries/CurrencyDelta.sol
```diff
@@ -0,0 +1,40 @@
...
+pragma solidity ^0.8.20;
+
+import {Currency} from "../types/Currency.sol";
+
+library CurrencyDelta {
+    uint256 constant CURRENCY_DELTA_SLOT = uint256(keccak256("CurrencyDelta")) - 1;
+
+    function setCurrencyDelta(address locker, Currency currency, int256 delta) internal {
+        uint256 slot = CURRENCY_DELTA_SLOT;
+
```
snreynolds, line 11: actually could you potentially write a helper called hashKey = _computeKey(locker, currency)? this whole chunk is repeated in both set/get

snreynolds (changes requested):

src/PoolManager.sol
hensha256: Yeah i wasnt 100% on it either. Would you prefer it as
`CurrencyDelta.get(msg.sender, currency)` and `CurrencyDelta.set(msg.sender, currency, delta)`
or
`msg.sender.getCurrencyDelta(currency)` and `msg.sender.setCurrencyDelta(currency, delta)`

src/libraries/CurrencyDelta.sol
```diff
@@ -0,0 +1,38 @@
...
+import {Currency} from "../types/Currency.sol";
+
+library CurrencyDelta {
+    uint256 constant CURRENCY_DELTA_SLOT = uint256(keccak256("CurrencyDelta")) - 1;
+
+    function _computeSlot(address locker, Currency currency) internal pure returns (bytes32 hashSlot) {
+        uint256 slot = CURRENCY_DELTA_SLOT;
+
+        assembly {
```
ewilz, line 12: why is this done in assembly? cheaper?

hensha256: Yeeah its quite a lot cheaper. And its like another 1000 cheaper if instead we do 
```
        assembly {
            mstore(0, locker)
            mstore(32, currency)
            hashSlot := keccak256(0, 64)
        }
```

with a single hash and not have a `CURRENCY_DELTA_SLOT` thats added into the process with a second hash. But I feel like its definitely best practice to add that just for if we ever had another mapping with these 2 parameters too. Thoughts?

src/PoolManager.sol
hensha256: took a poll and went with the first one!

src/libraries/CurrencyDelta.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.20;
+
+import {Currency} from "../types/Currency.sol";
+
+library CurrencyDelta {
+    uint256 constant CURRENCY_DELTA_SLOT = uint256(keccak256("CurrencyDelta")) - 1;
+
+    function _computeSlot(address locker, Currency currency) internal pure returns (bytes32 hashSlot) {
```
dianakocsis, line 9: Do we still want this to be called locker?

hensha256: oh yeah good point

src/libraries/CurrencyDelta.sol
snreynolds: This though is not standard for computing the mapping slot?

src/libraries/CurrencyDelta.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.20;
+
+import {Currency} from "../types/Currency.sol";
+
+library CurrencyDelta {
+    uint256 constant CURRENCY_DELTA_SLOT = uint256(keccak256("CurrencyDelta")) - 1;
```
snreynolds, line 7: Instead of this computing a hash every time we can just use the hex val and comment above?

hensha256: oh yeeaaah nice haha, i made that PR after this

src/libraries/CurrencyDelta.sol
hensha256: Yeah the standard way of computing a mapping is including a slot

snreynolds (approved):

PR #497, opened 2024-03-06
added helper function computeKey()

nishim3:
## Related Issue
Which issue does this pull request resolve?
Added a helper function _computeKey according to [this](https://github.com/Uniswap/v4-core/pull/495#discussion_r1513737654
) suggestion.
## Description of changes

PR #498, opened 2024-03-06
add _computeSlot helper

nishim3:
Follow up from #497 . 
Added a helper function _computeKey according to https://github.com/Uniswap/v4-core/pull/495#discussion_r1513737654 suggestion.


src/libraries/CurrencyDelta.sol
```diff
@@ -6,34 +6,31 @@ import {Currency} from "../types/Currency.sol";
...
             mstore(0, locker)
             mstore(32, slot)
             let intermediateHash := keccak256(0, 64)
 
             mstore(0, intermediateHash)
             mstore(32, currency)
-            let hashSlot := keccak256(0, 64)
+            hashSlot := keccak256(0, 64)
+        }
+        return hashSlot;
```
hensha256, line 20: you can get rid of this line now that its named in the return parameter

nishim3: Done!

hensha256 (approved):

PR #499, opened 2024-03-07
remove lock on initialize

dianakocsis:
## Related Issue
Remove lock from initialize #477

## Description of changes
- Removes the `isLocked` modifier on the `initialize` function in PoolManager.sol and updates the corresponding tests


hensha256 (approved):
🚀

Issue #469, opened 2024-01-31
Make updateDynamicSwapFee "push" rather than "pull", and remove event


danrobinson:
### Component

Gas Optimization

### Describe the suggested feature and problem it solves.

This is a gas optimization for `updateDynamicSwapFee`.

Currently, `updateDynamicSwapFee` is open to anyone and makes a call to `_fetchDynamicSwapFee` on the hook contract. This seems like a complicated architecture that wastes an extra call (in the potentially common case where the hook contract wants to update the swap fee during another function, such as in its `beforeSwap`).

It also emits a `DynamicSwapFeeUpdated` event. This seems like a waste of gas, particularly because the current swap fee can always be read on the pool itself and all changes to it will already be indexed in `traces` by looking at calls to the `updateDynamicSwapFee` function (or at least, they will once we turn it into a push model rather than pull).

### Describe the desired implementation.

Change [updateDynamicSwapFee](https://github.com/Uniswap/v4-core/blob/main/src/PoolManager.sol#L310) to something like:

```solidity
    function updateDynamicSwapFee(PoolKey memory key, uint24 newDynamicSwapFee) external {
        if (key.fee.isDynamicFee()) {
            if (msg.sender != address(key.hooks)) revert FeeUpdateUnauthorized();
            if (dynamicSwapFee >= MAX_SWAP_FEE) revert FeeTooLarge();
            PoolId id = key.toId();
            pools[id].setSwapFee(newDynamicSwapFee);
        } else {
            revert FeeNotDynamic();
        }
    }
```

PR #501, opened 2024-03-08
Push dynamic fees

hensha256:
## Related Issue
Which issue does this pull request resolve? #469 
Pushing of dynamic fees instead of pulling
Plus a few other refactor updates of swap and protocol fees to make them neater


src/PoolManager.sol
```diff
@@ -105,19 +104,23 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
...
-        if (key.fee.isStaticFeeTooLarge()) revert FeeTooLarge();
-
         // see TickBitmap.sol for overflow conditions that can arise from tick spacing being too large
         if (key.tickSpacing > MAX_TICK_SPACING) revert TickSpacingTooLarge();
         if (key.tickSpacing < MIN_TICK_SPACING) revert TickSpacingTooSmall();
         if (key.currency0 >= key.currency1) revert CurrenciesOutOfOrderOrEqual();
         if (!key.hooks.isValidHookAddress(key.fee)) revert Hooks.HookAddressNotValid(address(key.hooks));
 
+        uint24 swapFee = 0;
+        // Initial swap fee on dynamic fee pools is set to 0, unless updateDynamicSwapFee is called in the afterInitialize hook
```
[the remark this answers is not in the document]
saucepoint: i guess the other option is to add another parameter `initialDynamicFee`

It's a bit ugly and might be unused the majority of the time, but it would make dynamic-fee-pool-creation cheaper? 

leaning slightly towards the extra parameter because the thought of an afterInitialize call seems like a bunch of overhead for a simple operation

src/libraries/SwapFeeLibrary.sol
```diff
@@ -0,0 +1,29 @@
...
+    error FeeTooLarge();
+
+    uint24 public constant STATIC_FEE_MASK = 0x0FFFFF;
+    uint24 public constant DYNAMIC_FEE_FLAG = 0x800000;
+
+    // the swap fee is represented in hundredths of a bip, so the max is 100%
+    uint24 public constant MAX_SWAP_FEE = 1000000;
+
+    function isDynamicFee(uint24 self) internal pure returns (bool) {
+        return self & DYNAMIC_FEE_FLAG != 0;
```
saucepoint, line 19: curious on your thoughts around strictly forcing `key.fee = 0x800000` for dynamic-fee pools? would create uniformity for all dynamic fee pools

is there a reason why its `0x800000` and not `0xF00000`?

saucepoint (commented):
had some minor discussion bits -- otherwise looks p good to me

src/PoolManager.sol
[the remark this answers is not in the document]
saucepoint: oh another option is encoding the initial dynamic fee in the bottom 20 bits of `key.fee`

src/PoolManager.sol
```diff
@@ -292,15 +295,11 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
...
-        if (key.fee.isDynamicFee()) {
-            uint24 newDynamicSwapFee = _fetchDynamicSwapFee(key);
-            PoolId id = key.toId();
-            pools[id].setSwapFee(newDynamicSwapFee);
-            emit DynamicSwapFeeUpdated(id, newDynamicSwapFee);
-        } else {
-            revert FeeNotDynamic();
-        }
+    function updateDynamicSwapFee(PoolKey memory key, uint24 newDynamicSwapFee) external {
+        if (!key.fee.isDynamicFee() || msg.sender != address(key.hooks)) revert UnauthorizedDynamicSwapFeeUpdate();
```
hensha256, line 299: this doesnt check that the pool is initialized... i dont think it _needs_ to but that would be weird

src/PoolManager.sol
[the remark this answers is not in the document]
hensha256: >oh another option is encoding the initial dynamic fee in the bottom 20 bits of key.fee

OH YEEEEEEAH thats a great idea! I'll do that!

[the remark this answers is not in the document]
hensha256: i think we can have people pass in whatever fee. If the bit is on, its dynamic fee, if its off, its static fee.
And then we can _store_ `DYNAMIC_FEE_FLAG` in the slot0 to show its a dynamic fee pool?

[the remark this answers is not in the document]
saucepoint: if the top 4 bits 'key.fee' (0x800000) can signal that its a dynamic fee pool, i think its redundant to also store it in slot0

src/PoolManager.sol
```diff
@@ -291,15 +292,11 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
...
+        newDynamicSwapFee.validateSwapFee();
+        PoolId id = key.toId();
+        pools[id].setSwapFee(newDynamicSwapFee);
```
snreynolds, line 299: wait why are we storing it? I think this issue was not about changing whether its stored or not - just about changing the push vs. pull.

snreynolds (changes requested):

src/PoolManager.sol
snreynolds: ignore this^

src/PoolManager.sol
```diff
@@ -104,7 +103,10 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
         override
         returns (int24 tick)
     {
-        if (key.fee.isStaticFeeTooLarge()) revert FeeTooLarge();
+        // For dynamic fee pools, the initial fee is read from the lower bits of the fee parameter
+        uint24 swapFee = key.fee.getSwapFee();
+        swapFee.validateSwapFee();
+        if (key.fee.isDynamicFee()) key.fee = SwapFeeLibrary.DYNAMIC_FEE_FLAG;
```
snreynolds, line 109: This seems odd? The key.fee already has information about whether or not it is a DYNAMIC_FEE.

snreynolds: Ah interesting - This is set because you allow the lower bits of the key to be the initial dynamic fee. I would actually lean against this in my opinion. The `key` of a pool should only contain data that is unchanging because it is used to find unique pools and to submit a key with information that actually will be changed and actually is not part of the key used to initialize a pool is not good practice imo. I understand this may save gas but I think its quite confusing and not how this data should be used. 

It also makes this line of code very odd (hence my confusion in the comment above).

snreynolds (changes requested):
Just don't think we should be using the key to set an initial fee imo.. but happy to discuss more if there are strong opinions

src/PoolManager.sol
snreynolds: I think this is checked in `setSwapFee` ?

src/PoolManager.sol
```diff
@@ -104,19 +103,23 @@ contract PoolManager is IPoolManager, Fees, NoDelegateCall, ERC6909Claims {
...
-
         // see TickBitmap.sol for overflow conditions that can arise from tick spacing being too large
         if (key.tickSpacing > MAX_TICK_SPACING) revert TickSpacingTooLarge();
         if (key.tickSpacing < MIN_TICK_SPACING) revert TickSpacingTooSmall();
         if (key.currency0 >= key.currency1) revert CurrenciesOutOfOrderOrEqual();
         if (!key.hooks.isValidHookAddress(key.fee)) revert Hooks.HookAddressNotValid(address(key.hooks));
 
+        uint24 swapFee = 0;
+        // Initial swap fee on dynamic fee pools is set to 0, unless updateDynamicSwapFee is called in the afterInitialize hook
+        if (!key.fee.isDynamicFee()) {
```
snreynolds, line 114: discussed offline but could put this all in the lib:

```
uint24 swapFee = key.fee.getSwapFee(); // handles both static and dynamic
swapFee.validate();
```

snreynolds (dismissed):
looks good. left one comment up to you if you want to change or not though.

and then lastly wanted to close out [this discussion](https://github.com/Uniswap/v4-core/pull/501/files#r1518391768)

We check if the pool is initialized right now in main in [`setSwapFee`.](https://github.com/Uniswap/v4-core/blob/main/src/libraries/Pool.sol#L124) So I do think this we still have this check and I don't think we should remove it.

snreynolds (approved):

Issue #471, opened 2024-02-01
Skip calls to hooks when msg.sender is the hook contract


danrobinson:
### Component

Gas Optimization

### Describe the suggested feature and problem it solves.

In some cases, DEXes built on v4 may want to have the hook contract make calls into the pool (to swap, to provide liquidity, to donate etc). Examples include:

* Any design where the hook intercepts a user's trade and executes it on their behalf in multiple steps (such as to change fees or add or remove liquidity between ticks)
* Any design where all calls must be gated through a particular contract (which might as well be the hook contract)
* Any design where the hook manages its own liquidity, or manages liquidity for any users
* Any design where the hook uses donate()

In those cases, I would argue that it never makes sense for the pool to do any callbacks to the hook during the operation. If there is any desired logic that should happen before or after the operation, the hook contract can perform that logic before or after calling, rather than during a reentrant call. Having the swap call back into the hook involves a wasteful contract call.

One motivating example is any design where the hook intercepts a user's trade in a beforeSwap hook (such as using the design in #463) and splits the trade into multiple pieces (in order to do something in between each tick, such as donating to the pool, updating some accumulator, or computing the proper swap size). In this case, each tick cross would involve a wasted CALL to the hook.

### Describe the desired implementation.

Change the line [here](https://github.com/Uniswap/v4-core/blob/8109ec3c2f9db321ba48fff44ed429e6c1bd3eb3/src/libraries/Hooks.sol#L208) to:

```solidity
if (key.hooks.hasPermission(BEFORE_SWAP_FLAG) && (msg.sender != address(self)))
```

And make similar changes for all other hook calls.

### Additional context.

The main downside is the special-casing—it's possible it could be confusing for hook developers. On the other hand, I think it actually might be more intuitive in most cases—it automatically prevents a potentially infinite recursion when a `beforeSwap` hook calls `swap`, for example.

The main other cost is the additional gas of the check, but it should be on the order of ~10-20 gas per check.

PR #503, opened 2024-03-11
skip calls to hooks when msg.sender is hook contract + test

dianakocsis:
## Related Issue
Skip calls to hooks when msg.sender is the hook contract #471

## Description of changes
skips calls to hooks when msg.sender is the hook contract + tests


src/libraries/Hooks.sol
```diff
@@ -132,9 +132,13 @@ library Hooks {
...
         bytes calldata hookData
     ) internal {
-        if (params.liquidityDelta > 0 && key.hooks.hasPermission(BEFORE_ADD_LIQUIDITY_FLAG)) {
+        if (params.liquidityDelta > 0 && self.hasPermission(BEFORE_ADD_LIQUIDITY_FLAG) && (msg.sender != address(self)))
+        {
             self.callHook(abi.encodeWithSelector(IHooks.beforeAddLiquidity.selector, msg.sender, key, params, hookData));
-        } else if (params.liquidityDelta <= 0 && key.hooks.hasPermission(BEFORE_REMOVE_LIQUIDITY_FLAG)) {
+        } else if (
+            params.liquidityDelta <= 0 && self.hasPermission(BEFORE_REMOVE_LIQUIDITY_FLAG)
+                && (msg.sender != address(self))
```
hensha256, line 140: given its added for _every_ hook, i wonder if it should go inside the `callHook` function?

src/test/SkipCallsTestHook.sol
```diff
@@ -0,0 +1,62 @@
...
+import {IPoolManager} from "../interfaces/IPoolManager.sol";
+import {PoolKey} from "../types/PoolKey.sol";
+import {BalanceDelta} from "../types/BalanceDelta.sol";
+import {PoolId, PoolIdLibrary} from "../types/PoolId.sol";
+import {FeeLibrary} from "../../src/libraries/FeeLibrary.sol";
+import {IERC20Minimal} from "../interfaces/external/IERC20Minimal.sol";
+import {CurrencyLibrary, Currency} from "../types/Currency.sol";
+import {PoolTestBase} from "./PoolTestBase.sol";
+import {Test} from "forge-std/Test.sol";
+import "forge-std/console.sol";
```
hensha256, line 16: can remove this!

src/test/SkipCallsTestHook.sol
```diff
@@ -0,0 +1,62 @@
...
+contract SkipCallsTestHook is BaseTestHooks, Test {
+    using PoolIdLibrary for PoolKey;
+    using Hooks for IHooks;
+    using FeeLibrary for uint24;
+
+    uint256 public counter;
+    IPoolManager manager;
+    uint24 internal fee;
+
+    constructor() {}
```
hensha256, line 27: you can remove this given its empty 👯

src/libraries/Hooks.sol
dianakocsis: saves some gas too

src/libraries/Hooks.sol
```diff
@@ -96,10 +96,12 @@ library Hooks {
 
     /// @notice performs a hook call using the given calldata on the given hook
     function callHook(IHooks self, bytes memory data) internal {
-        (bytes4 expectedSelector, bytes4 selector) = _callHook(self, data);
+        if (msg.sender != address(self)) {
```
snreynolds, line 99: ooh this is an interesting line of code.. address(self) reads like address(this), but is actually checking address(hook)... I feel like I lean towards a rename of `callHook(IHooks hook, bytes memory data)`

snreynolds: could collapse the if statement by doing 
if (msg.sender == address(hook) return;
_callHook()
if(selector != expectedSelector) revert .. 

which may be clean? but try it out and see if you like it.. not that opinionated

src/test/PoolSwapTest.sol
```diff
@@ -66,9 +66,6 @@ contract PoolSwapTest is Test, PoolTestBase {
         (,, uint256 reserveAfter0, int256 deltaAfter0) = _fetchBalances(data.key.currency0, data.sender, address(this));
         (,, uint256 reserveAfter1, int256 deltaAfter1) = _fetchBalances(data.key.currency1, data.sender, address(this));
 
-        assertEq(reserveBefore0, reserveAfter0);
```
snreynolds, line 69: why do we remove this?

src/test/SkipCallsTestHook.sol
```diff
@@ -0,0 +1,59 @@
...
+    {
+        counter++;
+        callSwap(key, params, hookData);
+        return IHooks.beforeSwap.selector;
+    }
+
+    function callSwap(PoolKey calldata key, IPoolManager.SwapParams calldata params, bytes calldata hookData) public {
+        IPoolManager(msg.sender).swap(key, params, hookData);
+        address payer = abi.decode(hookData, (address));
+        int256 delta0 = IPoolManager(msg.sender).currencyDelta(address(this), key.currency0);
```
snreynolds, line 51: Since you set the pool manager this can just be `IPoolManager(manager)`

src/test/SkipCallsTestHook.sol
```diff
@@ -0,0 +1,59 @@
...
+
+    uint256 public counter;
+    IPoolManager manager;
+    uint24 internal fee;
+
+    function setManager(IPoolManager _manager) external {
+        manager = _manager;
+    }
+
+    function setFee(uint24 _fee) external {
```
snreynolds, line 30: These are unneeded functions i believe for the purpose of this test

test/SkipCallsTestHook.t.sol
```diff
@@ -0,0 +1,70 @@
...
+        vm.etch(address(skipCallsTestHook), address(impl).code);
+        deployFreshManagerAndRouters();
+        skipCallsTestHook.setManager(IPoolManager(manager));
+
+        (currency0, currency1) = deployMintAndApprove2Currencies();
+        (key,) = initPoolAndAddLiquidity(
+            currency0,
+            currency1,
+            IHooks(address(skipCallsTestHook)),
+            FeeLibrary.DYNAMIC_FEE_FLAG,
```
snreynolds, line 51: Why is it also a dynamic fee hook? Seems unnecessary for the purpose of this test?

test/SkipCallsTestHook.t.sol
```diff
@@ -0,0 +1,70 @@
...
+            SQRT_RATIO_1_1,
+            ZERO_BYTES
+        );
+    }
+
+    function test_beforeSwap_invalidSender() public {
```
snreynolds, line 57: what do you mean by `invalid_sender` in the test name?

I think this is just testing that the hook's call to beforeSwap doesnt trigger the hook again yes? thus, the count to beforeSwap is 1

snreynolds (changes requested):

src/libraries/Hooks.sol
dianakocsis: I think I like that! Honestly it took me a while to figure out what was what so I think this will make it more clear

src/test/PoolSwapTest.sol
dianakocsis: There could be a swap within the hook making the reserves not equal to each other so I removed it. Also since reservesOf might be removed entirely anyways

src/test/SkipCallsTestHook.sol
dianakocsis: Oh ya i needed them since I made the hook with a dynamic fee but ill change that so i dont need these

test/SkipCallsTestHook.t.sol
dianakocsis: I was thinking more like invalid caller aka the hook, but ill rename it to `test_beforeSwap_skipIfCalledByHook()` to make it more clear

snreynolds (commented):
changes look good. @hensha256 brought up a good point that changing away from self diverges from how we do this in other libs (Pool.sol, Position.sol) so I'm actually inclined to change it back... and we can discuss more later. Sorry for back and forth there

snreynolds (dismissed):

test/SkipCallsTestHook.t.sol
```diff
@@ -0,0 +1,240 @@
...
+    PoolSwapTest.TestSettings testSettings =
+        PoolSwapTest.TestSettings({withdrawTokens: true, settleUsingTransfer: true, currencyAlreadySent: false});
+
+    function deploy(SkipCallsTestHook skipCallsTestHook) public {
+        SkipCallsTestHook impl = new SkipCallsTestHook();
+        vm.etch(address(skipCallsTestHook), address(impl).code);
+        deployFreshManagerAndRouters();
+        skipCallsTestHook.setManager(IPoolManager(manager));
+
+        (currency0, currency1) = deployMintAndApprove2Currencies();
```
snreynolds, line 37: Why not just put all of this in setUp and then do the init/addLiq as needed in each of the test cases? I think it may be a bit easier to follow what exactly is being called to understand the expected count amount. And I dont think the hook address needs to change each time ie all of this can just happen once? Unless Im missing something

src/test/SkipCallsTestHook.sol
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    function beforeAddLiquidity(
+        address,
+        PoolKey calldata key,
+        IPoolManager.ModifyLiquidityParams calldata,
+        bytes calldata hookData
+    ) external override returns (bytes4) {
+        counter++;
+        _initialize(key, Constants.SQRT_RATIO_1_1, hookData);
```
snreynolds, line 55: Maybe overkill but it seems like we are just calling the same function on the pool again (except for this.. this one does initialize).... Maybe instead we actually fuzz the function thats called inside? That way we are testing that no matter which hook callback we are in, any action on the pool called from the hook skips the call? So it could be any of initialize/swap/modifyPositon/donate

test/SkipCallsTestHook.t.sol
```diff
@@ -0,0 +1,240 @@
...
+                            & ~Hooks.AFTER_ADD_LIQUIDITY_FLAG & ~Hooks.AFTER_REMOVE_LIQUIDITY_FLAG & ~Hooks.BEFORE_SWAP_FLAG
+                            & ~Hooks.AFTER_SWAP_FLAG & ~Hooks.BEFORE_DONATE_FLAG & ~Hooks.AFTER_DONATE_FLAG
+                    )
+            )
+        );
+
+        deploy(skipCallsTestHook);
+
+        MockERC20(Currency.unwrap(key.currency0)).approve(address(skipCallsTestHook), Constants.MAX_UINT256);
+        MockERC20(Currency.unwrap(key.currency1)).approve(address(skipCallsTestHook), Constants.MAX_UINT256);
```
snreynolds, line 126: assert counter is 0 before the call in each of these? that makes it super clear we are expecting a change to counter.. I think you do it below but not here/ i think its missing in some other tests as well

test/SkipCallsTestHook.t.sol
```diff
@@ -0,0 +1,240 @@
...
+
+        deploy(skipCallsTestHook);
+
+        assertEq(skipCallsTestHook.counter(), 0);
+        MockERC20(Currency.unwrap(key.currency0)).approve(address(skipCallsTestHook), Constants.MAX_UINT256);
+        MockERC20(Currency.unwrap(key.currency1)).approve(address(skipCallsTestHook), Constants.MAX_UINT256);
+        donateRouter.donate(key, 100, 200, abi.encode(address(this)));
+        assertEq(skipCallsTestHook.counter(), 1);
+    }
+}
```
snreynolds, line 240: Could we add a test where we call again on the pool and the counter should be 2 but not 4?

snreynolds (changes requested):

test/SkipCallsTestHook.t.sol
dianakocsis: so make one hook and turn all flags on? and check that count is total 10 at the end?

src/test/PoolSwapTest.sol
```diff
@@ -52,23 +52,16 @@ contract PoolSwapTest is PoolTestBase {
...
 
         BalanceDelta delta = manager.swap(data.key, data.params, data.hookData);
 
-        (, uint256 poolBalanceAfter0, int256 deltaAfter0) =
-            _fetchBalances(data.key.currency0, data.sender, address(this));
-        (, uint256 poolBalanceAfter1, int256 deltaAfter1) =
-            _fetchBalances(data.key.currency1, data.sender, address(this));
-
-        require(poolBalanceBefore0 == poolBalanceAfter0, "poolBalanceBefore0 is not equal to poolBalanceAfter0");
-        require(poolBalanceBefore1 == poolBalanceAfter1, "poolBalanceBefore1 is not equal to poolBalanceAfter1");
```
hensha256, line 71: why are these being removed in this PR?

test/SkipCallsTestHook.t.sol
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    function test_beforeInitialize_skipIfCalledByHook() public {
+        SkipCallsTestHook skipCallsTestHook = SkipCallsTestHook(
+            address(
+                uint160(0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF)
+                    & uint160(
+                        ~Hooks.AFTER_INITIALIZE_FLAG & ~Hooks.BEFORE_ADD_LIQUIDITY_FLAG & ~Hooks.AFTER_ADD_LIQUIDITY_FLAG
+                            & ~Hooks.BEFORE_REMOVE_LIQUIDITY_FLAG & ~Hooks.AFTER_REMOVE_LIQUIDITY_FLAG & ~Hooks.BEFORE_SWAP_FLAG
+                            & ~Hooks.AFTER_SWAP_FLAG & ~Hooks.BEFORE_DONATE_FLAG & ~Hooks.AFTER_DONATE_FLAG
```
hensha256, line 57: so these lines can become
```
        address(
            uint160(
                uint256(type(uint160).max) & clearAllHookPermisssionsMask | Hooks.AFTER_INITIALIZE_FLAG
            )
        )
```

hensha256: not sure why it needs to go via uint256... you could try making `clearAllHookPermisssionsMask` a uint160 too

test/SkipCallsTestHook.t.sol
```diff
@@ -0,0 +1,255 @@
...
+                        ~Hooks.BEFORE_INITIALIZE_FLAG & ~Hooks.BEFORE_ADD_LIQUIDITY_FLAG & ~Hooks.AFTER_ADD_LIQUIDITY_FLAG
+                            & ~Hooks.BEFORE_REMOVE_LIQUIDITY_FLAG & ~Hooks.AFTER_REMOVE_LIQUIDITY_FLAG & ~Hooks.BEFORE_SWAP_FLAG
+                            & ~Hooks.AFTER_SWAP_FLAG & ~Hooks.BEFORE_DONATE_FLAG & ~Hooks.AFTER_DONATE_FLAG
+                    )
+            )
+        );
+
+        deploy(skipCallsTestHook);
+
+        assertEq(skipCallsTestHook.counter(), 1);
```
hensha256, line 81: maybe a comment saying why after `deploy` it would be 1 - i had to go and read that function to know why (same in the test above)

test/SkipCallsTestHook.t.sol
```diff
@@ -0,0 +1,255 @@
...
+                            & ~Hooks.BEFORE_REMOVE_LIQUIDITY_FLAG & ~Hooks.AFTER_REMOVE_LIQUIDITY_FLAG & ~Hooks.BEFORE_SWAP_FLAG
+                            & ~Hooks.AFTER_SWAP_FLAG & ~Hooks.BEFORE_DONATE_FLAG & ~Hooks.AFTER_DONATE_FLAG
+                    )
+            )
+        );
+
+        deploy(skipCallsTestHook);
+
+        approveAndAddLiquidity(skipCallsTestHook);
+        assertEq(skipCallsTestHook.counter(), 1);
```
hensha256, line 119: again comments - this surprised me because i thought it would be 1 after `deploy` so 2 after adding liquidity - but its because theres no initialize hook at all.
Might be nice to add
```
        assertEq(skipCallsTestHook.counter(), 0);
```
before adding liq too

test/SkipCallsTestHook.t.sol
```diff
@@ -0,0 +1,255 @@
...
+                            & ~Hooks.AFTER_ADD_LIQUIDITY_FLAG & ~Hooks.AFTER_REMOVE_LIQUIDITY_FLAG & ~Hooks.BEFORE_SWAP_FLAG
+                            & ~Hooks.AFTER_SWAP_FLAG & ~Hooks.BEFORE_DONATE_FLAG & ~Hooks.AFTER_DONATE_FLAG
+                    )
+            )
+        );
+
+        deploy(skipCallsTestHook);
+        approveAndAddLiquidity(skipCallsTestHook);
+
+        modifyLiquidityRouter.modifyLiquidity(key, REMOVE_LIQ_PARAMS, abi.encode(address(this)));
```
hensha256, line 139:
```
        assertEq(skipCallsTestHook.counter(), 0);
```

above this line too

hensha256 (changes requested):
generally looking great!! couple of small things but we're nearly there

src/test/PoolSwapTest.sol
snreynolds: I think its because the hook might do a swap/add liq and settle that is nested. meaning the balance would have changed before the outermost swap settles

test/SkipCallsTestHook.t.sol
snreynolds: i think you commented something similar in #548 which I can do over there

test/SkipCallsTestHook.t.sol
snreynolds: yeah i think theres still a couple more spots the assert counter is 0 would be nice to have

snreynolds (changes requested):

test/SkipCallsTestHook.t.sol
dianakocsis: ah so much nicer! thank you!!

src/test/PoolSwapTest.sol
dianakocsis: yep exactly, the hook does its own swap which changes the balance

test/ERC6909Claims.t.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity ^0.8.15;
 import {Test} from "forge-std/Test.sol";
 import {CurrencyLibrary, Currency} from "../src/types/Currency.sol";
 import {MockERC6909Claims} from "../src/test/MockERC6909Claims.sol";
+import "forge-std/console.sol";
```
hensha256, line 7: remove!

hensha256 (approved):

snreynolds (approved):

Issue #490, opened 2024-02-29
Apply fee on top of rather than subtracted from swap fee


danrobinson:
### Component

General design optimization (improving efficiency, cleanliness, or developer experience)

### Describe the suggested feature and problem it solves.

Right now, the protocol fee is specified as a denominator of a fraction that is multiplied by the swap fee to determine the protocol fee, with the resulting protocol fee subtracted from the swap fee before the latter is applied. This seems weird for a few reasons.

* Specifying a denominator is just weirdly imprecise and inconsistent with how fees are measured everywhere else.
* Charging as a fraction of swap fees means that the imposed protocol fee would be higher on pools that charge higher fees even for the same asset pair, which doesn’t feel right when you think about it.
* It could allow some hooks to evade protocol fees by charging swap fee in a different way (such as using the deltas in #482). 
* It seems less natural for a change to result in liquidity providers receiving a lower fee on a particular swap than they would expect; rather than resulting in swappers paying a higher fee than they expect, given that liquidity providers and hooks would usually be considered “slower” than swappers. (This argument arguably also applies to how protocol fee worked in v2 and v3, but seems stronger in v4 due to hooks.)

### Describe the desired implementation.

Have `protocolFee` specified in pips (hundredth of a basis point), like swap fee, with some cap. During a swap, have it applied on top of the existing swap fee (so it is effectively paid by swappers), rather than taking away from the swap fee earned by LPs.

Since 2^16 pips would be 6.5%, which is much higher than where you’d want to set the cap, you shouldn’t have to change the size or type of the variable.

### Describe alternatives.

We could have the protocol fee be subtracted from the fees earned by LPs, rather than added on top and paid by swappers. This raises the question of what would happen if protocol fee is higher than swap fee.

PR #504, opened 2024-03-12
Change protocol fee from denominator to bips

dianakocsis:
## Related Issue
Change protocol fee from denominator to pips, and apply on top of rather than subtracted from swap fee #490

## Description of changes
Half of the related issue: changes protocol fee from denominator to bips, protocol fee is still subtracted from swap fee


src/Fees.sol
```diff
@@ -14,7 +14,7 @@ abstract contract Fees is IFees, Owned {
     using FeeLibrary for uint24;
     using CurrencyLibrary for Currency;
 
-    uint8 public constant MIN_PROTOCOL_FEE_DENOMINATOR = 4;
+    uint16 public constant MAX_PROTOCOL_FEE = 2500;
```
snreynolds, line 17: Could we add a comment this is denominated in bps?

src/libraries/Pool.sol
```diff
@@ -326,7 +325,7 @@ library Pool {
 
         SwapCache memory cache = SwapCache({
             liquidityStart: self.liquidity,
-            protocolFee: params.zeroForOne ? uint8(slot0Start.protocolFee % 256) : uint8(slot0Start.protocolFee >> 8)
+            protocolFee: params.zeroForOne ? uint16(slot0Start.protocolFee % 4096) : uint16(slot0Start.protocolFee >> 12)
```
snreynolds, line 328: Maybe add helpers in the fee library to extract the 1->0 and 0->1 fees?

```
slot0Start.protocolFee.toZeroForOne();
slot0Start.protocolFee.toOneForZero();
```

snreynolds: cause you could use them also in the `_isFeeWithinBounds` function

snreynolds: & then also would be nice/easier to write unit tests making sure that the 0 to 1 and 1 to 0 extraction are as expected

test/PoolManager.t.sol
```diff
@@ -1000,12 +999,12 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
     }
 
     function test_collectProtocolFees_nativeToken_accumulateFees_gas() public {
-        uint16 protocolFee = 1028; // 00000100 00000100
+        uint24 protocolFee = 10242500; // 100111000100 100111000100
```
snreynolds, line 1002: Could you make this number a constant in the test file? `MAX_FEE_BOTH_TOKENS`? // 2500 2500

snreynolds (changes requested):

src/libraries/Pool.sol
dianakocsis: we dont have a Fees library anymore, only SwapFeeLibrary...I could make a ProtocolFeeLibrary?

snreynolds: sure that works, just more so meant to put it in a helper function/lib so you can reuse it since this math is done in a few places

snreynolds (commented):
looks great, just one comment about moving the fee extraction for 0 to 1 and vice versa  into a lib/helper func

snreynolds (approved):

src/libraries/ProtocolFeeLibrary.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.20;
+
+import {PoolKey} from "../types/PoolKey.sol";
+
+library ProtocolFeeLibrary {
```
hensha256, line 6: great library addition 🙌

snreynolds (dismissed):

test/PoolManager.t.sol
```diff
@@ -52,13 +54,15 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
     );
-    event ProtocolFeeUpdated(PoolId indexed id, uint16 protocolFee);
+    event ProtocolFeeUpdated(PoolId indexed id, uint24 protocolFee);
     event Transfer(
         address caller, address indexed sender, address indexed receiver, uint256 indexed id, uint256 amount
     );
 
     PoolEmptyUnlockTest emptyUnlockRouter;
 
+    uint24 constant MAX_FEE_BOTH_TOKENS = 10242500; // 2500 2500
```
hensha256, line 64: nit - i think this might be easier to read in hex. 0x9c49c4 - just because its clearly the same thing repeated twice
OR you could write it as like `(2500 << 12) & 2500`

test/PoolManager.t.sol
```diff
@@ -741,13 +745,13 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
 
-    function test_swap_accruesProtocolFees(uint8 protocolFee1, uint8 protocolFee0) public {
-        protocolFee0 = uint8(bound(protocolFee0, 4, type(uint8).max));
-        protocolFee1 = uint8(bound(protocolFee1, 4, type(uint8).max));
+    function test_swap_accruesProtocolFees(uint16 protocolFee0, uint16 protocolFee1) public {
+        protocolFee0 = uint16(bound(protocolFee0, 1, 2500));
+        protocolFee1 = uint16(bound(protocolFee1, 1, 2500));
 
-        uint16 protocolFee = (uint16(protocolFee1) << 8) | (uint16(protocolFee0) & uint16(0xFF));
+        uint24 protocolFee = (uint24(protocolFee1) << 12) | (uint24(protocolFee0) & uint24(0x1000));
```
hensha256, line 752: `uint24(protocolFee0) & uint24(0x1000)`

that feels wrong to me - isnt that canceling out the protocol fee? like if protocolFee0 is `1` then `uint24(protocolFee0) & uint24(0x1000)` is going to be 0?

hensha256: i think you can just remove the &

test/PoolManager.t.sol
```diff
@@ -989,18 +990,17 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
         manager.setProtocolFee(key);
 
         (Pool.Slot0 memory slot0,,,) = manager.pools(key.toId());
-        assertEq(slot0.protocolFee, protocolFee);
+        assertEq(slot0.protocolFee, MAX_FEE_BOTH_TOKENS);
 
         swapRouter.swap(
             key,
-            IPoolManager.SwapParams(true, 10000, SQRT_RATIO_1_2),
+            IPoolManager.SwapParams(true, -10000, SQRT_RATIO_1_2),
```
hensha256, line 1003: why is this sign flipped?

hensha256 (commented):
few thoughts

test/PoolManager.t.sol
snreynolds: hex would be cheaper potentially? unless the optimizer does this for us

test/PoolManager.t.sol
snreynolds: Isn't this just clearing the extra bits? we're fuzzing with uint16s but they each should be 12 bits so they pack to a uint24

snreynolds: oh hm so it should be `& 0xFFF` no?

hensha256: but it already does a `bound` above so i dont think any masking is needed

dianakocsis: oh yeah i think i just need to remove the &

test/PoolManager.t.sol
dianakocsis: since we're sending in exact input 1000 of token0 for token1

src/ProtocolFees.sol
```diff
@@ -44,21 +47,19 @@ abstract contract ProtocolFees is IProtocolFees, Owned {
...
-            uint16 fee0 = fee % 256;
-            uint16 fee1 = fee >> 8;
-            // The fee is specified as a denominator so it cannot be LESS than the MIN_PROTOCOL_FEE_DENOMINATOR (unless it is 0).
-            if (
-                (fee0 != 0 && fee0 < MIN_PROTOCOL_FEE_DENOMINATOR) || (fee1 != 0 && fee1 < MIN_PROTOCOL_FEE_DENOMINATOR)
-            ) {
+            uint16 fee0 = fee.getZeroForOneFee();
+            uint16 fee1 = fee.getOneForZeroFee();
+            // The fee is represented in bips so it cannot be GREATER than the MAX_PROTOCOL_FEE.
+            if ((fee0 > MAX_PROTOCOL_FEE) || (fee1 > MAX_PROTOCOL_FEE)) {
```
hensha256, line 62: i'm wondering if we should put this into the protocolFeeLibrary like 
```
return fee0.isValid() && fee1.isValid()
```

and then 

```
function isValid(uint24 self) ... {
   return self > MAX;
}
```

and then we can put MAX_PROTOCOL_FEE into that library instead

hensha256: and then i feel like `BIPS_DENOMINATOR` would ideally be in there too? if theres a neat way to put it in there and not in the pool contract? even if its just then that the pool contract does `ProtocolFeeLibrary.BIPS_DENOMINATOR`

snreynolds: I generally feel like what we have now is a pretty good structure to be honest? This function is already supposed to check for validity of the fee so it seems almost overkill to have another library that also checks for validity?

Also in general I think moving things to libraries is good when its reused and we dont ever check validity anywhere else so not really sure its that helpful.

snreynolds: I do like the idea though of maybe moving` BPS_DENOMINATOR`. What about moving that to ProtocolFeeLibrary and then in Pool.sol we could have a call thats like `cache.protocolFee.convertToBPS(step.feeAmount`);

snreynolds: Or I'd also just be fine leaving as is tbh.. maybe that last line is harder to read than seeing the calculation in Pool.sol

src/test/ProtocolFeeControllerTest.sol
```diff
@@ -9,36 +9,36 @@ import {PoolKey} from "../types/PoolKey.sol";
...
     }
 }
 
 /// @notice Returns an out of bounds protocol fee
 contract OutOfBoundsProtocolFeeControllerTest is IProtocolFeeController {
-    function protocolFeeForPool(PoolKey memory /* key */ ) external pure returns (uint16) {
-        // set both swap and withdraw fees to 1, which is less than MIN_PROTOCOL_FEE_DENOMINATOR
-        return 0x001001;
+    function protocolFeeForPool(PoolKey memory /* key */ ) external pure returns (uint24) {
+        // set both swap and withdraw fees to 2501, which is greater than MAX_PROTOCOL_FEE
```
snreynolds, line 34: Can remove the withdraw comment, I think that is leftover from an old old protocol fee implementation

snreynolds (dismissed):
looks good! one tiny comment then we're good to merge

snreynolds (approved):

PR #511, opened 2024-03-14
make reservesOf transient

snreynolds:
## Related Issue

## Description of changes


src/test/PoolDonateTest.sol
```diff
@@ -52,14 +57,22 @@ contract PoolDonateTest is PoolTestBase, Test {
...
+        // and check that there should be no change in balances from the donate call.
+        manager.sync(data.key.currency0);
+        manager.sync(data.key.currency1);
         (,, uint256 reserveAfter0, int256 deltaAfter0) = _fetchBalances(data.key.currency0, data.sender, address(this));
         (,, uint256 reserveAfter1, int256 deltaAfter1) = _fetchBalances(data.key.currency1, data.sender, address(this));
 
         assertEq(reserveBefore0, reserveAfter0);
         assertEq(reserveBefore1, reserveAfter1);
 
+        console2.log(reserveAfter0);
```
hensha256, line 74: can remove console logs on this file

src/PoolManager.sol
```diff
@@ -257,8 +265,8 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     function take(Currency currency, address to, uint256 amount) external override noDelegateCall onlyWhenUnlocked {
         // subtraction must be safe
         _accountDelta(currency, -(amount.toInt128()));
-        reservesOf[currency] -= amount;
         currency.transfer(to, amount);
+        sync(currency);
```
danrobinson, line 269: Why sync here?

src/PoolManager.sol
```diff
@@ -137,6 +138,13 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
         Lock.lock();
     }
 
+    /// @inheritdoc IPoolManager
+    function sync(Currency currency) public returns (uint256 balance) {
+        balance = currency.balanceOfSelf();
+        if (balance == 0) balance = ZERO_BALANCE;
```
ewilz, line 144: haha too bad tstore couldn't have a default value it's always initialized with.

.forge-snapshots/removeLiquidity with empty hook.snap
```diff
@@ -1 +1 @@
-96164
\ No newline at end of file
+98094
```
snreynolds, line 1: this is odd

.forge-snapshots/swap against liquidity.snap
```diff
@@ -1 +1 @@
-100359
\ No newline at end of file
+103858
```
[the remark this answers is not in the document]
snreynolds: & gas increases

.forge-snapshots/swap with dynamic fee.snap
```diff
@@ -1 +1 @@
-133664
\ No newline at end of file
+135727
```
snreynolds, line 1: odd

src/PoolManager.sol
```diff
@@ -275,11 +282,11 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
         if (currency.isNative()) {
             paid = msg.value;
         } else {
-            uint256 reservesBefore = reservesOf[currency];
-            reservesOf[currency] = currency.balanceOfSelf();
-            paid = reservesOf[currency] - reservesBefore;
+            uint256 reservesBefore = Reserves.get(currency);
+            if (reservesBefore == 0) revert ReservesMustBeSynced();
+            uint256 reservesNow = sync(currency);
```
hensha256, line 287: imo this doesnt need to be a call to `sync` we just need to read the current levels and not store them?

src/libraries/Reserves.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.20;
+
+import {Currency, CurrencyLibrary} from "../types/Currency.sol";
+
+library Reserves {
+    using CurrencyLibrary for Currency;
+
+    uint256 constant RESERVES_OF_SLOT = uint256(keccak256("ReservesOf")) - 1;
```
hensha256, line 9: can hardcode! thatll help with gas

src/libraries/Reserves.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.20;
+
+import {Currency, CurrencyLibrary} from "../types/Currency.sol";
+
+library Reserves {
+    using CurrencyLibrary for Currency;
```
hensha256, line 7: whats this for?

src/PoolManager.sol
```diff
@@ -139,6 +140,13 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
         Lock.lock();
     }
 
+    /// @inheritdoc IPoolManager
+    function sync(Currency currency) public returns (uint256 balance) {
```
hensha256, line 144: feel like we should be adding the same modifiers as the other functions? like onlyWhenUnlocked

src/PoolManager.sol
```diff
@@ -275,11 +282,11 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
         if (currency.isNative()) {
             paid = msg.value;
         } else {
-            uint256 reservesBefore = reservesOf[currency];
-            reservesOf[currency] = currency.balanceOfSelf();
-            paid = reservesOf[currency] - reservesBefore;
+            uint256 reservesBefore = Reserves.get(currency);
```
hensha256, line 285: might be nicer as `currency.getReserves()` like our other libraries

src/interfaces/IPoolManager.sol
```diff
@@ -32,6 +32,8 @@ interface IPoolManager is IProtocolFees, IERC6909Claims {
     /// @notice PoolKey must have currencies where address(currency0) < address(currency1)
     error CurrenciesOutOfOrderOrEqual();
 
+    /// @notice Thrown when the transient reserves are not in sync with the current balance
+    error ReservesMustBeSynced();
     /// @notice Thrown when a call to updateDynamicSwapFee is made by an address that is not the hook,
```
hensha256, line 37: nit: empty line between errors like the other errors

src/test/PoolSwapTest.sol
```diff
@@ -106,13 +106,17 @@ contract PoolSwapTest is PoolTestBase {
 
         if (deltaAfter0 < 0) {
             if (data.testSettings.currencyAlreadySent) {
+                // Todo: This is a downside of sync!!! We must call sync before transferring in assets.
```
hensha256, line 109: maybe we should just get rid of these test cases? or add sync before them like we would do if we were swap V2->V4

src/PoolManager.sol
```diff
@@ -275,11 +282,11 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
...
+            if (reservesBefore == 0) revert ReservesMustBeSynced();
```
hensha256, line 286: would be great to see a test for calling settle without calling sync first

hensha256 (changes requested):
left some comments! mostly looking great would just love more tests

src/PoolManager.sol
snreynolds: I don't think its necessary to call it inside the lock, like don't see any security vulns with it. But do you prefer it for consistency?

hensha256: mm no thats ok im happy with it outside

src/PoolManager.sol
snreynolds: I added these tests. I think there are likely more invariant/integ tests but either can followup after this (add to testing doc) or include here. up to you. would love another look though!

.forge-snapshots/swap skips hook call if hook is caller.snap
```diff
@@ -1 +1 @@
-156254
\ No newline at end of file
+159491
```
snreynolds, line 1: ope... why is this more expensive?

src/PoolManager.sol
```diff
@@ -144,6 +146,13 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
         Lock.lock();
     }
 
+    /// @inheritdoc IPoolManager
+    function sync(Currency currency) public returns (uint256 balance) {
+        balance = currency.balanceOfSelf();
+        if (balance == 0) balance = ZERO_BALANCE;
```
hensha256, line 152: i'm wondering if this should go inside the Reserves file?
Then in PoolManager we just have like `setReserves` and `getReserves` and `Reserves.sol` defines `ZERO_BALANCE` and `ReservesMustBeSynced ` and handles the `if` statements internally? what do you think?

hensha256: hm i do then see than `getReserves` wouldnt be a good view function if it sometimes reverts...
just feels like `ZERO_BALANCE` should live in there not here

src/interfaces/IPoolManager.sol
```diff
@@ -204,4 +211,6 @@ interface IPoolManager is IProtocolFees, IERC6909Claims {
     /// @param nSlots Number of slots to load into return value
     /// @return value The value of the sload-ed slots concatenated as dynamic bytes
     function extsload(bytes32 slot, uint256 nSlots) external view returns (bytes memory value);
+
+    function getReserves(Currency currency) external view returns (uint256 balance);
```
hensha256, line 215: natspec! good to say in it that these are just temporary values not the true reserves at all times

src/PoolManager.sol
hensha256: mm the more i think about it the more im not sure 😆

src/test/BadRouter.sol
```diff
@@ -0,0 +1,138 @@
...
+import {BalanceDelta, BalanceDeltaLibrary} from "../types/BalanceDelta.sol";
+import {PoolKey} from "../types/PoolKey.sol";
+import {IHooks} from "../interfaces/IHooks.sol";
+import {Hooks} from "../libraries/Hooks.sol";
+import {PoolTestBase} from "./PoolTestBase.sol";
+import {Hooks} from "../libraries/Hooks.sol";
+import {IHooks} from "../interfaces/IHooks.sol";
+import {IERC20Minimal} from "../interfaces/external/IERC20Minimal.sol";
+
+contract BadRouter is PoolTestBase {
```
hensha256, line 15: LOL it took me a second to figure out why the router is bad. Could you add a comment above just saying that its bc it doesnt call sync?

src/test/SkipCallsTestHook.sol
```diff
@@ -163,9 +164,10 @@ contract SkipCallsTestHook is BaseTestHooks, Test {
 
         assert(delta0 < 0 || delta1 < 0);
         assert(!(delta0 > 0 || delta1 > 0));
-
+        manager.sync(key.currency0);
```
hensha256, line 167: mm this is ok but do you think that this contract should just inherit from PoolTestBase so that it can access `_settle` and `_take` instead of custom impl?
Can fix that in a different PR if you think thats better

test/utils/Deployers.sol
```diff
@@ -114,6 +114,25 @@ contract Deployers {
         return (currency0, currency1);
     }
 
+    function deployMintAndApproveCurrency() internal returns (Currency currency) {
```
hensha256, line 117: i feel like `deployMintAndApprove2Currencies` should just call this twice? to save duplication of the list of approvals needed. Then it can just set the 2 addresses returned to 0 and 1 and return them

src/PoolManager.sol
saucepoint: i like the idea of consolidating some logic to the Reserves library, especially for storing the ZERO_BALANCE constant

---

as for making `getReserves` revert or not, it might sense to do so. I'm more concerned of devs/builders/integrators assuming that `0` is a valid return

src/test/SkipCallsTestHook.sol
saucepoint: i think sara was open to adding this library to core:

https://github.com/saucepoint/v4-periphery/blob/position-manager/contracts/libraries/CurrencySettleTake.sol (needs to be updated to call .sync() before calling .transferFrom())

the library should replace these calls + PoolTestBase's _settle and _take with something thats a bit more syntactic sugary 😸 

---

In any case, I'd like to see a unified way of settling/taking across core + periphery. either with a library or a contract

saucepoint: tangentially, the library works better in periphery after PoolManager has been deployed 😭 

you can save the PoolManager address as a constant and go from:

`key.currency0.settle(IPoolManager(manager), address(payer), amount, useClaims)`
to
`key.currency.settle(address(payer), amount, useClaims)`

src/PoolManager.sol
snreynolds: hm I slightly lean that it should live in sync only because get/set I see are just the interfaces for writing to transient storage and don't need to/(probably shouldn't) handle logic that is specific to implementation details of the sync function.

src/test/SkipCallsTestHook.sol
snreynolds: Yeah +1 to @saucepoint's lib. Definitely in favor of adding it. Let's push this to a followup PR. Outlined issue here. @saucepoint can you take this?

Re your second point... could we just have one in core for testing purposes? And then expose that one in periphery for prod use case?

src/libraries/Reserves.sol
```diff
@@ -0,0 +1,40 @@
...
+
+import {Currency} from "../types/Currency.sol";
+
+library Reserves {
+    /// uint256(keccak256("ReservesOf")) - 1
+    uint256 constant RESERVES_OF_SLOT = uint256(0x1e0745a7db1623981f0b2a5d4232364c00787266eb75ad546f190e6cebe9bd95);
+    /// @notice The transient reserves for pools with no balance is set to the max as a sentinel to track that it has been synced.
+    uint256 public constant ZERO_BALANCE = type(uint256).max;
+
+    /// @notice Thrown when the transient reserves are not in sync with the current balance
```
hensha256, line 12: nit: this comment isnt quite right. When `ReservesMustBeSynced` is emitted, someone is calling `settle` which means that the current balance != transient reserves because they want to pay the difference.
```
/// @notice Thrown when someone has not called sync before calling settle
```
or something like that

src/test/BadRouter.sol
```diff
@@ -0,0 +1,139 @@
...
+    ) external payable returns (BalanceDelta delta) {
+        delta = abi.decode(
+            manager.unlock(abi.encode(CallbackData(msg.sender, testSettings, key, params, hookData))), (BalanceDelta)
+        );
+
+        uint256 ethBalance = address(this).balance;
+        if (ethBalance > 0) CurrencyLibrary.NATIVE.transfer(msg.sender, ethBalance);
+    }
+
+    function unlockCallback(bytes calldata rawData) external returns (bytes memory) {
```
hensha256, line 51: is this not all stuff the ActionsRouter can do too?

src/test/SkipCallsTestHook.sol
```diff
@@ -199,6 +202,8 @@ contract SkipCallsTestHook is BaseTestHooks, Test {
         address payer = abi.decode(hookData, (address));
         int256 delta0 = IPoolManager(manager).currencyDelta(address(this), key.currency0);
         int256 delta1 = IPoolManager(manager).currencyDelta(address(this), key.currency1);
+        manager.sync(key.currency0);
+        manager.sync(key.currency1);
         IERC20Minimal(Currency.unwrap(key.currency0)).transferFrom(payer, address(manager), uint256(-delta0));
         IERC20Minimal(Currency.unwrap(key.currency1)).transferFrom(payer, address(manager), uint256(-delta1));
```
hensha256, line 208: technically this test is a bad practice bd if the `currency1` calls `settle(currency0)` in `transferFrom`, it would be able to take those tokens that have been sent in but not settled.
So maybe good practice just to change the order here to sync-send-settle-sync-send-settle

test/Reserves.t.sol
```diff
@@ -0,0 +1,50 @@
...
+contract ReservesTest is Test {
+    using Reserves for Currency;
+
+    Currency currency0;
+
+    function setUp() public {
+        currency0 = Currency.wrap(address(0xbeef));
+    }
+
+    function test_get_reverts_withoutSet() public {
```
hensha256, line 17: nit: `_getReserves_`

test/Reserves.t.sol
```diff
@@ -0,0 +1,50 @@
...
+        vm.expectRevert(Reserves.ReservesMustBeSynced.selector);
+        uint256 value = currency0.getReserves();
```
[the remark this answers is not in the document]
hensha256: i dont understand what youre asking if we should do - what would the wrapper do?

test/Sync.t.sol
```diff
@@ -0,0 +1,248 @@
...
+    function test_sync_balanceIsNonZero() public {
+        uint256 currency0Balance = currency0.balanceOf(address(manager));
+        assertGt(currency0Balance, uint256(0));
+
+        // Without calling sync, getReserves should revert.
+        vm.expectRevert(Reserves.ReservesMustBeSynced.selector);
+        manager.getReserves(currency0);
+
+        uint256 balance = manager.sync(currency0);
+        assertEq(balance, currency0Balance, "balance not equal");
```
hensha256, line 52: nice to add another 
``` 
       assertEq(manager.getReserves(currency2), balance);
```

to show now that sync has been called it will not revert

test/Sync.t.sol
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    function test_settle_withBalance() public {
```
hensha256, line 55: nit: could you change `withBalance` and `withNoBalance` to `withStartingBalance` and `withNoStartingBalance` or something like that? wasnt sure what you meant at first

src/test/BadRouter.sol
```diff
@@ -0,0 +1,139 @@
...
+import {PoolKey} from "../types/PoolKey.sol";
+import {IHooks} from "../interfaces/IHooks.sol";
+import {Hooks} from "../libraries/Hooks.sol";
+import {PoolTestBase} from "./PoolTestBase.sol";
+import {Hooks} from "../libraries/Hooks.sol";
+import {IHooks} from "../interfaces/IHooks.sol";
+import {IERC20Minimal} from "../interfaces/external/IERC20Minimal.sol";
+
+/// @notice A Swap router that doesn't call sync.
+contract BadRouter is PoolTestBase {
```
hensha256, line 16: `BadRouter` -> `BadSwapRouter` ? throughout

test/Sync.t.sol
```diff
@@ -0,0 +1,248 @@
...
+        assertEq(manager.getReserves(currency2), 0);
+
+        Actions[] memory actions = new Actions[](2);
+        actions[0] = Actions.SETTLE;
+
+        bytes[] memory params = new bytes[](2);
+        params[0] = abi.encode(currency2);
+
+        actions[1] = Actions.ASSERT_DELTA_EQUALS;
+        params[1] = abi.encode(currency2, address(router), 0);
```
hensha256, line 121: niceee

src/test/SkipCallsTestHook.sol
snreynolds: good point - sauces lib will fix this and make sure people dont make the same mistake. Will change here though.

src/libraries/Reserves.sol
```diff
@@ -0,0 +1,40 @@
...
+    /// uint256(keccak256("ReservesOf")) - 1
+    uint256 constant RESERVES_OF_SLOT = uint256(0x1e0745a7db1623981f0b2a5d4232364c00787266eb75ad546f190e6cebe9bd95);
+    /// @notice The transient reserves for pools with no balance is set to the max as a sentinel to track that it has been synced.
+    uint256 public constant ZERO_BALANCE = type(uint256).max;
+
+    /// @notice Thrown when someone has not called sync before calling settle for the first time.
+    error ReservesMustBeSynced();
+
+    function setReserves(Currency currency, uint256 value) internal {
+        if (value == 0) value = ZERO_BALANCE;
```
hensha256, line 16: i think we need a  `if (value == ZERO_BALANCE) revert()`
because...
- contract has `balance==type(uint256).max`
- call `sync` which passes in `type(uint256).max` which gets stored
- call `settle` immediately which performs `reservesNow - reservesBefore` which is `type(uint256).max - 0` and then settles `type(uint256).max` to the caller

i dont really see anything you can do with that attack because you'd basically have to mint `max` to the contract and then youd be able to claim them... feels pointless. But I think we should add the check.

snreynolds: Yeah interesting - this makes me want to just use a different implementation of storing a flag if somethings been synced or not

snreynolds: Ok I believe we are deeming this a non issue because we assume that token balances are less than `type(uint128).max `

src/interfaces/IPoolManager.sol
dianakocsis: +1

test/Sync.t.sol
```diff
@@ -0,0 +1,244 @@
...
+        manager.getReserves(currency2);
+
+        manager.sync(currency2);
+        assertEq(manager.getReserves(currency2), 0);
+
+        Actions[] memory actions = new Actions[](2);
+        actions[0] = Actions.SETTLE;
+
+        bytes[] memory params = new bytes[](2);
+        params[0] = abi.encode(currency2);
```
hensha256, line 114: super super nit but i would rather this test was like the others and did
```
Actions[] memory actions = new Actions[](2);
bytes[] memory params = new bytes[](2);
```

and then

```
        actions[0] = Actions.SETTLE;
        params[0] = abi.encode(currency2);
```

much easier to read what the first action is that way

test/Sync.t.sol
```diff
@@ -0,0 +1,244 @@
...
+
+    /// @notice When there is no actual balance in the pool, but reservesBefore is outdated (as sync has not been called),
+    /// a delta of type(uint256).max - reservesBefore should not be applied, and no amount should be takeable from the pool.
+    function test_settle_afterTake_doesNotApplyDelta() public {
+        Currency currency3 = deployMintAndApproveCurrency();
+
+        // Approve the router for a transfer.
+        MockERC20(Currency.unwrap(currency3)).approve(address(router), type(uint256).max);
+
+        // Sync has not been called on currency0.
```
hensha256, line 130:
```suggestion
        // Sync has not been called on currency3.
```

test/Sync.t.sol
```diff
@@ -0,0 +1,244 @@
...
+        params[0] = abi.encode(currency2);
+
+        actions[1] = Actions.ASSERT_DELTA_EQUALS;
+        params[1] = abi.encode(currency2, address(router), 0);
+
+        router.executeActions(actions, params);
+    }
+
+    /// @notice When there is no actual balance in the pool, but reservesBefore is outdated (as sync has not been called),
+    /// a delta of type(uint256).max - reservesBefore should not be applied, and no amount should be takeable from the pool.
```
hensha256, line 123:
```suggestion
    /// a delta of ZERO_BALANCE - reservesBefore should not be applied when calling settle, and no amount should be takeable from the pool.
```

hensha256: or something like this... took me a long time to understand what this test is actually testing

snreynolds: ok I updated extensively haha

hensha256 (dismissed):

hensha256 (approved):

PR #513, opened 2024-03-18
Hardcode constants

hensha256:
## Related Issue
Constants are not evaluated at compile time, but at runtime, making these `keccak256` constants expensive.


.forge-snapshots/addLiquidity with empty hook.snap
```diff
@@ -1 +1 @@
-312159
\ No newline at end of file
+311322
```
codyborn, line 1: are these gas estimations?

codyborn (approved):

.forge-snapshots/addLiquidity with empty hook.snap
snreynolds: yes these are gas snapshots. you can see that we set them with `snapStart("<>")`

snreynolds (dismissed):

snreynolds (approved):

Issue #512, opened 2024-03-18
Support testing/experimentation with other environment than foundry


Amxx:
### Describe the suggested feature and problem it solves.

[PoolTestBase.sol](https://github.com/Uniswap/v4-core/tree/main/src/test/PoolTestBase.sol) is used for testing both here and in v4-periphery, where [HookEnabledSwapRouter.sol](https://github.com/Uniswap/v4-periphery/blob/main/test/utils/HookEnabledSwapRouter.sol) depends on it. `PoolTestBase.sol` depends on `forge-std/Test.sol`, which is foundry specific. (for `assertGt` and `assertLt`).

These contract are usefull to testing potential integration with UniswapV4, both using foundry and other environment. In particular, my testing involves nes EIP candidate that I have implemented in hardhat-core and in ethers.js. 

I lost the hability to test since I pulled a new version of v4-core and v4-periphery, because this dependency on the Foundry Vm.

### Describe the desired implementation.

Would it be possible to avoid depending on the Foundry Vm for that (and use a normal `require` or a custom error) ? You're making it unecessarily difficult to test using other environments.

### Describe alternatives.

Provide a set of "preset" contracts that implement a minimum viable example of v4 (manager + provide liquidity + do a swap) that have no foundry dependency

### Additional context.

[This is the tests I'm working on](https://github.com/Amxx/EIP-5806/tree/main/packages/test-uniswap) (impact of EIP-5806 tx on usage and gas cost).

saucepoint:
thanks for flagging this. im in favor of the change! we should be able to replace all the asserts with require statements

especially because the contracts are in 'src/' and not 'test/'

hensha256:
Fixed in #516

PR #516, opened 2024-03-18
make PoolTestBase.sol foundry independent

nishim3:
## Related Issue
Support testing/experimentation with other environment than foundry #512

## Description of changes
Made the ```PoolTestBase.sol``` and some other tests in the ```src``` folder foundry independent.


src/test/PoolDonateTest.sol
```diff
@@ -48,18 +48,18 @@ contract PoolDonateTest is PoolTestBase {
         (,, uint256 reserveBefore1, int256 deltaBefore1) =
             _fetchBalances(data.key.currency1, data.sender, address(this));
 
-        assertEq(deltaBefore0, 0);
-        assertEq(deltaBefore1, 0);
+        require(deltaBefore0 == 0);
```
hensha256, line 51: Now that we cannot use asserts, it would be great to have revert reasons in these errors in my opinion. Nothing fancy just something like
`require(deltaBefore0 == 0, 'deltaBefore0 not 0');

or something like that?

hensha256: or `deltaBefore0 not equal to 0`

hensha256: sorry i know that will be a bit boring to do on all these lines 😆 itll just be very helpful for debugging tests

nishim3: I will do it. While I am at it, would it be a good idea to make the other tests in src foundry independent as well?

nishim3: > or `deltaBefore0 not equal to 0`
I have gone with this template for the error messages.

src/test/PoolTestBase.sol
```diff
@@ -27,7 +26,7 @@ abstract contract PoolTestBase is Test, IUnlockCallback {
     }
 
     function _settle(Currency currency, address payer, int128 amount, bool settleUsingTransfer) internal {
-        assertLt(amount, 0);
+        require(amount < 0, "amount is not lesser than zero");
```
hensha256, line 29:
```suggestion
        require(amount < 0, "amount is not less than zero");
```

src/test/PoolTakeTest.sol
```diff
@@ -39,18 +38,18 @@ contract PoolTakeTest is Test, PoolTestBase {
...
+        require(deltaAfter == -amount.toInt128(), "deltaAfter is not equal to -amount.toInt128()");
 
-        assertEq(userBalAfter - userBalBefore, amount);
-        assertEq(pmBalBefore - pmBalAfter, amount);
-        assertEq(reserveBefore - reserveAfter, amount);
-        assertEq(reserveBefore - reserveAfter, amount);
+        require(userBalAfter - userBalBefore == amount, "userBalAfter - userBalBefore is not equal to amount");
+        require(pmBalBefore - pmBalAfter == amount, "pmBalBefore - pmBalAfter is not equal to amount");
+        require(reserveBefore - reserveAfter == amount, "reserveBefore - reserveAfter is not equal to amount");
+        require(reserveBefore - reserveAfter == amount, "reserveBefore - reserveAfter is not equal to amount");
```
hensha256, line 52: my this line is identical to the line above, can you remove it please?

src/test/PoolSwapTest.sol
```diff
@@ -58,36 +57,48 @@ contract PoolSwapTest is Test, PoolTestBase {
...
+                require(
+                    deltaAfter0 == data.params.amountSpecified,
+                    "deltaAfter0 is not equal to data.params.amountSpecified"
+                );
+                require(deltaAfter1 > 0, "deltaAfter1 is not greater than 0");
             } else {
                 // exact output, 0 for 1
-                assertLt(deltaAfter0, 0);
-                assertEq(deltaAfter1, data.params.amountSpecified);
+                require(deltaAfter0 < 0, "deltaAfter0 is not lesser than zero");
```
hensha256, line 81: another "lesser" here - please can you replace all the "lesser than" with "less than" ?

hensha256 (approved):

Issue #230, opened 2023-06-13
EIP-173 for Owned.sol


z0r0z:
### Describe the suggested feature and problem it solves.

Contract ownership discoverability benefit. Reasons no?

zerosnacks:
It would also be recommended to use the 2-step ownership model to prevent any mistakes during transfer: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable2Step.sol or if you want to cut down on gas https://github.com/Vectorized/solady/blob/main/src/auth/Ownable.sol

PR #520, opened 2024-03-20
Use solmate owned

hensha256:
## Related Issue
#230 to meet EIP, and also good to rely on a battletested implementation

## Description of changes


.forge-snapshots/swap with hooks.snap
```diff
@@ -1 +1 @@
-104143
\ No newline at end of file
+103987
```
hensha256, line 1: it just makes no sense 😭

saucepoint: wonder if running `forge test --isolate` before and after this change will still introduce a diff

hensha256: good q i'll try it

hensha256: nope sadly they all still change - different values ofc but they still all change by like <200 ish. Could be that thats an expected result but its definitely weird

snreynolds (commented):
We don't currently have a test for this in main but since we're changing this code I think we may as well add one now and it will check the onlyOwner code in the dependency... But we should really be testing that `setProtocolFeeController` reverts if called not by the owner

test/PoolManager.t.sol
```diff
@@ -78,6 +77,16 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
         assertEq(address(manager.protocolFeeController()), address(feeController));
     }
 
+    function test_setProtocolFeeController_failsIfNotOwner() public {
```
snreynolds, line 80: technically i think this will be moved out of this file (like a lot of other tests tbh) but since we haven't migrated the tests out of here I think its fine for this PR. and whoever is assigned with ProtocolFees.t.sol will move it (I think its me so I can do it when I pick that up).

snreynolds (dismissed):

snreynolds (approved):

PR #522, opened 2024-03-20
Optimize settle

snreynolds:
## Related Issue

## Description of changes


src/PoolManager.sol
```diff
@@ -270,9 +270,14 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
         onlyWhenUnlocked
         returns (uint256 paid)
     {
-        uint256 reservesBefore = reservesOf[currency];
-        reservesOf[currency] = currency.balanceOfSelf();
-        paid = reservesOf[currency] - reservesBefore;
+        if (currency.isNative()) {
+            paid = msg.value;
```
hensha256, line 274: you should also remove the receive function so that ETH is less likely to be sent in outside this call

hensha256: and have you double checked there’s no other payable functions anywhere?

snreynolds: good call - removed it from `unlock`. and then removed `receive`

hensha256 (approved):

PR #529, opened 2024-03-26
fuzz test doesnt check when SwapFeeLibrary.STATIC_FEE_MASK = 1000000

dianakocsis:


snreynolds:
forge to improve dictionary to catch this case

snreynolds:
tracking the issue here! https://github.com/foundry-rs/foundry/issues/7551

Issue #416, opened 2023-11-20
Update tests to reference `IPoolManager`


saucepoint:
### Component

General design optimization (improving efficiency, cleanliness, or developer experience)

### Describe the suggested feature and problem it solves.

Following up on https://github.com/Uniswap/v4-core/pull/412, the `IPoolManager` was found to be out of sync with the external signatures in `PoolManager`

This issue suggests updating the tests to interact with the PoolManager through its `IPoolManager`

### Describe the desired implementation.

Update tests that directly use `PoolManager` to use `IPoolManager` instead

### Describe alternatives.

N/A

nishim3:
I have gone through the repo, and I cannot see any tests where the ```PoolManager``` is being called instead of the ```IPoolManager``` contract. I think this issue is resolved now.

saucepoint:
looks like most tests were refactored to use the `Deployers` test helpers which is a good start. But it looks like `Deployers` is still using PoolManager and not the interface.

while I think this issue should be resolved before deployment, I think preemptively using the interface adds friction to core contributions (changes to function signatures need to be updated twice)

PR #531, opened 2024-03-26
update tests to use the IPoolManager

dianakocsis:
## Related Issue
Update tests to reference IPoolManager #416

## Description of changes
have tests interact with the PoolManager through its IPoolManager


src/interfaces/IProtocolFees.sol
```diff
@@ -2,14 +2,13 @@
 pragma solidity ^0.8.19;
 
 import {Currency} from "../types/Currency.sol";
+import {IProtocolFeeController} from "./IProtocolFeeController.sol";
 
 interface IProtocolFees {
     /// @notice Thrown when not enough gas is provided to look up the protocol fee
     error ProtocolFeeCannotBeFetched();
     /// @notice Thrown when the call to fetch the protocol fee reverts or returns invalid data.
     error ProtocolFeeControllerCallFailedOrInvalidResult();
```
dianakocsis, line 11: This only gets thrown in PoolManager. Wasn't sure if we should move it to the IPoolManager interface or keep it in IProtocolFees since it's related to protocol fees

hensha256: good question... i guess in IPoolManager because its not thrown in ProtocolFees

test/PoolManagerInitialize.t.sol
```diff
@@ -49,7 +49,7 @@ contract PoolManagerInitializeTest is Test, Deployers, GasSnapshot {
         });
     }
 
-    function test_initialize(PoolKey memory key0, uint160 sqrtPriceX96) public {
+    function test_initialize_fuzz(PoolKey memory key0, uint160 sqrtPriceX96) public {
```
snreynolds, line 52: test_fuzz_initialize just so that we can do things like `forge test --mt test_fuzz `to run all fuzz

src/interfaces/IProtocolFees.sol
snreynolds: Really what we need to do is move any protocol fee related code into ProtocolFees so then it can be in IProtocolFees. Im happy to have this be resolved in a separate issue though

dianakocsis: I was looking into it and its tough with our structure right now since `setProtocolFee` in PoolManager.sol calls `setProtocolFee` on the pool. So if we moved that function into ProtocolFees, ProtocolFees would need to have access to the pools mapping

saucepoint (commented):
We should probably update `AmountHelpers.getMaxAmountInForPool`, which looks like its still referencing `PoolManager` and not the interface. or just remove it entirely?

its not used by anything / is a dead import on PoolManager.t.sol, and it might be buggy since its using `getAmount0ForLiquidity` for amount1:
```solidity
amount1 = LiquidityAmounts.getAmount0ForLiquidity(sqrtPriceX96Upper, sqrtPriceX96, liquidity);
```

test/PoolManagerInitialize.t.sol
```diff
@@ -316,10 +317,10 @@ contract PoolManagerInitializeTest is Test, Deployers, GasSnapshot {
         );
         manager.initialize(uninitializedKey, sqrtPriceX96, ZERO_BYTES);
         // protocol fees should default to 0
-        (Pool.Slot0 memory slot0,,,) = manager.pools(uninitializedKey.toId());
-        assertEq(slot0.protocolFee, 0);
+        (,, uint16 slot0ProtocolFee,) = manager.getSlot0(key.toId());
```
saucepoint, line 320: did you mean to replace `uninitializedKey` with `key` here?

test/PoolManagerInitialize.t.sol
```diff
@@ -340,8 +341,8 @@ contract PoolManagerInitializeTest is Test, Deployers, GasSnapshot {
...
```
saucepoint, line 344: same

test/PoolManagerInitialize.t.sol
```diff
@@ -361,8 +362,8 @@ contract PoolManagerInitializeTest is Test, Deployers, GasSnapshot {
...
```
saucepoint, line 365: same as above

test/PoolManagerInitialize.t.sol
```diff
@@ -382,8 +383,8 @@ contract PoolManagerInitializeTest is Test, Deployers, GasSnapshot {
...
```
saucepoint, line 386: same

test/PoolManagerInitialize.t.sol
dianakocsis: yess thanks

src/interfaces/IProtocolFees.sol
hensha256: yeah +1 to diana she showed me what that PR would look like and it was pretty horrified to have to move the pools mapping into the protocol fees contract 😢

snreynolds: Note that we chatted offline to try this in a sep PR.

snreynolds (approved):

PR #534, opened 2024-03-26
add swap fee lib tests

snreynolds:
## Related Issue

## Description of changes


test/libraries/SwapFeeLibrary.t.sol
```diff
@@ -0,0 +1,77 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+import "src/libraries/SwapFeeLibrary.sol";
+import "forge-std/Test.sol";
+
+contract SwapFeeLibraryTest is Test {
+    function test_isDynamicFee_returnsTrue() public {
+        uint24 dynamicFee = 0x800000;
```
hensha256, line 9: i know its tested in the fuzz but i would love to see this test also do `0xFFFFFF` is true. Like to show explicitly that only that 1 bit matters.

test/libraries/SwapFeeLibrary.t.sol
```diff
@@ -0,0 +1,77 @@
...
+        assertTrue(SwapFeeLibrary.isDynamicFee(dynamicFee));
+    }
+
+    function test_isDynamicFee_returnsFalse() public {
+        uint24 dynamicFee = 0x7FFFFF;
+        assertFalse(SwapFeeLibrary.isDynamicFee(dynamicFee));
+    }
+
+    function test_fuzz_isDynamicFee(uint24 fee) public {
+        if (fee >> 23 == 1) {
```
hensha256, line 19: could rewrite this fuzz on 1 line as just the following
```
assertEq(fee >> 23 == 1, SwapFeeLibrary.isDynamicFee(fee))
```

what do you think?

test/libraries/SwapFeeLibrary.t.sol
```diff
@@ -0,0 +1,77 @@
...
+            assertTrue(SwapFeeLibrary.isDynamicFee(fee));
+        } else {
+            assertFalse(SwapFeeLibrary.isDynamicFee(fee));
+        }
+    }
+
+    function test_validate_doesNotRevert() public pure {
+        uint24 fee = 500000; // 50%
```
hensha256, line 27: maybe explicitly test 0? just good to get edge case tests in there

test/libraries/SwapFeeLibrary.t.sol
```diff
@@ -0,0 +1,77 @@
...
+        SwapFeeLibrary.validate(fee);
+    }
+
+    function test_fuzz_validate(uint24 fee) public {
+        if (fee > 1000000) {
+            vm.expectRevert(SwapFeeLibrary.FeeTooLarge.selector);
+            SwapFeeLibrary.validate(fee);
+        } else {
+            SwapFeeLibrary.validate(fee);
+        }
```
hensha256, line 48: can simplify to
```
        if (fee > 1000000) {
            vm.expectRevert(SwapFeeLibrary.FeeTooLarge.selector);
        }
        SwapFeeLibrary.validate(fee);
```

test/libraries/SwapFeeLibrary.t.sol
```diff
@@ -0,0 +1,77 @@
...
+        assertEq(SwapFeeLibrary.getSwapFee(staticFee), staticFee);
+    }
+
+    function test_getSwapFee_revertsWithFeeTooLarge_forStaticFee() public {
+        uint24 staticFee = 1000001;
+        vm.expectRevert(SwapFeeLibrary.FeeTooLarge.selector);
+        SwapFeeLibrary.getSwapFee(staticFee);
+    }
+
+    function test_getSwapFee_forDynamicFeeIsZero() public {
```
hensha256, line 62: not specifically related to your PR but it feels off for `getSwapFee` to return 0 for dynamic fee pools. I then went back and checked why and its because this function is only used in `initialized`. I wonder if we should rename it `getInitialSwapFee`? so that its more explicit that thats the only time it should be used?

hensha256 (commented):
looking lovely!! Left a bunch of comments but i think some of them just come down to what style we choose to code these tests in - e.g. adding a few more explicit edge case tests outside of fuzzing

test/libraries/SwapFeeLibrary.t.sol
snreynolds: gotcha - sure! fine to add non fuzzs on edge case values

test/libraries/SwapFeeLibrary.t.sol
snreynolds: YES I actually had the same thought process. Let me write up an issue.

snreynolds: glad you agree haha 😄

hensha256 (approved):

Issue #538, opened 2024-03-27
Use CurrencyDelta for Currency


snreynolds:
### Describe the suggested feature and problem it solves.

I think it would be a readability improvement if we used the CurrencyDelta library with the other syntax, especially now that we are passing in a target address.

ie instead of `CurrencyDelta.get(target, currency);` we could do `currency.getDeltaFor(target)`;

nishim3:
I want to try my hand at this. Can you assign it to me please?

PR #539, opened 2024-03-27
Use CurrencyDelta for Currency

nishim3:
## Related Issue
#538 
## Additional Comments
Would it be a good idea to make CurrencyDelta a part of the Currency library? I think it is not the best to use two libraries for the same type...


eth0izzle (approved):

hensha256:
I think for now we will keep the libraries separate but will consider it in future

hensha256 (approved):

PR #542, opened 2024-03-29
Uncaught foundry assertion

dianakocsis:
hard code test case not always caught with fuzzing


test/libraries/Pool.t.sol
```diff
@@ -126,4 +126,58 @@ contract PoolTest is Test {
             assertLe(state.slot0.sqrtPriceX96, params.sqrtPriceLimitX96);
         }
     }
+
+    function test_unit_Swap() public {
+        // Assumptions tested in PoolManager.t.sol
+        uint160 sqrtPriceX96 = 79228162514264337593543950336;
+        uint24 swapFee = 0;
+        Pool.SwapParams memory params = Pool.SwapParams({tickSpacing: -2448282, zeroForOne: false, amountSpecified: 2459, sqrtPriceLimitX96: 183941092569902643409258474997558974124631523329 });
+        params.tickSpacing = int24(bound(params.tickSpacing, TickMath.MIN_TICK_SPACING, TickMath.MAX_TICK_SPACING));
```
dianakocsis, line 135: Since tickSpacing is always going to be positive (since MIN_TICK_SPACING is 1), should we make tickSpacing in SwapParams a uint24 instead of an int24?

PR #543, opened 2024-04-01
Nonzero delta count test

dianakocsis:
## Related Issue
- review of Nonzero delta count test

## Description of changes
remove unnecessary constants and add more asserts


snreynolds (approved):
looks good unit testing wise. I think when we move to integ testing we will need to add a test that asserts the count is 0 before and after a lock, but definitely not fit for this unit test

PR #545, opened 2024-04-02
clean up delegatecall unit tests

snreynolds:
## Related Issue

## Description of changes

### DIFF
--- a/src/test/NoDelegateCallTest.sol
+++ b/src/test/NoDelegateCallTest.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.20;
 
-import {NoDelegateCall} from "../NoDelegateCall.sol";
+import {NoDelegateCall} from "src/NoDelegateCall.sol";
 
 contract NoDelegateCallTest is NoDelegateCall {
     function canBeDelegateCalled() public view returns (uint256) {
--- a/test/NoDelegateCall.t.sol
+++ b/test/NoDelegateCall.t.sol
@@ -3,18 +3,17 @@ pragma solidity ^0.8.20;
 
 import {GasSnapshot} from "forge-gas-snapshot/GasSnapshot.sol";
 import {Test} from "forge-std/Test.sol";
-import {NoDelegateCallTest} from "../src/test/NoDelegateCallTest.sol";
+import {NoDelegateCallTest} from "src/test/NoDelegateCallTest.sol";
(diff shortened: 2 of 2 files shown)


test/NoDelegateCall.t.sol
```diff
@@ -3,45 +3,44 @@ pragma solidity ^0.8.20;
...
-    error DelegateCallNotAllowed();
-
     NoDelegateCallTest noDelegateCallTest;
 
     function setUp() public {
         noDelegateCallTest = new NoDelegateCallTest();
     }
 
-    function testGasOverhead() public {
+    function test_gas_noDelegateCallOverhead() public {
```
hensha256, line 16:
```suggestion
    function test_gas_noDelegateCall() public {
```

src/test/NoDelegateCallTest.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.20;
 
-import {NoDelegateCall} from "../NoDelegateCall.sol";
+import {NoDelegateCall} from "src/NoDelegateCall.sol";
```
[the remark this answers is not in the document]
snreynolds: good catch!

test/NoDelegateCall.t.sol
```diff
@@ -4,44 +4,43 @@ pragma solidity ^0.8.20;
...
     }
 
-    function testCanCallIntoPrivateMethodWithModifier() public view {
+    function test_callPrivateMethodWithModifer_succeeds() public view {
         noDelegateCallTest.callsIntoNoDelegateCallFunction();
     }
 
-    function testCannotDelegateCallPrivateMethodWithModifier() public {
-        vm.expectRevert(DelegateCallNotAllowed.selector);
+    function test_delegateCallPrivateMethodWithModifier_revertsWithDelegateCallNotAllowed() public {
```
hensha256, line 42: i feel like the naming of this test and the test above are confusing. If a method is private, by definition you cannot delegate call it because its not visible externally... so what are these tests doing?

snreynolds: Tried a new name. Its a bit verbose so lmk if you have a suggestion but it's testing that a private function with the delegatecall modifier will succeed (without a delegatecall) or revert (when there is a delegatecall). There is an external entrypoint that then calls the private function.

hensha256: ohhh i see 😅 thats a tough test to name lol

hensha256 (approved):

PR #547, opened 2024-04-02
make protocol fee additive

dianakocsis:
## Related Issue
Apply fee on top of rather than subtracted from swap fee #490


test/libraries/Pool.t.sol
```diff
@@ -76,14 +81,17 @@ contract PoolTest is Test {
         state.modifyLiquidity(params);
     }
 
-    function testSwap(uint160 sqrtPriceX96, uint24 swapFee, Pool.SwapParams memory params) public {
+    function testSwap(uint160 sqrtPriceX96, uint24 swapFee, uint24 protocolFee, Pool.SwapParams memory params) public {
         // Assumptions tested in PoolManager.t.sol
         params.tickSpacing = int24(bound(params.tickSpacing, TickMath.MIN_TICK_SPACING, TickMath.MAX_TICK_SPACING));
```
dianakocsis, line 86: Since tickSpacing is always going to be positive (since MIN_TICK_SPACING is 1), should we make tickSpacing in SwapParams a uint24 instead of an int24?

test/libraries/Pool.t.sol
```diff
@@ -76,14 +81,17 @@ contract PoolTest is Test {
...
-        swapFee = uint24(bound(swapFee, 0, 999999));
+        swapFee = uint24(bound(swapFee, 0, 1000000));
+        protocolFee = uint24(bound(protocolFee, 0, 0x3E83E8));
```
saucepoint, line 88: i see the magic hex is `(1000 << 12) | 1000`, but we should use the MAX_FEE_BOTH_TOKENS constant here for readability

src/libraries/SwapMath.sol
```diff
@@ -80,9 +97,12 @@ library SwapMath {
...
             if (exactIn && sqrtRatioNextX96 != sqrtRatioTargetX96) {
                 // we didn't reach the target, so take the remainder of the maximum input as fee
-                feeAmount = uint256(-amountRemaining) - amountIn;
+                totalFeeAmount = uint256(-amountRemaining) - amountIn;
             } else {
-                feeAmount = FullMath.mulDivRoundingUp(amountIn, feePips, 1e6 - feePips);
+                if (protocolFeePips > 0) {
+                    protocolFeeAmount = FullMath.mulDivRoundingUp(amountIn, protocolFeePips, 1e6 - protocolFeePips);
+                }
+                totalFeeAmount = FullMath.mulDivRoundingUp(amountIn - protocolFeeAmount, swapFeePips, 1e6 - swapFeePips);
```
saucepoint, line 105: the natspec comment indicates that `totalFeeAmount` is swap fee + protocol fee. this line doesnt smell right if `totalFeeAmount` is supposed to be the sum of both?

If it is supposed to be the sum of both fees, you're also adding protocol fees to LP fees in Pool.sol?

i.e.:
```
state.feeGrowthGlobalX128 += FullMath.mulDiv(step.totalFeeAmount, FixedPoint128.Q128, state.liquidity);
```

dianakocsis: It is supposed to be sum of both! First the protocol fee is taken off and then the swap fee is taken afterwards on the remaining. I'll fix it so it's not also adding the protocol fees to LP fees in Pool.sol

test/libraries/Pool.t.sol
dianakocsis: 👍

src/libraries/SwapMath.sol
```diff
@@ -102,7 +102,8 @@ library SwapMath {
                 if (protocolFeePips > 0) {
```
dianakocsis, line 102: if `exactIn && sqrtRatioNextX96 != sqrtRatioTargetX96`, the whole fee is going to LP and nothing to protocol, which is why some tests are failing. How should protocol fee be calculated here?

src/libraries/SwapMath.sol
```diff
@@ -96,12 +97,12 @@ library SwapMath {
 
             if (exactIn && sqrtRatioNextX96 != sqrtRatioTargetX96) {
```
dianakocsis, line 98: if exactIn && sqrtRatioNextX96 != sqrtRatioTargetX96, the whole fee is going to LP and nothing to protocol, which is why some tests are failing. How should protocol fee be calculated here?

src/libraries/Pool.sol
```diff
@@ -61,14 +61,17 @@ library Pool {
...
         // the current price
         uint160 sqrtPriceX96;
         // the current tick
         int24 tick;
-        // protocol swap fee, taken as a % of the LP swap fee
+        // protocol fee, expressed in hundredths of a bip
         // upper 12 bits are for 1->0, and the lower 12 are for 0->1
-        // the maximum is 2500 - meaning the maximum protocol fee is 25%
+        // the maximum is 1000 - meaning the maximum protocol fee is 0.1%
         uint24 protocolFee;
```
hensha256, line 75: maybe another comment like
```
// the protocolFee is taken from the input first, then the swapFee is taken from the remaining input
```

feel free to change as you like!

src/libraries/Pool.sol
```diff
@@ -381,21 +385,17 @@ library Pool {
...
 
-            // if the protocol fee is on, calculate how much is owed, decrement feeAmount, and increment protocolFee
             if (cache.protocolFee > 0) {
-                // calculate the amount of the fee that should go to the protocol
-                uint256 delta = step.feeAmount * cache.protocolFee / ProtocolFeeLibrary.BIPS_DENOMINATOR;
-                // subtract it from the regular fee and add it to the protocol fee
                 unchecked {
-                    step.feeAmount -= delta;
-                    feeForProtocol += delta;
+                    feeForProtocol += FullMath.mulDiv(step.feeAmount, cache.protocolFee, swapFee + cache.protocolFee);
```
hensha256, line 390: I think we can actually leave this block pretty similar to before 
```
// calculate the amount of the fee that should go to the protocol
uint256 delta = FullMath.mulDiv(...)
// subtract it from the regular fee and add it to the protocol fee
unchecked {
       step.feeAmount -= delta;
       feeForProtocol += delta;
}
```

That way we then just have the swap fee left in step.feeAmount and the block below can be left the same too! The protocol fee and swap fee have to add to the total step.feeAmount so no need to calculate them both separately :)

src/libraries/ProtocolFeeLibrary.sol
```diff
@@ -2,11 +2,11 @@
...
 library ProtocolFeeLibrary {
-    // Max protocol fee is 25% (2500 bips)
-    uint16 public constant MAX_PROTOCOL_FEE = 2500;
+    // Max protocol fee is 0.1%
+    uint16 public constant MAX_PROTOCOL_FEE = 1000;
 
-    // Total bips
-    uint16 internal constant BIPS_DENOMINATOR = 10_000;
+    // the protocol fee is represented in hundredths of a bip
+    uint24 internal constant BIPS_DENOMINATOR = 1_000_000;
```
hensha256, line 9: should be PIPS now not BIPS

src/libraries/Pool.sol
```diff
@@ -365,7 +368,8 @@ library Pool {
                 ) ? params.sqrtPriceLimitX96 : step.sqrtPriceNextX96,
                 state.liquidity,
                 state.amountSpecifiedRemaining,
-                swapFee
+                // use the effective fee
+                uint24(cache.protocolFee + swapFee - (uint32(cache.protocolFee) * uint32(swapFee)) / MAX_EFFECTIVE_FEE)
```
hensha256, line 372: i wonder if this should be like
`cache.protocolFee.calculateEffectiveFee(swapFee)`
and put it in the ProtocolFeeLibrary?

and then we can get rid of `MAX_EFFECTIVE_FEE` and just use `PIPS_DENOMINATOR` instead?

Although it feels weird to have to choose between ProtocolFeeLibrary and SwapFeeLibrary here hahah

test/libraries/Pool.t.sol
```diff
@@ -17,7 +17,10 @@ contract PoolTest is Test {
 
     Pool.State state;
 
-    function testPoolInitialize(uint160 sqrtPriceX96, uint16 protocolFee, uint24 dynamicFee) public {
+    uint24 constant MAX_PROTOCOL_FEE = 1000; // 0.1%
```
hensha256, line 20: would rather this was
```
uint24 constant MAX_PROTOCOL_FEE = ProtocolFeeLibrary. MAX_PROTOCOL_FEE
```

So that if ever it changes it updates here too

Same with the one below and SwapFeeLibrary.MAX_SWAP_FEE

hensha256 (changes requested):
Looking good! Left some thoughts!

src/libraries/Pool.sol
dianakocsis: hmm yeah that feels weird putting it in the protocol fee library because it's not just about protocol fees

src/libraries/Pool.sol
```diff
@@ -365,7 +369,8 @@ library Pool {
...
```
hensha256, line 373: this whole sum can be unchecked - that will save a good amount of gas. Currently this PR adds 1000 gas to a simple swap 😢

src/libraries/Pool.sol
```diff
@@ -384,7 +389,7 @@ library Pool {
             // if the protocol fee is on, calculate how much is owed, decrement feeAmount, and increment protocolFee
             if (cache.protocolFee > 0) {
                 // calculate the amount of the fee that should go to the protocol
-                uint256 delta = step.feeAmount * cache.protocolFee / ProtocolFeeLibrary.BIPS_DENOMINATOR;
+                uint256 delta = FullMath.mulDiv(step.feeAmount, cache.protocolFee, swapFee + cache.protocolFee);
```
hensha256, line 392: does this need mulDiv? I feel like it shouldnt be overflowing uint256

src/libraries/ProtocolFeeLibrary.sol
```diff
@@ -20,11 +20,17 @@ library ProtocolFeeLibrary {
...
             if ((fee0 > MAX_PROTOCOL_FEE) || (fee1 > MAX_PROTOCOL_FEE)) {
                 return false;
             }
         }
         return true;
     }
+
+    function calculateEffectiveFee(uint24 self, uint24 swapFee) internal pure returns (uint24) {
+        unchecked {
+            return uint24(self + swapFee - (uint32(self) * uint32(swapFee)) / PIPS_DENOMINATOR);
```
shuhuiluo, line 33:
```suggestion
            return uint24(uint256(self) + swapFee - (uint256(self) * swapFee) / PIPS_DENOMINATOR);
```
Saves 42 gas.

src/libraries/Pool.sol
shuhuiluo: `swapFee + cache.protocolFee` can be unchecked.

src/libraries/Pool.sol
```diff
@@ -365,7 +369,8 @@ library Pool {
                 ) ? params.sqrtPriceLimitX96 : step.sqrtPriceNextX96,
                 state.liquidity,
                 state.amountSpecifiedRemaining,
-                swapFee
+                // use the effective fee
+                cache.protocolFee == 0 ? swapFee : uint24(cache.protocolFee).calculateEffectiveFee(swapFee)
```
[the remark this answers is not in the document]
dianakocsis: cache.protocolFee is a uint16, but the library is for uint24

src/libraries/ProtocolFeeLibrary.sol
dianakocsis: thanks!

src/libraries/ProtocolFeeLibrary.sol
```diff
@@ -2,11 +2,11 @@
 pragma solidity ^0.8.20;
 
 library ProtocolFeeLibrary {
-    // Max protocol fee is 25% (2500 bips)
-    uint16 public constant MAX_PROTOCOL_FEE = 2500;
+    // Max protocol fee is 0.1%
```
snreynolds, line 5: could you also add comment here that its denominated in pips

src/libraries/ProtocolFeeLibrary.sol
```diff
@@ -20,11 +20,17 @@ library ProtocolFeeLibrary {
         if (self != 0) {
             uint16 fee0 = getZeroForOneFee(self);
             uint16 fee1 = getOneForZeroFee(self);
-            // The fee is represented in bips so it cannot be GREATER than the MAX_PROTOCOL_FEE.
+            // The fee is represented in pips so it cannot be GREATER than the MAX_PROTOCOL_FEE.
```
snreynolds, line 23:
```suggestion
            // The fee is represented in pips and it cannot be greater than the MAX_PROTOCOL_FEE.
```

i think some of the punctuation/language is leftover from old protocol comments. this seems better to me bc its not due to the pips denomination, and is no longer a denominator so specifying GREATER in all caps is funny now haha

test/libraries/ProtocolFeeLibrary.t.sol
```diff
@@ -0,0 +1,58 @@
...
+        assertTrue(ProtocolFeeLibrary.validate(fee));
+    }
+
+    function test_fuzz_calculateEffectiveFeeDoesNotOverflow(uint24 self, uint24 swapFee) public {
+        swapFee = uint24(bound(swapFee, 0, SwapFeeLibrary.MAX_SWAP_FEE));
+        self = uint24(bound(self, 0, ProtocolFeeLibrary.MAX_PROTOCOL_FEE));
+        assertGe(ProtocolFeeLibrary.calculateEffectiveFee(self, swapFee), swapFee);
+    }
+
+    function test_calculateEffectiveFee() public {
```
snreynolds, line 50: can you add one more non fuzz where swap fee is 0

src/libraries/ProtocolFeeLibrary.sol
dianakocsis: ✅

src/libraries/SwapMath.sol
```diff
@@ -82,7 +82,7 @@ library SwapMath {
                 // we didn't reach the target, so take the remainder of the maximum input as fee
                 feeAmount = uint256(-amountRemaining) - amountIn;
             } else {
-                feeAmount = FullMath.mulDivRoundingUp(amountIn, feePips, 1e6 - feePips);
+                feeAmount = feePips == 1e6 ? amountIn : FullMath.mulDivRoundingUp(amountIn, feePips, 1e6 - feePips);
```
shuhuiluo, line 85: Consider an exact output swap with fixed input args except `feePips`.
A reasonable expectation would be

> The fee amount monotonically increases as `feePips` increases.

On Ln 85, `amountIn` is determined by `sqrtRatioCurrentX96`, `sqrtRatioTargetX96`, `liquidity` and `amountRemaining`. So given a fixed `amountIn`, `feeAmount` is only a function of `feePips`. Under the current implementation:

- If `feePips == 0`, `feeAmount == 0`.
- If `feePips == 5e5`, `feeAmount == amountIn`.
- If `feePips == 1e6 - 1`, `feeAmount == 999999 * amountIn`.
- If `feePips == 1e6`, `feeAmount == amountIn`.

It breaks the above expectation. It also enables the change to pool price when the fee is 100%, which is disabled for the `exactIn` case.

dianakocsis: Thanks @shuhuiluo we disabled the case for 100% swap fee on exactOut

src/libraries/Pool.sol
```diff
@@ -384,7 +394,7 @@ library Pool {
             // if the protocol fee is on, calculate how much is owed, decrement feeAmount, and increment protocolFee
             if (cache.protocolFee > 0) {
                 // calculate the amount of the fee that should go to the protocol
-                uint256 delta = step.feeAmount * cache.protocolFee / ProtocolFeeLibrary.BIPS_DENOMINATOR;
+                uint256 delta = step.feeAmount * cache.protocolFee / (swapFee + cache.protocolFee);
```
hensha256, line 397: i feel like this can be unchecked? i guess it wasnt unchecked before either...
but `max(uint256) * max(protocolFee)` is not going to overflow
and `swapFee + protocolFee` is also not going to overflow?

test/libraries/ProtocolFeeLibrary.t.sol
```diff
@@ -0,0 +1,70 @@
...
+    function test_fuzz_calculateEffectiveFeeDoesNotOverflow(uint24 self, uint24 swapFee) public {
+        swapFee = uint24(bound(swapFee, 0, SwapFeeLibrary.MAX_SWAP_FEE));
+        self = uint24(bound(self, 0, ProtocolFeeLibrary.MAX_PROTOCOL_FEE));
+        assertGe(ProtocolFeeLibrary.calculateEffectiveFee(self, swapFee), swapFee);
+    }
+
+    function test_fuzz_calculateEffectiveFeeNeverEqualsMax(uint24 self, uint24 swapFee) public {
+        swapFee = uint24(bound(swapFee, 0, SwapFeeLibrary.MAX_SWAP_FEE - 1));
+        self = uint24(bound(self, 0, ProtocolFeeLibrary.MAX_PROTOCOL_FEE));
+        assertLt(ProtocolFeeLibrary.calculateEffectiveFee(self, swapFee), SwapFeeLibrary.MAX_SWAP_FEE);
```
snreynolds, line 53: Can't it be <= ?

test/libraries/ProtocolFeeLibrary.t.sol
```diff
@@ -0,0 +1,70 @@
...
+        assertEq(ProtocolFeeLibrary.calculateEffectiveFee(self, swapFee), SwapFeeLibrary.MAX_SWAP_FEE);
+
+        swapFee = 3000;
+        assertEq(ProtocolFeeLibrary.calculateEffectiveFee(self, swapFee), 3997);
+
+        swapFee = 0;
+        assertEq(ProtocolFeeLibrary.calculateEffectiveFee(self, swapFee), ProtocolFeeLibrary.MAX_PROTOCOL_FEE);
+
+        self = 0;
+        assertEq(ProtocolFeeLibrary.calculateEffectiveFee(self, swapFee), 0);
```
snreynolds, line 68: would love to see one more accuracy test here that is a non zero swap fee but 0 protocol fee
& maybe one for the minimum value of the effective fee that is non zero?

src/libraries/Pool.sol
snreynolds, line 397: I think this is incorrect in the case when swapFee + cache.protocolFee > 100%. I think the denominator should actually be the "effective fee". Or am I misunderstanding this calculation?

src/libraries/ProtocolFeeLibrary.sol
```diff
@@ -20,11 +22,19 @@ library ProtocolFeeLibrary {
...
-            // The fee is represented in bips so it cannot be GREATER than the MAX_PROTOCOL_FEE.
+            // The fee is represented in pips and it cannot be greater than the MAX_PROTOCOL_FEE.
             if ((fee0 > MAX_PROTOCOL_FEE) || (fee1 > MAX_PROTOCOL_FEE)) {
                 return false;
             }
         }
         return true;
     }
+
+    function calculateEffectiveFee(uint24 self, uint24 swapFee) internal pure returns (uint24) {
```
snreynolds, line 33: Can you also add natspec here that the effective fee can never be more than 100%

snreynolds: (or MAX SWAP FEE)

src/libraries/Pool.sol
shuhuiluo: `swapFee + protocolFee` may overflow unless written as `uint256(swapFee) + protocolFee`.

shuhuiluo: If I understand correctly, it goes like this.
```
feeAmount = amountIn - amountIn * (1 - swapFee) * (1 - protocolFee) = amountIn * effectiveFee
effectiveFee = swapFee + protocolFee - swapFee * protocolFee
delta = amountIn * (1 - swapFee) * protocolFee = feeAmount * (1 - swapFee) * protocolFee / effectiveFee
```

test/libraries/ProtocolFeeLibrary.t.sol
dianakocsis: i dont think we want effective fee to ever be 100% though if swap fee is not 100%

test/DynamicFees.t.sol
```diff
@@ -149,6 +150,130 @@ contract TestDynamicFees is Test, Deployers, GasSnapshot {
         assertEq(_fetchPoolSwapFee(key), 123);
     }
 
+    function test_updateDynamicSwapFee_100PercentFee_AmountIn() public {
+        assertEq(_fetchPoolSwapFee(key), 0);
+
+        dynamicFeesHooks.setFee(1000000);
+
+        IPoolManager.SwapParams memory params =
+            IPoolManager.SwapParams({zeroForOne: true, amountSpecified: -100, sqrtPriceLimitX96: SQRT_RATIO_1_2});
```
hensha256, line 159: would love for these tests to have a higher amountSpecified. Just because this is swapping 100 wei of a token which is ~nothing haha. Maybe 1e18?

hensha256: And same with all of them. Or even 123456789 or something so we can see what happens with odd numbers where more rounding will happen - e.g. the test below takes 50%

src/libraries/Pool.sol
```diff
@@ -365,7 +374,8 @@ library Pool {
                 ) ? params.sqrtPriceLimitX96 : step.sqrtPriceNextX96,
                 state.liquidity,
                 state.amountSpecifiedRemaining,
-                swapFee
+                // use the effective fee
+                cache.protocolFee == 0 ? swapFee : uint24(cache.protocolFee).calculateEffectiveFee(swapFee)
```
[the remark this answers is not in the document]
hensha256: yeah thought about this more, definitely we shoudl return effectiveSwapFee, and emit that in the event not just lp fee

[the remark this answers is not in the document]
dianakocsis: when i asked cody about this a few weeks ago he said the swap fee should be in the event but i agree the effective fee would make the most sense

src/libraries/Pool.sol
```diff
@@ -383,10 +393,11 @@ library Pool {
 
             // if the protocol fee is on, calculate how much is owed, decrement feeAmount, and increment protocolFee
             if (cache.protocolFee > 0) {
-                // calculate the amount of the fee that should go to the protocol
-                uint256 delta = step.feeAmount * cache.protocolFee / ProtocolFeeLibrary.BIPS_DENOMINATOR;
-                // subtract it from the regular fee and add it to the protocol fee
                 unchecked {
+                    // calculate the amount of the fee that should go to the protocol
+                    uint256 delta =
+                        step.feeAmount * cache.protocolFee / uint24(cache.protocolFee).calculateEffectiveFee(swapFee);
```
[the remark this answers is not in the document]
dianakocsis: yeah the reason i didnt do that was because there are no more variables left to use haha but since we might be removing swapfee it should work

src/libraries/ProtocolFeeLibrary.sol
```diff
@@ -20,11 +22,20 @@ library ProtocolFeeLibrary {
...
         }
         return true;
     }
+
+    // Effective fee can never exceed 100% (1e6 pips)
+    function calculateEffectiveFee(uint24 self, uint24 swapFee) internal pure returns (uint24) {
+        unchecked {
+            uint256 numerator = uint256(self) * uint256(swapFee);
+            uint256 divider = PIPS_DENOMINATOR;
+            return uint24(uint256(self) + swapFee - UnsafeMath.divRoundingUp(numerator, divider));
```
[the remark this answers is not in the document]
dianakocsis: 👍

test/DynamicFees.t.sol
```diff
@@ -149,6 +150,130 @@ contract TestDynamicFees is Test, Deployers, GasSnapshot {
...
+            PoolSwapTest.TestSettings({withdrawTokens: true, settleUsingTransfer: true, currencyAlreadySent: false});
+
+        vm.expectEmit(true, true, true, true, address(manager));
+        emit Swap(key.toId(), address(swapRouter), -1000, 0, 79228162514264337593543950336, 1e18, -1, 1000000);
+
+        snapStart("swap with dynamic fee and protocol fee");
+        swapRouter.swap(key, params, testSettings, ZERO_BYTES);
+        snapEnd();
+
+        uint256 expectedProtocolFee = uint256(-params.amountSpecified) * 1000 / 1e6;
```
shuhuiluo, line 273: If the protocol fee is taken after the swap fee is deducted, shouldn't this be zero?

dianakocsis: protocol fee gets taken first if it's on! and then the swap fee is taken from the remaining

shuhuiluo: Thanks. That makes sense.

src/libraries/Pool.sol
```diff
@@ -396,7 +396,7 @@ library Pool {
                     // calculate the amount of the fee that should go to the protocol
                     // fee amount has already been subtracted from the amountIn
                     uint256 delta =
-                        (step.amountIn + step.feeAmount) * cache.protocolFee / 1e6;
+                        (step.amountIn + step.feeAmount) * cache.protocolFee / ProtocolFeeLibrary.PIPS_DENOMINATOR;
```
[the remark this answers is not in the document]
dianakocsis: `PIPS_DENOMINATOR` is a uint24, but why would it revert if uint256?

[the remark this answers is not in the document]
shuhuiluo: I mean the increase in gas and bytecode size will be reverted if `PIPS_DENOMINATOR` is declared as a `uint256` which is how it is used.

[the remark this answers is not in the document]
dianakocsis: gotcha, switched it

src/libraries/Pool.sol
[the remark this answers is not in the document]
hensha256: lol the everlasting stack issues. Maybe its time to go to via-ir soon 😬

src/libraries/Pool.sol
```diff
@@ -383,10 +394,12 @@ library Pool {
 
             // if the protocol fee is on, calculate how much is owed, decrement feeAmount, and increment protocolFee
             if (cache.protocolFee > 0) {
-                // calculate the amount of the fee that should go to the protocol
-                uint256 delta = step.feeAmount * cache.protocolFee / ProtocolFeeLibrary.BIPS_DENOMINATOR;
-                // subtract it from the regular fee and add it to the protocol fee
                 unchecked {
+                    // calculate the amount of the fee that should go to the protocol
+                    // fee amount has already been subtracted from the amountIn
```
snreynolds, line 399: i think the (step.amountIn + step.feeAmount) addition is still a little misleading & want the comment to be ultra clear

maybe something like 
// step.amountIn does not include the swap fee. we add it back to get the total amountIn to calculate the amount of fees owed to the protocol


not really sure?? open to suggestions I just find this part of the line extremely confusing `(step.amountIn + step.feeAmount)`

src/libraries/ProtocolFeeLibrary.sol
```diff
@@ -20,11 +22,19 @@ library ProtocolFeeLibrary {
...
+            // The fee is represented in pips and it cannot be greater than the MAX_PROTOCOL_FEE.
             if ((fee0 > MAX_PROTOCOL_FEE) || (fee1 > MAX_PROTOCOL_FEE)) {
                 return false;
             }
         }
         return true;
     }
+
+    // Effective fee can never exceed 100% (1e6 pips)
+    function calculateEffectiveFee(uint24 self, uint24 swapFee) internal pure returns (uint24) {
```
snreynolds, line 34: maybe we can comment the original formula?
// the lp fee is applied after the protocol fee, and the total fee is capped at 100%
// swapFee = protocolFee * lpFee(1_000_000 - protocolFee) / 1_000_000

test/DynamicFees.t.sol
```diff
@@ -149,6 +151,186 @@ contract TestDynamicFees is Test, Deployers, GasSnapshot {
         assertEq(_fetchPoolSwapFee(key), 123);
     }
 
+    function test_updateDynamicSwapFee_100PercentFee_AmountIn() public {
```
snreynolds, line 154: is this applying a protocol fee? unclear from the test

and is it just a gas snapshot or is it just setting that the dynamic fee is set?

src/libraries/Pool.sol
```diff
@@ -328,6 +332,10 @@ library Pool {
 
         bool exactInput = params.amountSpecified < 0;
 
+        if (!exactInput && (slot0Start.swapFee == SwapFeeLibrary.MAX_SWAP_FEE)) {
+            revert CannotSpecifyOutputAmountWithMaxSwapFee();
```
snreynolds, line 336: this error is a bit long, and when writing the name for the revert test I thought it could maybe be a bit more concise:

CannotSpecifyOutputAmountWithMaxSwapFee -> InvalidFeeForExactOut()

and then the test name would be more clear:

test_swap_revertsWith_InvalidFeeForExactOut_whenFeeIsMax or something like that

test/DynamicFees.t.sol
```diff
@@ -149,6 +151,186 @@ contract TestDynamicFees is Test, Deployers, GasSnapshot {
...
+        emit Swap(key.toId(), address(swapRouter), -202, 100, 79228162514264329670727698909, 1e18, -1, 500000);
+
+        snapStart("swap with 50 Percent swap fee amountOut");
+        swapRouter.swap(key, params, testSettings, ZERO_BYTES);
+        snapEnd();
+
+        assertEq(_fetchPoolSwapFee(key), 500000);
+    }
+
+    function test_updateDynamicSwapFee_100PercentFee_AmountOut_NoProtocol() public {
```
snreynolds, line 214: this is a revert test so I'd name it something like
`test_swap_revertsWith_<ERROR>/Case`

test/libraries/ProtocolFeeLibrary.t.sol
snreynolds: oh I thought this was still fuzzing if swap fee is set to 100%. I see we have a test below that does that exact case though

test/DynamicFees.t.sol
```diff
@@ -149,6 +151,186 @@ contract TestDynamicFees is Test, Deployers, GasSnapshot {
...
+        IPoolManager.SwapParams memory params =
+            IPoolManager.SwapParams({zeroForOne: true, amountSpecified: 100, sqrtPriceLimitX96: SQRT_RATIO_1_2});
+        PoolSwapTest.TestSettings memory testSettings =
+            PoolSwapTest.TestSettings({takeClaims: false, settleUsingBurn: false});
+
+        vm.expectRevert(Pool.CannotSpecifyOutputAmountWithMaxSwapFee.selector);
+        swapRouter.swap(key, params, testSettings, ZERO_BYTES);
+    }
+
+    function test_updateDynamicSwapFee_99PercentFee_AmountOut_WithProtocol() public {
```
snreynolds, line 228: should probably get some tests that swap larger amounts? i don't see those but maybe i missed it

test/DynamicFees.t.sol
```diff
@@ -149,6 +151,186 @@ contract TestDynamicFees is Test, Deployers, GasSnapshot {
...
+            PoolSwapTest.TestSettings({takeClaims: false, settleUsingBurn: false});
+
+        vm.expectEmit(true, true, true, true, address(manager));
+        emit Swap(key.toId(), address(swapRouter), -101000000, 100, 79228162514264329670727698909, 1e18, -1, 999999);
+
+        snapStart("swap with dynamic fee and protocol fee");
+        swapRouter.swap(key, params, testSettings, ZERO_BYTES);
+        snapEnd();
+
+        uint256 expectedProtocolFee = uint256(101000000) * 1000 / 1e6;
```
snreynolds, line 248: where are you getting this value? it should be the delta right? which we return from swap.

test/DynamicFees.t.sol
dianakocsis: no protocol fee, i made it more clear

test/DynamicFees.t.sol
dianakocsis: i did fuzz testing for all int256s if that's good?

src/interfaces/IPoolManager.sol
```diff
@@ -199,7 +200,7 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload {
     function settle(Currency token) external payable returns (uint256 paid);
 
     /// @notice Updates the pools swap fees for the a pool that has enabled dynamic swap fees.
```
hensha256, line 202: comment out of date now! swap -> lp

src/libraries/Pool.sol
```diff
@@ -338,6 +346,8 @@ library Pool {
         });
 
         StepComputations memory step;
+        swapFee =
+            cache.protocolFee == 0 ? slot0Start.lpFee : uint24(cache.protocolFee).calculateSwapFee(slot0Start.lpFee);
```
hensha256, line 350: do we have any more stack room in this function? 
would love to see `slot0Start.lpFee` and `cache.protocolFee` cached. But I realise i'm probably asking for too much 💀

src/libraries/Pool.sol
```diff
@@ -383,10 +393,12 @@ library Pool {
...
             if (cache.protocolFee > 0) {
-                // calculate the amount of the fee that should go to the protocol
-                uint256 delta = step.feeAmount * cache.protocolFee / ProtocolFeeLibrary.BIPS_DENOMINATOR;
-                // subtract it from the regular fee and add it to the protocol fee
                 unchecked {
+                    // step.amountIn does not include the swap fee, as it's already been taken from it,
+                    // so add it back to get the total amountIn and use that to calculate the amount of fees owed to the protocol
+                    uint256 delta =
+                        (step.amountIn + step.feeAmount) * cache.protocolFee / ProtocolFeeLibrary.PIPS_DENOMINATOR;
+                    // subtract it from the regular fee and add it to the protocol fee
```
hensha256, line 401:
```suggestion
                    // subtract it from the total fee and add it to the protocol fee
```

src/libraries/Pool.sol
```diff
@@ -328,6 +332,10 @@ library Pool {
 
         bool exactInput = params.amountSpecified < 0;
 
+        if (!exactInput && (slot0Start.lpFee == LPFeeLibrary.MAX_LP_FEE)) {
```
hensha256, line 335: i wonder if it would make sense to just move this check to after `swapFee = ` and then check 
```
if (!exactInput && swapFee == LP...)
```

i know you fuzz that this cannot happen. But given it won't increase gas it feels like an extra safe way to do things.

test/DynamicFees.t.sol
hensha256: still would love this 👀

test/DynamicFees.t.sol
```diff
@@ -146,17 +148,186 @@ contract TestDynamicFees is Test, Deployers, GasSnapshot {
...
+
+        vm.expectEmit(true, true, true, true, address(manager));
+        emit Swap(key.toId(), address(swapRouter), -100, 49, 79228162514264333632135824623, 1e18, -1, 500000);
+
+        swapRouter.swap(key, params, testSettings, ZERO_BYTES);
+
+        assertEq(_fetchPoolLPFee(key), 500000);
+    }
+
+    function test_updateDynamicLPFee_50PercentLPFee_AmountOut_NoProtocol() public {
```
hensha256, line 190: msot of these new tests are named `test_updateDynamicLPFee_` but imo it feels like its really a test of swap as a whole not just of that function - so i think i'd rather `test_swap_`

src/libraries/Pool.sol
dianakocsis: yeah i like that. increases gas by a little, but seems safer

test/DynamicFees.t.sol
```diff
@@ -146,17 +148,186 @@ contract TestDynamicFees is Test, Deployers, GasSnapshot {
...
+
+        IPoolManager.SwapParams memory params = IPoolManager.SwapParams({
+            zeroForOne: true,
+            amountSpecified: amountSpecified,
+            sqrtPriceLimitX96: SQRT_RATIO_1_2
+        });
+        PoolSwapTest.TestSettings memory testSettings =
+            PoolSwapTest.TestSettings({takeClaims: false, settleUsingBurn: false});
+
+        snapStart("swap with dynamic fee and protocol fee");
```
hensha256, line 315: this snapshot should be in a different test because itll just be writing every fuzz run and leaving us with the final run which might change. better to do a gas test of some "normal" values to do the snapshot and then do fuzz without gas snaps

dianakocsis: oh okay nice, I just removed it since we already have "swap with lp fee and protocol fee"

src/libraries/Pool.sol
dianakocsis: thanks @shuhuiluo did the refactor to decrease bytecode size

test/libraries/ProtocolFeeLibrary.t.sol
```diff
@@ -0,0 +1,73 @@
...
+pragma solidity ^0.8.20;
+
+import "src/libraries/ProtocolFeeLibrary.sol";
+import "src/libraries/LPFeeLibrary.sol";
+import "forge-std/Test.sol";
+
+contract ProtocolFeeLibraryTest is Test {
+    function test_zeroForOne() public {
+        uint24 fee = uint24(ProtocolFeeLibrary.MAX_PROTOCOL_FEE) << 12 | ProtocolFeeLibrary.MAX_PROTOCOL_FEE;
+        assertEq(ProtocolFeeLibrary.getZeroForOneFee(fee), uint24(ProtocolFeeLibrary.MAX_PROTOCOL_FEE));
```
hensha256, line 11: for this test and the one below, could you make the 2 values different? because from this test we cannot check whether its getting the correct one or not
like
```
uint24(ProtocolFeeLibrary.MAX_PROTOCOL_FEE - 1) << 12 | ProtocolFeeLibrary.MAX_PROTOCOL_FEE;
```

then you know if its getting left or right number

hensha256 (approved):

test/DynamicFees.t.sol
```diff
@@ -146,17 +148,184 @@ contract TestDynamicFees is Test, Deployers, GasSnapshot {
...
+
+        vm.expectEmit(true, true, true, true, address(manager));
+        emit Swap(key.toId(), address(swapRouter), -100, 98, 79228162514264329749955861424, 1e18, -1, 1122);
+
+        swapRouter.swap(key, params, testSettings, ZERO_BYTES);
+
+        assertEq(_fetchPoolLPFee(key), 123);
+    }
+
+    function test_fuzz_ProtocolAndLPFee(uint24 lpFee, uint16 protocolFee0, uint16 protocolFee1, int256 amountSpecified)
```
snreynolds, line 291: nice fuzz

snreynolds (approved):

src/libraries/Pool.sol
shuhuiluo: I think caching `params.sqrtPriceLimitX96` saves more.

PR #548, opened 2024-04-02
add reverting tests, fix fuzz for hooks.t.sol

snreynolds:
## Related Issue

## Description of changes


test/libraries/Hooks.t.sol
```diff
@@ -194,7 +199,7 @@ contract HooksTest is Test, Deployers, GasSnapshot {
     }
 
     // hook validation
-    function test_ValidateHookAddress_noHooks(uint160 addr) public {
+    function test_validateHookAddress_noHooks(uint160 addr) public {
```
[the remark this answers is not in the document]
hensha256: quite a few in this file missing it

test/libraries/Hooks.t.sol
```diff
@@ -741,10 +746,13 @@ contract HooksTest is Test, Deployers, GasSnapshot {
...
     }
 
-    function test_validateHookAddress_failsAllHooks(uint152 addr, uint8 mask) public {
-        uint160 preAddr = uint160(uint256(addr));
-        vm.assume(mask != 0xff8);
-        IHooks hookAddr = IHooks(address(uint160(preAddr) | (uint160(mask) << 151)));
+    function test_validateHookAddress_failsAllHooks(uint160 addr, uint16 mask) public {
+        uint160 preAddr = uint160(uint256(addr) & clearAllHookPermisssionsMask);
+        // Set the upper `hooksPermissionCount` number of bits to get the full mask in uint16.
+        uint16 allHooksMask = (~uint16(0) << uint16(16 - hookPermissionCount));
```
hensha256, line 752: This isnt really test-specified so no need to execute it every time the test runs. Feels like it could sit globally with `clearAllHookPermisssionsMask`?

hensha256: actually on that note why is `clearAllHookPermisssionsMask` in the setup function - can't these both just be immutables that are calculated once in the contract and never again?

test/libraries/Hooks.t.sol
```diff
@@ -741,10 +746,13 @@ contract HooksTest is Test, Deployers, GasSnapshot {
         assertTrue(hookAddr.hasPermission(Hooks.AFTER_DONATE_FLAG));
     }
 
-    function test_validateHookAddress_failsAllHooks(uint152 addr, uint8 mask) public {
-        uint160 preAddr = uint160(uint256(addr));
-        vm.assume(mask != 0xff8);
-        IHooks hookAddr = IHooks(address(uint160(preAddr) | (uint160(mask) << 151)));
+    function test_validateHookAddress_failsAllHooks(uint160 addr, uint16 mask) public {
+        uint160 preAddr = uint160(uint256(addr) & clearAllHookPermisssionsMask);
```
hensha256, line 750: feels weird to go 160->256->160. Can we keep this always in a 256?

test/libraries/Hooks.t.sol
```diff
@@ -764,9 +772,9 @@ contract HooksTest is Test, Deployers, GasSnapshot {
     }
 
     function test_validateHookAddress_failsNoHooks(uint160 addr, uint16 mask) public {
-        uint160 preAddr = addr & uint160(0x007ffffFfffffffffFffffFFfFFFFFFffFFfFFff);
-        mask = mask & 0xff80; // the last 7 bits are all 0, we just want a 9 bit mask
-        vm.assume(mask != 0); // we want any combination except no hooks
+        uint160 preAddr = addr & (~uint160(0) >> hookPermissionCount);
```
hensha256, line 775: This can be `uint160(uint256(addr) & clearAllHookPermisssionsMask)` again

hensha256 (changes requested):
Just a few cleanup things but great new tests

test/libraries/Hooks.t.sol
snreynolds: do you mean keep always in a uint160?  I think we want to work with addresses (uint160)

test/libraries/Hooks.t.sol
snreynolds: yeah sure, its just calculated `clearAllHookPermisssionsMask` based on numHookPermissions. I added numHookPermissions so the tests would be super easy to update as we added more permissions. I'd like to keep it that way. But this var and the allHooksMask can be set at the top level

snreynolds: To close the loop here `hookPermissionCount` and `clearAllHookPermisssionsMask` are all declared in Deployers.sol (fixed in sep PR). We only really use allHooksMask here once in this file and no where else so I left it.

test/libraries/Hooks.t.sol
snreynolds: changed so that clearAllHookPermisssionsMask is uint160

test/libraries/Hooks.t.sol
```diff
@@ -881,11 +886,13 @@ contract HooksTest is Test, Deployers, GasSnapshot {
...
-        uint160 preAddr = uint160(uint256(addr));
-        mask = mask & 0xfffc; // the last 7 bits are all 0, we just want a 14 bit mask
-        vm.assume(mask != 0xfffc); // we want any combination except all hooks
-        IHooks hookAddr = IHooks(address(uint160(preAddr) | (uint160(mask) << 151)));
+    function test_fuzz_validateHookAddress_failsAllHooks(uint160 addr, uint16 mask) public {
+        uint160 preAddr = addr & clearAllHookPermisssionsMask;
+        // Set the upper `hooksPermissionCount` number of bits to get the full mask in uint16.
+        uint16 allHooksMask = (~uint16(0) << uint16(16 - hookPermissionCount));
+        // We want any combination except all hooks.
+        vm.assume(mask < allHooksMask);
```
hensha256, line 894: the name of the test is `failsAllHooks` as if it has all hooks - but this test explicitly _doesnt_ have all hooks?

hensha256: OH i see its that `validateHookPermissions` has all hooks

hensha256 (dismissed):

dianakocsis:
can you make all the import paths in Hooks.t.sol relative?

dianakocsis:
i forget, did we want to change the hook flags to be uint160 instead of uint256?

dianakocsis (approved):

Issue #553, opened 2024-04-03
Move hook bit flags to LSBs


Philogy:
### Component

Hooks

### Describe the suggested feature and problem it solves.

Instead of using the most significant bits of the hook address to configure the different hook flags I suggest using the lower, least significant bits. This will allow people to mine and use classical vanity addresses with leading zero bytes for their hooks without worrying about the impact that might have on the hook's permissions.

nishim3:
It will cause problems in using ```vm.etch```. It does not deal well with address values in a range of [1-10](https://github.com/foundry-rs/foundry/issues/1159).

Philogy:
> It will cause problems in using ```vm.etch```. It does not deal well with address values in a range of [1-10](https://github.com/foundry-rs/foundry/issues/1159).
> 
> 

What does that have to do with my proposal? I was only suggesting the lower bits are interpreted as the hook flags to allow for addresses with some leading zeros not addresses in the range 1-10.

nishim3:
Sorry for the misunderstanding! I have opened a PR with what I think is the desired implementation. Do have a look.

PR #558, opened 2024-04-04
Change Hook Flags to LSB

nishim3:
## Related Issue
#553 
## Description of changes


test/SkipCallsTestHook.t.sol
```diff
@@ -28,11 +28,11 @@ contract SkipCallsTest is Test, Deployers, GasSnapshot {
     PoolSwapTest.TestSettings testSettings =
         PoolSwapTest.TestSettings({withdrawTokens: true, settleUsingTransfer: true, currencyAlreadySent: false});
 
-    uint160 clearAllHookPermisssionsMask;
+    uint256 clearAllHookPermisssionsMask;
     uint256 hookPermissionCount = 10;
 
     function setUp() public {
-        clearAllHookPermisssionsMask = ~uint160(0) >> hookPermissionCount;
+        clearAllHookPermisssionsMask = uint256(1 >> (hookPermissionCount));
```
Philogy, line 35: This seems wrong, I think you want `~((1 << hookPermissionCount) - 1)`

test/SkipCallsTestHook.t.sol
```diff
@@ -55,7 +55,9 @@ contract SkipCallsTest is Test, Deployers, GasSnapshot {
 
     function test_beforeInitialize_skipIfCalledByHook() public {
         SkipCallsTestHook skipCallsTestHook = SkipCallsTestHook(
-            address(uint160(type(uint160).max & clearAllHookPermisssionsMask | Hooks.BEFORE_INITIALIZE_FLAG))
+            address(
+                uint160(type(uint160).max & clearAllHookPermisssionsMask | Hooks.BEFORE_INITIALIZE_FLAG) + 100000000000
```
Philogy, line 59: If the mask actually cleared the hook flags the addition would result in an overflow, remove.

Philogy (changes requested):
The clearing mask is calculated wrong. Fixing that will make the additions wrong so remove those too.

test/libraries/Hooks.t.sol
```diff
@@ -25,16 +25,16 @@ contract HooksTest is Test, Deployers, GasSnapshot {
...
+    address payable ALL_HOOKS_ADDRESS = payable(0x00000000000000000000000000000000000003fF);
     MockHooks mockHooks;
 
     // Update this value when you add a new hook flag. And then update all appropriate asserts.
     uint256 hookPermissionCount = 10;
     uint256 clearAllHookPermisssionsMask;
 
     function setUp() public {
-        clearAllHookPermisssionsMask = uint256(~uint160(0) >> (hookPermissionCount));
+        clearAllHookPermisssionsMask = uint256(1 >> (hookPermissionCount));
```
Philogy, line 37: This is incorrect, it should be `uint256(~uint160(0) << hookPermissionCount)`

test/libraries/Hooks.t.sol
```diff
@@ -712,7 +712,7 @@ contract HooksTest is Test, Deployers, GasSnapshot {
 
     function test_validateHookAddress_allHooks(uint160 addr) public {
         uint160 preAddr = uint160(uint256(addr) & clearAllHookPermisssionsMask);
-        uint160 allHookBitsFlipped = (~uint160(0)) << uint160((160 - hookPermissionCount));
+        uint160 allHookBitsFlipped = (~uint160(0));
```
Philogy, line 715: This is all bits flipped, it should be `((1 << hookPermissionCount) - 1)`

test/libraries/Hooks.t.sol
```diff
@@ -765,9 +765,9 @@ contract HooksTest is Test, Deployers, GasSnapshot {
 
     function test_validateHookAddress_failsNoHooks(uint160 addr, uint16 mask) public {
         uint160 preAddr = addr & uint160(0x007ffffFfffffffffFffffFFfFFFFFFffFFfFFff);
```
Philogy, line 767: You need to adjust this mask so that the lower bits are cleared.

Philogy (commented):
Found more stuff

snreynolds:
Hey couple thoughts here - if we are actually going to change how hook address flags are set to enable leading 0s/vanity addresses we have been brainstorming a different way, that actually just removed the flags from the address entirely. 

In this design, a hook address is any address with at least `MAX_NUM_FLAG_PERMISSIONS` of leading zeros. When we initialize a new pool with hooks and flags we just mask the flag bits in the pool key for the hook, and save the state according to the masked address. To call hook from pool key, we just zero out the leading flags. 

This is better for a few reasons than current design because we'd be forcing leading zeros, and it would allow one hook contract to be associated with multiple pools and different permissions on each pool. 

If we're going to change the design for hook flags, I think I'm definitely in favor of this design. Let me know thoughts cc @Philogy @nishim3

nishim3:
@snreynolds I am ok with this implementation.

akshatmittal:
@snreynolds If it matters at all, I very much like the approach of removing the bits from the address entirely and instead enforcing a minimum number of leading 0 bytes. Makes it much cleaner and allows for more optimization where necessary.

nishim3:
Hey @snreynolds! Any update on the design?

snreynolds:
> Hey @snreynolds! Any update on the design?

Hey! I think we'd love to go forward with the original suggested design - moving the permissions to the lower bits. If you could update this PR to the base branch @hensha256 can give a review. Thanks!

hensha256:
Hey @nishim3! Just pinging here to get main updated. Otherwise i will do so myself later!

nishim3:
The tests are passing locally. Some issue with the lib folder is causing tests to fail here...Help appreciated

.forge-snapshots/update dynamic fee in before swap.snap
```diff
@@ -1 +1 @@
-157503
\ No newline at end of file
+124589
```
hensha256, line 1: please can you run the tests with `forge test --isolate` and push the snapshots? Looks to me like you ran them with `forge test`

src/libraries/Hooks.sol
```diff
@@ -117,7 +117,7 @@ library Hooks {
         // If a hook contract is set, it must have at least 1 flag set, or have a dynamic fee
         return address(self) == address(0)
             ? !fee.isDynamicFee()
-            : (uint160(address(self)) >= AFTER_REMOVE_LIQUIDITY_RETURNS_DELTA_FLAG || fee.isDynamicFee());
+            : (uint160(address(self)) & uint160((1 << 14) - 1) > 0 || fee.isDynamicFee());
```
hensha256, line 120: might be nice to make a new constant `ALL_HOOK_MASK = uint160((1 << 14) - 1))` or something to put here

test/DynamicFees.t.sol
```diff
@@ -28,14 +28,14 @@ contract TestDynamicFees is Test, Deployers, GasSnapshot {
     DynamicFeesTestHook dynamicFeesHooks = DynamicFeesTestHook(
         address(
             uint160(
-                uint256(type(uint160).max) & clearAllHookPermisssionsMask | Hooks.BEFORE_SWAP_FLAG
+                uint256(type(uint160).max) & clearAllHookPermissionsMask | Hooks.BEFORE_SWAP_FLAG
```
hensha256, line 31: nice 😁

test/libraries/Hooks.t.sol
```diff
@@ -30,7 +30,7 @@ contract HooksTest is Test, Deployers, GasSnapshot {
     using StateLibrary for IPoolManager;
 
     /// 1111 1111 1111 1100
```
hensha256, line 32: comments needs updating :)

test/libraries/Hooks.t.sol
```diff
@@ -986,20 +986,20 @@ contract HooksTest is Test, Deployers, GasSnapshot {
...
+            Hooks.isValidHookAddress(IHooks(0x1000000000000000000000000000000000000000), LPFeeLibrary.DYNAMIC_FEE_FLAG)
         );
         assertTrue(
             Hooks.isValidHookAddress(
-                IHooks(0x0000000000000000000000000000000000000001), LPFeeLibrary.DYNAMIC_FEE_FLAG | uint24(3000)
+                IHooks(0x1000000000000000000000000000000000000000), LPFeeLibrary.DYNAMIC_FEE_FLAG | uint24(3000)
             )
         );
-        assertTrue(Hooks.isValidHookAddress(IHooks(0x8000000000000000000000000000000000000000), 3000));
+        assertTrue(Hooks.isValidHookAddress(IHooks(0x000000000A000000000000000000000000002000), 3000));
```
hensha256, line 996: i think this should be 4000. Youve translated it equivalently but i think reading the test name it should have 0 flags. So should end with 4000 not 2000

test/utils/Deployers.sol
```diff
@@ -80,8 +80,8 @@ contract Deployers {
     PoolKey uninitializedNativeKey;
 
     // Update this value when you add a new hook flag.
-    uint256 hookPermissionCount = 14;
-    uint160 clearAllHookPermisssionsMask = ~uint160(0) >> (hookPermissionCount);
+    uint160 hookPermissionCount = 14;
+    uint160 clearAllHookPermissionsMask = uint160(~((1 << hookPermissionCount) - 1));
```
hensha256, line 84: what youve got here is all 0s - it doesnt just clear the hook permissions it clears everything. I think instead it would be better as `~uint160(0) << hookPermissionCount` to have all 1s except the 14 lowest bits

hensha256 (changes requested):
Few more thoughts! We would _love_ to merge this today if youre able to do these things 🚀

test/libraries/Hooks.t.sol
hensha256: Ohhh yes sorry i didnt see this has no dynamic fee. The test clearly says its to test hooks _with_ a dynamic fee, and we already have a test for noFlagsNoDynamicFee. So I think just remove this line.

THank you!

src/libraries/Hooks.sol
```diff
@@ -12,7 +12,7 @@ import {ParseBytes} from "../libraries/ParseBytes.sol";
 
 /// @notice V4 decides whether to invoke specific hooks by inspecting the leading bits of the address that
```
hensha256: `leading` -> least significant / final

src/libraries/Hooks.sol
```diff
@@ -12,7 +12,7 @@ import {ParseBytes} from "../libraries/ParseBytes.sol";
...
 /// the hooks contract is deployed to.
-/// For example, a hooks contract deployed to address: 0x9000000000000000000000000000000000000000
+/// For example, a hooks contract deployed to address: 0x0000000000000000000000000000000000002400
 /// has leading bits '1001' which would cause the 'before initialize' and 'after add liquidity' hooks to be used.
```
hensha256: same here

src/libraries/Hooks.sol
hensha256: oh you already did this sorry

.forge-snapshots/addLiquidity with empty hook.snap
```diff
@@ -1 +1 @@
-285671
\ No newline at end of file
+266944
```
hensha256, line 1: hm this shouldnt be so big 🤔

test/utils/Constants.sol
```diff
@@ -14,8 +14,8 @@ library Constants {
     uint160 constant MAX_UINT160 = type(uint160).max;
 
     address constant ADDRESS_ZERO = address(0);
-    address constant EMPTY_HOOKS = address(0xf000000000000000000000000000000000000000);
-    address constant MOCK_HOOKS = address(0xfF00000000000000000000000000000000000000);
+    address constant EMPTY_HOOKS = address(0x000000000000000000000000000000000000f000);
```
hensha256, line 17: ah this should be `3c00` not `f000` to be equivalent.
side note the name `EMPTY_HOOKS` doesnt make sense but thts a pre-existing issue that we will fix sometime soon

hensha256 (approved):

PR #559, opened 2024-04-05
optimize extsload

0xalpharush:
## Related Issue
Replaces https://github.com/Uniswap/v4-core/pull/255

## Description of changes
hoist loop invariant code, strength reduction (mul -> add, mul-by-power-of-2 -> shift)


src/PoolManager.sol
```diff
@@ -311,12 +311,18 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     }
 
     function extsload(bytes32 startSlot, uint256 nSlots) external view returns (bytes memory) {
-        bytes memory value = new bytes(32 * nSlots);
+        // For values of nSlots that cause an overflow, an out-of-gas exception will be thrown.
+        bytes memory value = new bytes(nSlots << 5);
```
shuhuiluo, line 315:
```Solidity
        assembly {
            let endSlot := add(startSlot, nSlots)
            let valuePtr
            for {} 1 {} {
                mstore(valuePtr, sload(startSlot))
                valuePtr := add(valuePtr, 0x20)
                startSlot := add(startSlot, 1)
                if iszero(lt(startSlot, endSlot)) { break }
            }
            // For values of nSlots that cause an overflow, an out-of-gas exception will be thrown.
            return(0, shl(5, nSlots))
        }
```
Because `extsload` is an external function that enters a new call context and exits right after execution, we can disregard Solidity's memory management convention and directly return a slice of memory.

shuhuiluo (changes requested):
@hensha256 The current function signature is not as flexible as the following.
```Solidity
    function extsload(bytes32[] calldata slots) external view returns (bytes32[] memory) {
        assembly ("memory-safe") {
            // abi offset for dynamic array
            mstore(0, 0x20)
            mstore(0x20, slots.length)
            let end := add(0x40, shl(5, slots.length))
            let memptr := 0x40
            let calldataptr := slots.offset
            for { } 1 { } {
                mstore(memptr, sload(calldataload(calldataptr)))
                memptr := add(memptr, 0x20)
                calldataptr := add(calldataptr, 0x20)
                if iszero(lt(memptr, end)) { break }
            }
            return(0, end)
        }
    }
```

snreynolds:
Going to close #255 then since this replaces it.

PR #560, opened 2024-04-05
Optimize `TickBitmap`

shuhuiluo:
## Related Issue

## Description of changes

`TickBitmap` has been rewritten heavily using inline assembly because such a core library deserves the deepest level of optimizations. One may argue readability has taken a hit. But the original code wasn't that readable to start with given all the explicit type casting. The new implementation should be understandable with cleaner control flow and extensive comments.

A notable change involves inverting if-else statements as explained in [The RareSkills Book of Solidity Gas Optimization](https://www.rareskills.io/post/gas-optimization#viewer-fp51d).


shuhuiluo:
Broken apart to #653 and #654.

PR #482, opened 2024-02-15
Hooks return deltas

hensha256:
Closes #463

PR #562, opened 2024-04-05
beforeSwap returning a dynamic fee

### LINKED ISSUE #554: Return swapFee in beforeSwap for hooks with relevant permission
### Component

Hooks, Gas Optimization

```
    function beforeSwap(
        address sender,
        PoolKey calldata key,
        bytes calldata hookData
```

### Describe alternatives.

### Additional context.

_No response_

saucepoint:
## Related Issue
Closes #554, but **is blocked by hooks-return-delta** as it does most of the work around reading return values from beforeSwap

- [ ] merge hooks-return-delta
- [x] increase test coverage & fuzz
- [ ] define the sentinel _somewhere_ (SwapFeeLibrary?)

## Description of changes

---


snreynolds:
blocked by #482

PR #564, opened 2024-04-05
Move msg.sender check to modifier

hensha256:
## Related Issue
When merging main into #482 it got a bit messy - this modifier will be cleaner for that PR
Also I added a gas test - this change saves 750 gas on that case 👯


.forge-snapshots/swap skips hook call is hook is caller.snap
```diff
@@ -0,0 +1 @@
+156294
```
snreynolds, line 1: You tested this gas against the old impl?

snreynolds (commented):
yeah the repetition is less nice, but if it saves gas i think its p clean

.forge-snapshots/swap skips hook call is hook is caller.snap
hensha256: oops i didnt reply
on main: 157001

snreynolds (approved):

snreynolds (approved):

PR #568, opened 2024-04-06
Use `addDelta` for adding liquidity deltas in `Pool`

shuhuiluo:
## Related Issue

Bad gas.

## Description of changes
A new function `toUint128` has been introduced in the `SafeCast` library. Simplified the logic for adding and subtracting liquidity deltas in `Pool.sol` using the `addDelta` function. Made a corresponding change in the `Pool.t.sol` test file to expect a `SafeCastOverflow` error instead of a `Panic` for a negative `liquidityDelta`.


src/libraries/Position.sol
```diff
@@ -54,9 +58,7 @@ library Position {
             if (_self.liquidity == 0) revert CannotUpdateEmptyPosition(); // disallow pokes for 0 liquidity positions
             liquidityNext = _self.liquidity;
         } else {
-            liquidityNext = liquidityDelta < 0
-                ? _self.liquidity - uint128(-liquidityDelta)
-                : _self.liquidity + uint128(liquidityDelta);
+            liquidityNext = Pool.addDelta(_self.liquidity, liquidityDelta).toUint128();
```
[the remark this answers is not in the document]
shuhuiluo: I actually intended to bring back [`LiquidityMath`](https://github.com/Uniswap/v3-core/blob/main/contracts/libraries/LiquidityMath.sol) but wanted approval from you first.

src/libraries/Pool.sol
```diff
@@ -603,4 +597,15 @@ library Pool {
             liquidityNet = info.liquidityNet;
         }
     }
+
+    /// @notice Add a signed liquidity delta to liquidity
+    /// @dev Must call toUint128 on the result to revert if it overflows or underflows. This is to allow optimizer inlining.
+    /// @param x The liquidity before change
+    /// @param y The delta by which liquidity should be changed
+    /// @return z The liquidity after the change, declared as uint256 to prevent implicit conversion when calling toUint128
+    function addDelta(uint128 x, int128 y) internal pure returns (uint256 z) {
```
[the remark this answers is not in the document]
shuhuiluo: I see your point. Using `toUint128` within `addDelta` would increase gas on `addLiquidity` and `removeLiquidity` snapshots by 187. Because it disallows the optimizer to inline the function when it may revert.

> To save gas, it can inline short functions that do not contain control-flow branches or opcodes with side-effects.

Reference:
https://soliditylang.org/blog/2021/03/02/saving-gas-with-simple-inliner/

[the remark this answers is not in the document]
shuhuiluo: I decided to inline `toUint128` within `addDelta` which ended up saving more gas. Thanks for the prompt. :grinning:

src/libraries/SafeCast.sol
```diff
@@ -6,6 +6,14 @@ pragma solidity ^0.8.20;
 library SafeCast {
     error SafeCastOverflow();
 
+    /// @notice Cast a uint256 to a uint128, revert on overflow
+    /// @param y The uint256 to be downcasted
+    /// @return z The downcasted integer, now type uint128
+    function toUint128(uint256 y) internal pure returns (uint128 z) {
```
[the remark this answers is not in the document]
shuhuiluo: Now that `toUint128` is inlined within `addDelta`, I plan to revert the changes in `SafeCast` and do it in a separate PR.

hensha256:
Ok yeah this looks great. 2 things
- please can you add a fuzz test in a new LiquidityMath.t.sol
- i think i would prefer `using LiquidityMath for uint128` in those files and then `liquidity.addDelta(delta)`

shuhuiluo:
> Ok yeah this looks great. 2 things
> 
> * please can you add a fuzz test in a new LiquidityMath.t.sol
> * i think i would prefer `using LiquidityMath for uint128` in those files and then `liquidity.addDelta(delta)`

@hensha256 It is probably a compiler bug because the syntactic sugar increases gas by 12. See https://github.com/Uniswap/v4-core/pull/568/commits/466033ede4ca1b42b04004397e22969ab32d120f.

test/libraries/LiquidityMath.t.sol
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /// @notice Test the equivalence of the new `addDelta` and the reference implementation
+    function test_addDelta_fuzz(uint128 x, int128 y) public {
+        vm.assume(y != type(int128).min);
+        try liquidityMath.addDelta(x, y) returns (uint128 z) {
+            assertEq(z, liquidityMathRef.addDelta(x, y));
+        } catch (bytes memory reason) {
+            assertEq(bytes4(reason), SafeCast.SafeCastOverflow.selector);
+            vm.expectRevert();
```
hensha256, line 55: is the error `vm.expectRevert(abi.encodeWithSignature("Panic(uint256)", 0x11));` ?
if so can we add the reason to all the `liquidityMathRef reverts in this file pls?

src/test/LiquidityMathTest.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+import {LiquidityMath} from "../libraries/LiquidityMath.sol";
+
+contract LiquidityMathTest {
```
hensha256, line 6: I feel like we dont need this file right? Can't `LiquidityMath.t.sol` just call the library directly?
I know some of our other test files are like that too, I want to clean them up

test/libraries/LiquidityMath.t.sol
```diff
@@ -0,0 +1,59 @@
...
+        liquidityMath.addDelta(uint128(type(int128).max), type(int128).min);
+    }
+
+    /// @notice Test the revert reason for overflow
+    function test_addDelta_throwsForOverflow() public {
+        vm.expectRevert(SafeCast.SafeCastOverflow.selector);
+        liquidityMath.addDelta(type(uint128).max, 1);
+    }
+
+    function test_addDelta_sub_int128min_throwsForReferenceOnly() public {
```
hensha256, line 37: oh woah nice edgecase

hensha256 (commented):
Looking great! Just a couple of things.

test/libraries/LiquidityMath.t.sol
shuhuiluo: Yes.

src/test/LiquidityMathTest.sol
shuhuiluo: I can move this to `LiquidityMath.t.sol`. But it's required to test positive and negative cases using [try/catch](https://docs.soliditylang.org/en/latest/control-structures.html#try-catch).

> The try keyword has to be followed by an expression representing an external function call or a contract creation (new ContractName()).

hensha256: ohhh very interesting, i like this being a contract in its own file then if its needed for try catch! Sorry for the back and forth 😅

hensha256 (approved):

PR #576, opened 2024-04-09
Optimize `SafeCast`

shuhuiluo:
## Related Issue

## Description of changes
In the `SafeCast` library, a new function `toUint128` has been added to support casting uint256 to uint128. Accompanying the update, a new fuzz test for `toUint128` is also added to verify the behavior when `toUint128` encounters a potential overflow condition.

Refactor `SafeCast.sol` library to produce optimal bytecode while keeping inline assembly to a minimal. As a result, the code optimizes casting operations which result in reduced gas usage. These changes affect numerous `.forge-snapshot` files, demonstrating an overall reduction in gas usage across multiple contracts.

@Vectorized Thanks for the inspiration in https://github.com/Vectorized/solady/pull/928 and https://github.com/Vectorized/solady/pull/929.


src/libraries/SafeCast.sol
```diff
@@ -10,31 +10,74 @@ library SafeCast {
     /// @param y The uint256 to be downcasted
     /// @return z The downcasted integer, now type uint160
     function toUint160(uint256 y) internal pure returns (uint160 z) {
-        z = uint160(y);
-        if (z != y) revert SafeCastOverflow();
+        /// @solidity memory-safe-assembly
+        assembly {
```
hensha256, line 14: couple of questions...
1. is it cheaper to remove the creation of `z` - i.e. `returns (uint160)`, and then just add `return y`?

shuhuiluo: The following produces the same gas.
```solidity
    function toUint160(uint256 y) internal pure returns (uint160) {
        assembly {
            if iszero(lt(y, shl(160, 1))) {
                // revert SafeCastOverflow();
                mstore(0, 0x93dafdf1)
                revert(0x1c, 0x04)
            }
        }
        return uint160(y);
    }
```

shuhuiluo: Bumping this up since readability improved a lot.

src/libraries/SafeCast.sol
```diff
@@ -6,35 +6,55 @@ pragma solidity ^0.8.20;
...
     /// @notice Cast a uint256 to a uint160, revert on overflow
-    /// @param y The uint256 to be downcasted
-    /// @return z The downcasted integer, now type uint160
-    function toUint160(uint256 y) internal pure returns (uint160 z) {
-        z = uint160(y);
-        if (z != y) revert SafeCastOverflow();
+    /// @param x The uint256 to be downcasted
+    /// @return The downcasted integer, now type uint160
+    function toUint160(uint256 x) internal pure returns (uint160) {
+        if (x >= 1 << 160) _revertOverflow();
```
snreynolds, line 23: technically 1<<160 and 1 << 128 can be constants which may make this even cheaper?

shuhuiluo: The compiler would compute it to a constant at compile time. It increases readability imo.

snreynolds (approved):
lgtm

snreynolds (approved):

PR #577, opened 2024-04-09
change protocol fee push to be via protocol fee controller

dianakocsis:
## Related Issue
Change protocol fee push to be via protocol fee controller

## Description of changes
`setProtocolFee` can only be called by fee controller


test/PoolManager.t.sol
```diff
@@ -948,42 +948,40 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
...
 
-        manager.setProtocolFeeController(revertingFeeController);
-        vm.expectRevert(IProtocolFees.ProtocolFeeControllerCallFailedOrInvalidResult.selector);
-        manager.setProtocolFee(key);
-
         manager.setProtocolFeeController(outOfBoundsFeeController);
-        vm.expectRevert(IProtocolFees.ProtocolFeeControllerCallFailedOrInvalidResult.selector);
-        manager.setProtocolFee(key);
+        uint24 newProtocolFee = outOfBoundsFeeController.protocolFeeForPool(key);
+        vm.prank(address(outOfBoundsFeeController));
```
hensha256, line 974: i think i'd rather we add a function to the outOfBoundsFeeController contract that calls the PoolManager itself. And then here you'd do like
```
vm.expectRevert(IProtocolFees.InvalidProtocolFee.selector);
outOfBoundsFeeController.setProtocolFee(key);
```

and the contract would then call 
```
manager.setProtocolFee(key, newProtocolFee)
```

instead of pranking the contract address

test/PoolManager.t.sol
```diff
@@ -1002,8 +1000,9 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
     function test_collectProtocolFees_ERC20_accumulateFees_gas() public {
         uint256 expectedFees = 7;
 
-        feeController.setProtocolFeeForPool(key.toId(), MAX_FEE_BOTH_TOKENS);
-        manager.setProtocolFee(key);
+        uint24 protocolFee = MAX_FEE_BOTH_TOKENS;
+        vm.prank(address(feeController));
```
hensha256, line 1004: and same with all the tests for feeController - better to code the feeController to call the manager

.forge-snapshots/simple swap.snap
```diff
@@ -1 +1 @@
-146476
\ No newline at end of file
+146564
```
hensha256, line 1: this makes no sense 😭

snreynolds (approved):
lgtm!

hensha256 (dismissed):

snreynolds (approved):

PR #578, opened 2024-04-10
Currency settle take library

### LINKED ISSUE #572: Add new currency lib for settle, update PoolTestBase, & ensure all contracts use this pattern
### Component

General design optimization (improving efficiency, cleanliness, or developer experience)

Most testing routers use `PoolTestBase.sol` under the hood to take/settle/transfer. I think we should try to use a standard method for this in our tests by:

1. updating PoolTestBase to use something like this library https://github.com/saucepoint/v4-periphery/blob/position-manager/contracts/libraries/CurrencySettleTake.sol
2. updating PoolTestBase to also allow for mint/burn/other ways of settling. This will be helpful for new kinds of testing we will be adding to core in the near future.
3. update any tests that settle without relying on PoolTestBase... (ie `SkipCallsTestHook.sol`)

_No response_

### Describe alternatives.

_No response_

### Additional context.

_No response_

saucepoint:
## Related Issue
Closes #572 

## Description of changes

Introduce a library for handling *currency* transfers

* The library handles the three token types: native, ERC20, and ERC6909 so that calls for settle/take/mint/burn do not need to be aware of currency type
* Enables clean syntactic sugar
* Bonus: saves gas!

---

**Notes**:
* open to calling it something other than `CurrencySettleTake`
* we could add specific calls for `.mint` and `.burn` but I think the purpose of the library is to uniformly handle token transfers. However, syntax such as `key.currency0.mint(...)` is much more obvious than `key.currency0.take(..., true)`


src/libraries/CurrencySettleTake.sol
```diff
@@ -0,0 +1,40 @@
...
+
+import {Currency, CurrencyLibrary} from "../types/Currency.sol";
+import {IERC20Minimal} from "../interfaces/external/IERC20Minimal.sol";
+import {IPoolManager} from "../interfaces/IPoolManager.sol";
+
+library CurrencySettleTake {
+    using CurrencyLibrary for Currency;
+
+    /// @notice Settle (pay) a currency to the PoolManager
+    /// @dev Burning a native 6909 token takes precedence over a native currency transfer
```
snreynolds, line 12: is there a particular reason burn takes precedent? idm either way was just curious

src/test/PoolClaimsTest.sol
```diff
@@ -39,10 +41,10 @@ contract PoolClaimsTest is PoolTestBase {
 
         if (data.deposit) {
             manager.mint(data.user, data.currency.toId(), uint128(data.amount));
```
snreynolds, line 43: this mint call and the burn call on L46 can now use the lib

src/test/PoolModifyLiquidityTest.sol
```diff
@@ -72,10 +74,10 @@ contract PoolModifyLiquidityTest is PoolTestBase {
             assert(!(delta0 > 0 || delta1 > 0));
         }
 
-        if (delta0 < 0) _settle(data.key.currency0, data.sender, int128(delta0), data.settleUsingTransfer);
-        if (delta1 < 0) _settle(data.key.currency1, data.sender, int128(delta1), data.settleUsingTransfer);
-        if (delta0 > 0) _take(data.key.currency0, data.sender, int128(delta0), data.withdrawTokens);
-        if (delta1 > 0) _take(data.key.currency1, data.sender, int128(delta1), data.withdrawTokens);
+        if (delta0 < 0) data.key.currency0.settle(manager, data.sender, uint256(-delta0), !data.settleUsingTransfer);
```
snreynolds, line 77: I'd just change the testSettings params to match the underlying burn/claims bool... ie 

```
    struct TestSettings {
        bool takeClaims; // instead of withdrawTokens
        bool settleWithBurn; // instead of settleUsingTransfer
        bool currencyAlreadySent;
    }
```

that way here you can just pass in data.settleWithBurn/data.takeClaims

src/test/PoolDonateTest.sol
```diff
@@ -57,10 +59,10 @@ contract PoolDonateTest is PoolTestBase {
         require(deltaAfter0 == -int256(data.amount0), "deltaAfter0 is not equal to -int256(data.amount0)");
         require(deltaAfter1 == -int256(data.amount1), "deltaAfter1 is not equal to -int256(data.amount1)");
 
-        if (deltaAfter0 < 0) _settle(data.key.currency0, data.sender, int128(deltaAfter0), true);
-        if (deltaAfter1 < 0) _settle(data.key.currency1, data.sender, int128(deltaAfter1), true);
-        if (deltaAfter0 > 0) _take(data.key.currency0, data.sender, int128(deltaAfter0), true);
-        if (deltaAfter1 > 0) _take(data.key.currency1, data.sender, int128(deltaAfter1), true);
+        if (deltaAfter0 < 0) data.key.currency0.settle(manager, data.sender, uint256(-deltaAfter0), false);
```
snreynolds, line 62: yay sweet - thanks for just changing to uint256. this was bugging me :P

snreynolds (changes requested):

snreynolds (changes requested):
nice looking good

src/libraries/CurrencySettleTake.sol
saucepoint: if native had priority, none of the 6909 claims would work for native currency. i.e. you wouldnt be able to burn a 6909-wrapped native eth, since the if-condition would short circuit on currency.isNative() and use native transfers

was caught in one of the tests

saucepoint: wonder if should update the code comment or just remove it entirely? maybe i change it to indicate that we want to support 6909-wrapped native currencies and we shouldnt short-circuit for native transfer?

saucepoint: moved to the comment to within the function

snreynolds: sweet looks good

snreynolds (approved):

snreynolds (approved):

PR #582, opened 2024-04-12
Refactor `Pool.swap` to cross multiple empty tick words at once

shuhuiluo:
## Related Issue

## Description of changes
This pull request is a continuation of #560.

Introduces a new function named `nextInitializedTick` in the `TickBitmap` library. This function returns the next initialized tick either to the left or right of a given tick, not limited to one word away. Corresponding tests have been added in the `TickBitmap.t.sol` file to verify the functionality.

Refactors the `Pool.swap` function and removes the limitations of transitioning to ticks within one word. This update simplifies the function by allowing transitions between any initialized ticks, which are not limited to the same word.

Gas snapshots reflect savings as large as `5000`. This is especially helpful for a large swap on a pool with thin liquidity and sparse tick map, which necessitates the crossing of multiple words in the tick bitmap. Such a test case may not be present yet to demonstrate the level of optimizations.

Changes to the `Pool.swap` function are minimal which gives confidence that the business logic would remain intact. Still, extra scrutiny and verification is warranted for such a core library.


hensha256:

PR #592, opened 2024-04-21
Refactor `Position.update`

shuhuiluo:
## Related Issue

## Description of changes
The `Position.update` method in the Position.sol library has been streamlined. The `liquidityNext` variable was removed and the `self.liquidity` is now updated directly inside the condition that checks if there are any changes in liquidity. Updating `self.liquidity` directly is more efficient, simplifies the logic, and reduces the bytecode size of the library.


hensha256 (dismissed):
nice

hensha256 (approved):

PR #595, opened 2024-04-23
Remove indexed modifier from select fields

codyborn:
## Description of changes

Saves ~100 gas on Swaps and Pool creation.

Removes the `indexed` modifier from the `sender` in the `Swap` event. This field is important for wallet providers but they will already have the transaction hash when sent by the user, which they can use to track the state change from this user.

Removes the `indexed` modifier from the `id` in the `Initialize` event. It doesn't make sense to index on this field since this is the first time that the PoolId will be created. Future events (`Swap`/`ModifyLiquidity`) can be filtered on this PoolId, but no one would be aware of the PoolId until the Initialize event is emitted.


hensha256:
thank you!!
2 things
- your commit needs to be signed for this repo! please can you remove your unsigned commit and replace it with a signed one?
- please can you run `forge fmt` 🙏

snreynolds (approved):

PR #596, opened 2024-04-23
Currency unit tests

hensha256:


snreynolds:
It looks like Currency.sol doesn't have full coverage - is that a mistake of the tool? You can run `forge coverage --report debug` to inspect what branches in the code are untested.

dianakocsis:
> forge coverage --report debug

I think it might be mistake of the tool. 
`Uncovered for src/types/Currency.sol:
- Statement (location: source ID 98, line 97, chars 3909-3932, hits: 0)`

It says this statement inside `isNative` isn't being covered, but it is
    function isNative(Currency currency) internal pure returns (bool) {
        return Currency.unwrap(currency) == Currency.unwrap(NATIVE);
    }

src/test/DynamicReturnFeeTestHook.sol
```diff
@@ -24,6 +24,7 @@ contract DynamicReturnFeeTestHook is BaseTestHooks {
 
     function beforeSwap(address, PoolKey calldata, IPoolManager.SwapParams calldata, bytes calldata)
         external
+        view
```
hensha256, line 27: compiler warning on main rn

dianakocsis (changes requested):

src/test/DynamicReturnFeeTestHook.sol
dianakocsis: nice noticed this too

test/types/Currency.t.sol
```diff
@@ -4,82 +4,133 @@ pragma solidity ^0.8.20;
...
+        assertEq(currencyTest.balanceOfSelf(erc20Currency), initialERC20Balance - amount);
+        assertEq(
+            currencyTest.balanceOfSelf(erc20Currency),
+            MockERC20(Currency.unwrap(erc20Currency)).balanceOf(address(currencyTest))
+        );
     }
 
-    function testCurrency_toId_native_returns0() public view {
-        assertEq(nativeCurrency.toId(), uint256(0));
+    function test_fuzz_balanceOf_native(uint256 amount) public {
```
dianakocsis, line 60: can you assert that the currencyTest balance is the prev balance minus amount here too?

dianakocsis: as well as `test_fuzz_balanceOf_token`, `test_fuzz_transfer_native`, and `test_fuzz_transfer_token`

dianakocsis (changes requested):

dianakocsis (approved):

PR #569, opened 2024-04-06
Optimize liquidity flipping calculations in `Pool`

shuhuiluo:
## Related Issue
Which issue does this pull request resolve?

## Description of changes
Optimize liquidity flipping calculation in `updateTick`. Add function `flipLiquidityDelta` to optimize liquidity delta operation. Optimize tick transition in `Pool.swap`.

PR #602, opened 2024-04-24
Optimize liquidity flipping calculations in `Pool`

shuhuiluo:
## Related Issue

## Description of changes
A new function `LiquidityMath.flipLiquidityDelta` is added to optimize the process of flipping the sign of a liquidity delta if certain conditions are met. Code complexity has been reduced by avoiding the need to use unchecked blocks and applied assembly for efficiency. The refactoring also resulted in changes in the snapshot files.


hensha256:
Replaces #569

src/libraries/Pool.sol
```diff
@@ -543,7 +544,15 @@ library Pool {
         }
 
         // when the lower (upper) tick is crossed left to right (right to left), liquidity must be added (removed)
-        int128 liquidityNet = upper ? liquidityNetBefore - liquidityDelta : liquidityNetBefore + liquidityDelta;
+        // Equivalent to `liquidityNet = upper ? liquidityNetBefore - liquidityDelta : liquidityNetBefore + liquidityDelta;`
+        // `int128 liquidityDelta` is passed from `modifyLiquidity` and should be sanitized in `PoolManager`
+        liquidityDelta = LiquidityMath.flipLiquidityDelta(liquidityDelta, upper);
```
hensha256, line 549: would love a comment here about why `liquidityDelta` cannot be min int128 to show the use of the new function is safe - thats already commented on the other use

src/libraries/LiquidityMath.sol
```diff
@@ -17,4 +19,19 @@ library LiquidityMath {
...
     }
+
+    /// @notice Flips the sign of a liquidity delta if a condition is true
+    /// @dev Equivalent to `res = flip ? -liquidityDelta : liquidityDelta;`
+    /// @dev `liquidityDelta` should never be `type(int128).min`
+    /// @param liquidityDelta The liquidity delta to potentially flip
+    /// @param flip Whether to flip the sign of the liquidity delta
+    /// @return res The potentially flipped liquidity delta
+    function flipLiquidityDelta(int128 liquidityDelta, bool flip) internal pure returns (int128 res) {
+        assembly {
```
hensha256, line 30: memory safe assembly comment?

src/libraries/Pool.sol
hensha256: oh ignore me thats what youre saying about the sanitization in PoolManager haha

src/libraries/LiquidityMath.sol
shuhuiluo: That wouldn't be necessary since the assembly block contains no memory opcodes.

src/libraries/Pool.sol
shuhuiluo: It's a little different. So `liquidityDelta` in `Pool.updateTick` is passed from `Pool.modifyLiquidity` and `PoolManager.modifyLiquidity`. And `PoolManager.modifyLiquidity` calls `toInt128()` on `params.liquidityDelta`. But after `toInt128()` a `int128` can still be `type(int128).min`.

shuhuiluo: First of all, the original ternary expressions revert on the negation of `type(int128).min`.

I am not sure where the comment "safe because liquidityNet cannot be type(int128).min" came from inside `Pool.swap`. I guess the max `liquidityNet` is limited by `tickSpacingToMaxLiquidityPerTick`.

If a `type(int128).min` is flipped in `flipLiquidityDelta`, instead of reverting, `1 << 127` will be returned which is a valid `int256` but not a valid `int128`.

There are two usages of `flipLiquidityDelta`.

- When `flipLiquidityDelta` is used on `liquidityNet` in `Pool.swap`, `liquidityNet` is immediately consumed by `addDelta` which treats it as an `int256` and only reverts if the resulting liquidity overflows `uint128`.
- When `flipLiquidityDelta` is used on `liquidityDelta` in `Pool.updateTick`, it is then used by `liquidityNet := add(liquidityNetBefore, liquidityDelta)` in the assembly block which also treats it as an `int256`. The validity of the sum is checked by `toInt128()`.

So even though the natspec of `flipLiquidityDelta` says "`liquidityDelta` should never be `type(int128).min`", its current usage is able to handle `type(int128).min`.

gretzke (approved):

PR #613, opened 2024-04-25
Use packed stack-based version of  Slot0 structure

0xVolosnikov:
This pull request does not relate to any issue. In my opinion, it is more convenient to discuss this change if you have an already written implementation in a pull request.

## Description of changes

The slot0 structure by design fits into one storage slot. However, when using the memory version of this structure, each field occupies a separate memory slot. In addition, the memory structure takes up more space on the solidity "stack".

The changes I propose allows to read this structure onto the stack and operate with the fields of this structure directly on the stack. As a result, the amount of memory used is reduced, which a little bit reduces gas costs and bytecode size.

Since the changes are more or less significant, I would first like to hear the opinion of the community and maintainers regarding the proposal to use a structure on the stack before adding tests and so on.


src/libraries/Pool.sol
```diff
@@ -105,9 +92,15 @@ library Pool {
...
-        if (tickUpper > TickMath.MAX_TICK) revert TickUpperOutOfBounds(tickUpper);
+        if (tickLower >= tickUpper) {
+            revert TicksMisordered(tickLower, tickUpper);
+        }
+        if (tickLower < TickMath.MIN_TICK) {
+            revert TickLowerOutOfBounds(tickLower);
+        }
+        if (tickUpper > TickMath.MAX_TICK) {
+            revert TickUpperOutOfBounds(tickUpper);
+        }
```
shuhuiluo, line 103: Unrelated code style changes.

src/libraries/Pool.sol
```diff
@@ -443,10 +441,12 @@ library Pool {
...
         }
 
-        (self.slot0.sqrtPriceX96, self.slot0.tick) = (state.sqrtPriceX96, state.tick);
+        self.slot0.storePacked(slot0Start.setTick(state.tick).setSqrtPriceX96(state.sqrtPriceX96));
 
         // update liquidity if it changed
-        if (cache.liquidityStart != state.liquidity) self.liquidity = state.liquidity;
+        if (cache.liquidityStart != state.liquidity) {
+            self.liquidity = state.liquidity;
+        }
```
shuhuiluo, line 449: Unrelated code style change.

src/types/Slot0.sol
```diff
@@ -0,0 +1,112 @@
...
+            _packed := sload(pointer.slot)
+        }
+    }
+}
+
+/// @dev This library is used to get/set values in packed structure on stack
+library Slot0PackedLibrary {
+    uint256 private constant UINT160_MASK = 0x00FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF;
+    uint256 private constant INT24_MASK = 0xFFFFFF;
+    uint256 private constant UINT24_MASK = 0xFFFFFF;
```
shuhuiluo, line 44:
```suggestion
    uint160 private constant UINT160_MASK = 0x00FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF;
    uint24 private constant INT24_MASK = 0xFFFFFF;
    uint24 private constant UINT24_MASK = 0xFFFFFF;
```

src/types/Slot0.sol
```diff
@@ -0,0 +1,112 @@
...
+
+    uint256 private constant TICK_OFFSET = 160;
+    uint256 private constant PROTOCOL_FEE_OFFSET = 184;
+    uint256 private constant LP_FEE_OFFSET = 208;
```
shuhuiluo, line 48:
```suggestion
    uint8 private constant TICK_OFFSET = 160;
    uint8 private constant PROTOCOL_FEE_OFFSET = 184;
    uint8 private constant LP_FEE_OFFSET = 208;
```

src/types/Slot0.sol
```diff
@@ -0,0 +1,112 @@
...
+        /// @solidity memory-safe-assembly
+        assembly {
+            _sqrtPriceX96 := and(UINT160_MASK, _packed)
+        }
+    }
+
+    function tick(Slot0Packed _packed) internal pure returns (int24 _tick) {
+        /// @solidity memory-safe-assembly
+        assembly {
+            _tick := and(INT24_MASK, shr(TICK_OFFSET, _packed))
```
shuhuiluo, line 61:
```suggestion
            _tick := signextend(2, shr(TICK_OFFSET, _packed))
```
Makes it two's complement.

src/types/Slot0.sol
```diff
@@ -0,0 +1,112 @@
...
+        /// @solidity memory-safe-assembly
+        assembly {
+            _protocolFee := and(UINT24_MASK, shr(PROTOCOL_FEE_OFFSET, _packed))
+        }
+    }
+
+    function lpFee(Slot0Packed _packed) internal pure returns (uint24 _lpFee) {
+        /// @solidity memory-safe-assembly
+        assembly {
+            _lpFee := and(UINT24_MASK, shr(LP_FEE_OFFSET, _packed))
```
shuhuiluo, line 75:
```suggestion
            _lpFee := shr(LP_FEE_OFFSET, _packed)
```

shuhuiluo (changes requested):

src/PoolManager.sol
```diff
@@ -114,8 +115,12 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
...
         if (key.tickSpacing > MAX_TICK_SPACING) revert TickSpacingTooLarge();
         if (key.tickSpacing < MIN_TICK_SPACING) revert TickSpacingTooSmall();
-        if (key.currency0 >= key.currency1) revert CurrenciesOutOfOrderOrEqual();
-        if (!key.hooks.isValidHookAddress(key.fee)) revert Hooks.HookAddressNotValid(address(key.hooks));
+        if (key.currency0 >= key.currency1) {
+            revert CurrenciesOutOfOrderOrEqual();
+        }
+        if (!key.hooks.isValidHookAddress(key.fee)) {
+            revert Hooks.HookAddressNotValid(address(key.hooks));
+        }
```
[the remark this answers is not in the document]
hensha256: agreed - we use foundry.toml's default line length

src/PoolManager.sol
```diff
@@ -59,9 +60,9 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
         override
         returns (uint160 sqrtPriceX96, int24 tick, uint24 protocolFee, uint24 lpFee)
     {
-        Pool.Slot0 memory slot0 = pools[id].slot0;
+        Slot0Packed slot0 = pools[id].slot0.loadPacked();
```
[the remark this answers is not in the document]
snreynolds: I see that Slot0 is still used - but why? If storage is already packed, why do we need to manually pack. 

Id prefer if we could write this such that we match how storage is packed and so we only really need to have the type

Slot0 is bytes32;

snreynolds:
Btw in general, I do really like this change and it would be great to get tests I think! cc @hensha256

src/PoolManager.sol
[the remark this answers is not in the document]
0xVolosnikov: > yeah you can run `forge fmt`

Doesn't help. Maybe I'm missing some additional config?

src/PoolManager.sol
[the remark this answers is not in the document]
0xVolosnikov: Built-in structure is somewhat easier for fields change process. So I tried to keep it. 
Anyway, I removed memory-based structure completely.

src/types/Slot0.sol
0xVolosnikov: For some reason gas increases

src/types/Slot0.sol
0xVolosnikov: No effects

src/types/Slot0.sol
shuhuiluo: Great. That means solc is smart enough to use `PUSH1` for these instead of `PUSH32` so bytecode size is already optimized!

src/types/Slot0.sol
```diff
@@ -0,0 +1,95 @@
...
+    // the current tick
+    int24 tick;
+    // protocol fee, expressed in hundredths of a bip
+    // upper 12 bits are for 1->0, and the lower 12 are for 0->1
+    // the maximum is 1000 - meaning the maximum protocol fee is 0.1%
+    // the protocolFee is taken from the input first, then the lpFee is taken from the remaining input
+    uint24 protocolFee;
+    // used for the lp fee, either static at initialize or dynamic via hook
+    uint24 lpFee;
+*/
```
snreynolds, line 18: Could you maybe just also add a comment that we dont use a struct so that we get the benefits of a packed structure when we reference slot0 in memory/stack

src/types/Slot0.sol
```diff
@@ -0,0 +1,95 @@
...
+
+type Slot0 is bytes32;
+
+using Slot0Library for Slot0 global;
+
+library Slot0Library {
+    uint160 private constant UINT160_MASK = 0x00FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF;
```
snreynolds, line 25: why is this not just `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF` ?

src/types/Slot0.sol
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    function tick(Slot0 _packed) internal pure returns (int24 _tick) {
+        /// @solidity memory-safe-assembly
+        assembly {
+            _tick := signextend(2, shr(TICK_OFFSET, _packed))
+        }
+    }
+
+    function protocolFee(Slot0 _packed) internal pure returns (uint24 _protocolFee) {
```
snreynolds, line 48: I wonder now that we are writing getters manually if we can just get rid of the getters for 0 to 1 and 1 to 0 in ProtocolFeeLibrary and just do it directly in this library? Maybe would save some overhead - I still like having a ProtcocolFeeLibrary but worth playing around if we can remove getZeroForOneFee / getOneForZeroFee

src/types/Slot0.sol
0xVolosnikov: Compiler treats it as an address literal and is really angry because of incorrectly formatted address

src/types/Slot0.sol
0xVolosnikov: Sounds reasonable! Then we can remove `protocolFee` completely and treat it as two packed `uint12` inside of a structure.

In this case, it is logical to change the public getter for slot0 too

src/libraries/Pool.sol
```diff
@@ -111,23 +98,28 @@ library Pool {
...
 
         tick = TickMath.getTickAtSqrtPrice(sqrtPriceX96);
 
-        self.slot0 = Slot0({sqrtPriceX96: sqrtPriceX96, tick: tick, protocolFee: protocolFee, lpFee: lpFee});
+        self.slot0 = 
+            Slot0.wrap(bytes32(0))
+                .setSqrtPriceX96(sqrtPriceX96)
+                .setTick(tick)
+                .setProtocolFee(protocolFee)
+                .setLpFee(lpFee);
```
shuhuiluo, line 110: How about a `pack` method if putting all components together? Should be more efficient.

0xVolosnikov: I'm afraid that this will lead to a somewhat significant increase in bytecode size for the sake of a tiny gain in a function that is called once. In my opinion, it's better to have smaller bytecode, but more aggressive optimizer settings

0xVolosnikov: I'll check what it does anyway

0xVolosnikov: @shuhuiluo I tried to implement it as:

```solidity
  function pack(uint160 _sqrtPriceX96, int24 _tick, uint24 _protocolFee, uint24 _lpFee) internal pure returns (Slot0 _result) {
        assembly {
          _result := and(UINT160_MASK, _sqrtPriceX96)
          _result := or(_result, shl(TICK_OFFSET, and(INT24_MASK, _tick)))
          _result := or(_result, shl(PROTOCOL_FEE_OFFSET, and(UINT24_MASK, _protocolFee)))
          _result := or(_result, shl(LP_FEE_OFFSET, and(UINT24_MASK, _lpFee)))
      }      
  }
```

and I see no changes in snapshots at all...

src/types/Slot0.sol
0xVolosnikov: Added

src/libraries/Pool.sol
shuhuiluo: Interesting. The compiler may be able to reduce the math and the masking on `0`. Still, appreciate it.

src/types/Slot0.sol
0xVolosnikov: I tried to implement `protocolFeeZeroForOne`/`protocolFeeOneForZero` as `uint12` in `Slot0` and do not see any gains in snapshots, looks like the optimizer itself substitutes and optimizes getters.
Probably such a change only makes sense to simplify the code, but imo the current version is already quite good.

shuhuiluo: Simple getters without side effects should be automatically inlined by the compiler.

src/types/Slot0.sol
hensha256: could just change it to `type(uint160).max`? That feels more reliable anyway as you dont need to check you have the right number of Fs

src/types/Slot0.sol
```diff
@@ -0,0 +1,98 @@
...
+ * int24 tick;
+ * 
+ * Protocol fee, expressed in hundredths of a bip, upper 12 bits are for 1->0, and the lower 12 are for 0->1
+ * the maximum is 1000 - meaning the maximum protocol fee is 0.1%
+ * the protocolFee is taken from the input first, then the lpFee is taken from the remaining input
+ * uint24 protocolFee;
+ * 
+ * Used for the lp fee, either static at initialize or dynamic via hook
+ * uint24 lpFee;
+ */
```
hensha256, line 22: imo worth writing in this comment about the unused bits. How many, where they are etc.

src/types/Slot0.sol
0xVolosnikov: > could just change it to `type(uint160).max`?

Only direct number constants can be used in assembly. We can replace it with `type(uin160).max` but it will be a bit messier

src/types/Slot0.sol
hensha256: Even a single line like

`24 bits empty | 24 bits lpFee | 24 bits protocolFee | 24 bits tick | 160 bits sqrtPriceX96`

i think would make this super clear what the layout is and the direction of the packing

0xVolosnikov: Added

src/types/Slot0.sol
```diff
@@ -0,0 +1,98 @@
...
+            _result := or(and(not(shl(TICK_OFFSET, INT24_MASK)), _packed), shl(TICK_OFFSET, and(INT24_MASK, _tick)))
+        }
+    }
+
+    function setProtocolFee(Slot0 _packed, uint24 _protocolFee) internal pure returns (Slot0 _result) {
+        /// @solidity memory-safe-assembly
+        assembly {
+            _result :=
+                or(
+                    and(not(shl(PROTOCOL_FEE_OFFSET, UINT24_MASK)), _packed),
```
hensha256, line 88: presumably optimiser is catching that `not(shl(PROTOCOL_FEE_OFFSET, UINT24_MASK))`  is always the same?

src/types/Slot0.sol
hensha256: oh yeah damn, annoying

src/types/Slot0.sol
0xVolosnikov: Will check it

0xVolosnikov: Indeed, attempts to use constants instead of `not(shl(...))` do not change anything or even make gas worse

src/types/Slot0.sol
```diff
@@ -0,0 +1,101 @@
...
+ * Used for the lp fee, either static at initialize or dynamic via hook
+ * uint24 lpFee;
+ */
+type Slot0 is bytes32;
+
+using Slot0Library for Slot0 global;
+
+library Slot0Library {
+    uint160 internal constant UINT160_MASK = 0x00FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF;
+    uint24 internal constant INT24_MASK = 0xFFFFFF;
```
0xVolosnikov, line 32: `INT24_MASK` can be removed since it is equal to `UINT24_MASK`

0xVolosnikov:
Added tests

test/types/Slot0.t.sol
```diff
@@ -0,0 +1,30 @@
...
+import {Test} from "forge-std/Test.sol";
+import {Slot0, Slot0Library} from "../../src/types/Slot0.sol";
+
+contract TestSlot0 is Test {
+    function test_slot0_constants_masks() public pure {
+        assertEq(Slot0Library.UINT160_MASK, type(uint160).max);
+        assertEq(Slot0Library.UINT24_MASK, type(uint24).max);
+    }
+
+    function test_slot0_fuzz_pack_unpack(uint160 sqrtPriceX96, int24 tick, uint24 protocolFee, uint24 lpFee) public pure {
```
hensha256: a nit to be in keeping with the repo

```suggestion
    function test_fuzz_slot0_pack_unpack(uint160 sqrtPriceX96, int24 tick, uint24 protocolFee, uint24 lpFee) public pure {
```

hensha256:
@snreynolds anything else you wanna see before we merge?

src/types/Slot0.sol
```diff
@@ -0,0 +1,101 @@
...
+ * Used for the lp fee, either static at initialize or dynamic via hook
+ * uint24 lpFee;
+ */
+type Slot0 is bytes32;
+
+using Slot0Library for Slot0 global;
+
+library Slot0Library {
+    uint160 internal constant UINT160_MASK = 0x00FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF;
+    uint24 internal constant UINT24_MASK = 0xFFFFFF;
```
hensha256, line 33: nit: would rather this was called `24_BIT_MASK` or something like that now that its used for both int24 and uint24

0xVolosnikov: Changed to `MASK_24_BITS`

hensha256 (commented):
1 final nit, plus you need to handle merge conflicts (sorry we're merging lots today!). Then we can merge!

src/PoolManager.sol
```diff
@@ -18,6 +18,7 @@ import {ERC6909Claims} from "./ERC6909Claims.sol";
 import {PoolId, PoolIdLibrary} from "./types/PoolId.sol";
 import {BalanceDelta, BalanceDeltaLibrary, toBalanceDelta} from "./types/BalanceDelta.sol";
 import {BeforeSwapDelta} from "./types/BeforeSwapDelta.sol";
+import {Slot0} from "./types/Slot0.sol";
```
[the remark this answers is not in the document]
0xVolosnikov: Removed

src/libraries/Pool.sol
```diff
@@ -297,25 +286,28 @@ library Pool {
...
-        state.tick = slot0Start.tick;
+        state.sqrtPriceX96 = slot0Start.sqrtPriceX96();
+        state.tick = slot0Start.tick();
         state.feeGrowthGlobalX128 = zeroForOne ? self.feeGrowthGlobal0X128 : self.feeGrowthGlobal1X128;
         state.liquidity = liquidityStart;
 
         // if the beforeSwap hook returned a valid fee override, use that as the LP fee, otherwise load from storage
-        slot0Start.lpFee =
-            params.lpFeeOverride.isOverride() ? params.lpFeeOverride.removeOverrideAndValidate() : slot0Start.lpFee;
+        {
```
hensha256, line 304: nice thank you!

hensha256 (approved):

PR #615, opened 2024-04-25
Optimize access to allowance mapping in ERC6909

0xVolosnikov:
## Description of changes

The proposed changes reduce the number of calculations of slot addresses in storage.


src/ERC6909.sol
```diff
@@ -44,8 +44,11 @@ abstract contract ERC6909 is IERC6909Claims {
 
     function transferFrom(address sender, address receiver, uint256 id, uint256 amount) public virtual returns (bool) {
         if (msg.sender != sender && !isOperator[sender][msg.sender]) {
-            uint256 allowed = allowance[sender][msg.sender][id];
-            if (allowed != type(uint256).max) allowance[sender][msg.sender][id] = allowed - amount;
+            mapping(uint256 => uint256) storage _allowances = allowance[sender][msg.sender];
```
shuhuiluo, line 47: A double nested mapping requires 3 `keccak256` to index. Consider using a flat mapping with key encoding similar to `positions`.

0xVolosnikov: > A double nested mapping requires 3 `keccak256` to index. Consider using a flat mapping with key encoding similar to `positions`.

Yeah, that would be cheaper to use. To remain compatible with the standard, we can add the appropriate getters. Moreover, this applies to all nested mappings here.

However, this is a more massive change and would probably be better done in a separate pull request. Wdyt?

hensha256: Yeah would like to see if you can do it similarly to how we have positions where we hash first and then have a `bytes32` mapping. This current change doesnt seem to improve gas either - just bytecode? So happy to close this one and would love to see the `bytes32` one instead! What do you think?

0xVolosnikov: > Yeah would like to see if you can do it similarly to how we have positions where we hash first and then have a `bytes32` mapping. This current change doesnt seem to improve gas either - just bytecode? So happy to close this one and would love to see the `bytes32` one instead! What do you think?

It can’t change only bytecode as I understand it. So maybe reason in the lack of test coverage here?

Agree with bytes32 implementation, please feel free to close this PR!

hensha256:
closing as agreed!

Issue #616, opened 2024-04-25
Pool.SwapState can be passed in Pool.Swap and modified by reference


0xVolosnikov:
### Component

Gas Optimization

### Describe the suggested feature and problem it solves.

This block of code allocates memory for the Pool.SwapState structure:

https://github.com/Uniswap/v4-core/blob/e3c4dadb47fde3b640166c5ab1ac241e13acac57/src/PoolManager.sol#L221-L229

But inside of Pool.swap function memory is being allocated again:

https://github.com/Uniswap/v4-core/blob/e3c4dadb47fde3b640166c5ab1ac241e13acac57/src/libraries/Pool.sol#L306

https://github.com/Uniswap/v4-core/blob/e3c4dadb47fde3b640166c5ab1ac241e13acac57/src/libraries/Pool.sol#L335-L342

Use of this structure can be implemented more efficiently if the structure is passed inside the swap method and modified internally by reference. This approach is less explicit and more difficult to understand, but by my estimation it saves about 200 gas.

### Describe the desired implementation.

```solidity
// PoolManager.swap
//...
        Pool.SwapState memory state;
        (delta, feeForProtocol, swapFee) = pools[id].swap(
            Pool.SwapParams({
                tickSpacing: key.tickSpacing,
                zeroForOne: params.zeroForOne,
                amountSpecified: params.amountSpecified,
                sqrtPriceLimitX96: params.sqrtPriceLimitX96
            }),
            state
        );
//...

// Pool.swap
    function swap(
        State storage self,
        SwapParams memory params,
        SwapState memory state
    )
        internal
        returns (BalanceDelta result, uint256 feeForProtocol, uint24 swapFee)
    {
//...
        state.amountSpecifiedRemaining = params.amountSpecified;
        state.amountCalculated = 0;
        state.sqrtPriceX96 = slot0Start.sqrtPriceX96;
        state.tick = slot0Start.tick;
        state.feeGrowthGlobalX128 = zeroForOne
            ? self.feeGrowthGlobal0X128
            : self.feeGrowthGlobal1X128;
        state.liquidity = cache.liquidityStart;
//...
```

PR #617, opened 2024-04-25
Optimize swap state structure allocation

0xVolosnikov:
## Related Issue
#616


src/libraries/Pool.sol
```diff
@@ -332,14 +332,12 @@ library Pool {
...
-        state = SwapState({
-            amountSpecifiedRemaining: params.amountSpecified,
-            amountCalculated: 0,
-            sqrtPriceX96: slot0Start.sqrtPriceX96,
-            tick: slot0Start.tick,
-            feeGrowthGlobalX128: zeroForOne ? self.feeGrowthGlobal0X128 : self.feeGrowthGlobal1X128,
-            liquidity: cache.liquidityStart
-        });
+        state.amountSpecifiedRemaining = params.amountSpecified;
+        state.amountCalculated = 0;
```
shuhuiluo, line 336:
```suggestion
```
Can be removed.

src/PoolManager.sol
```diff
@@ -219,13 +219,14 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
         uint256 feeForProtocol;
         uint24 swapFee;
         Pool.SwapState memory state;
-        (delta, feeForProtocol, swapFee, state) = pools[id].swap(
+        (delta, feeForProtocol, swapFee) = pools[id].swap(
```
shuhuiluo, line 222:
```suggestion
        (BalanceDelta delta, uint256 feeForProtocol, uint24 swapFee, Pool.SwapState memory state) = pools[id].swap(
```
That's a good find. Instead of making more intrusive changes, consider declaring types at assignment and returning `delta` explicitly while keeping the way you initialize `state` in `Pool.swap`.

shuhuiluo (changes requested):

src/PoolManager.sol
hensha256: Yep really like this PR! But agree that this comment might help it read easier 🙏

0xVolosnikov: This option uses a little more gas, but it is more readable and understandable

hensha256 (approved):

PR #619, opened 2024-04-26
Update tests to use lastCallGas and --isolate

hensha256:
- libraries still using snapStart and snapEnd as `lastCallGas` doesnt work for them
- new function snapLastCall which was merged to the submodule here https://github.com/marktoda/forge-gas-snapshot/pull/10
- tests must be run with `forge test --isolate`
- modifier `noIsolate` added for non-isolated tests e.g. transient storage tests


marktoda (commented):
lets gooo

marktoda (approved):

PR #621, opened 2024-04-28
Optimize ERC6909 storage layout

0xVolosnikov:
Optimized storage layout of ERC6909 mappings (as discussed in #615)

## Description of changes

Calculating the corresponding storage slots is simplified and requires only one keccak calculation. 

The algorithm for calculating addresses in storage differs from the standard one for Solidity (keccak(key.slot)), but this should not be a problem, since there are public getters


src/ERC6909.sol
```diff
@@ -87,14 +117,109 @@ abstract contract ERC6909 is IERC6909Claims {
...
+    }
+
+    function _getBalanceSlot(address owner, uint256 id) internal pure returns (bytes32 balanceSlot) {
+        // balanceSlot = keccak256(abi.encode(owner, id, _balances.slot));
+        /// @solidity memory-safe-assembly
+        assembly {
+            mstore(0x20, id)
+            mstore(0x00, owner)
+            let pointer := mload(0x40)
+            mstore(0x40, _balances.slot)
```
hensha256, line 196: I feel like we dont need to use `_balances.slot` at all. We need something to make sure that there cannot be clashes - but i feel like we could use something just packed into the top of the owner slot?

Like save some constant 
```
bytes1 constant BALANCE_SLOT = bytes1(bytes32(keccak256("BalanceSlot")))
```

And then for the hash just put that byte into the top of the same slot as `owner`?

hensha256: that will then save going into the 0x40 slot, and no need to backup and restore the fmp too

hensha256 (changes requested):
Left a few thoughts :)

src/ERC6909.sol
hensha256: Same for the others too of course
And really i would put the hardcoded value in the constant and just say the calculation in a comment because constants are evaluated at runtime time 🙏

src/ERC6909.sol
```diff
@@ -87,14 +117,109 @@ abstract contract ERC6909 is IERC6909Claims {
...
+            sstore(balanceSlot, balanceOfValue)
+        }
+    }
+
+    function _getOperatorSlot(address owner, address spender) internal pure returns (bytes32 operatorSlot) {
+        // balanceSlot = keccak256(abi.encode(owner, spender, _operators.slot));
+        /// @solidity memory-safe-assembly
+        assembly {
+            mstore(0x20, spender)
+            mstore(0x00, owner)
```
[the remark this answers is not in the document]
0xVolosnikov: Tried to minimize amount of stack operations. Will check if it makes sense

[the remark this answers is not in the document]
hensha256: Ok makes sense - if its better this way then keep it like this!

src/ERC6909.sol
0xVolosnikov: Nice idea! Will implement it

about constants: I thought that it was fixed in one of the previous Solidity versions, isn’t it?

hensha256: Sadly I dont think so. See PR #513 where I realised it wasnt haha.

0xVolosnikov:
It is better to double-check params cleaning

0xVolosnikov:
Addresses cleanup really bothered me, so I changed memory layout to clean higher bits before keccak. It is a little bit more gas-expensive, but safer in my opinion

0xVolosnikov:
Will simplify allowance spending

Issue #624, opened 2024-04-29
Add a "sparse" `extsload` method that is able to load and return from a list of keys


Philogy:
### Component

General design optimization (improving efficiency, cleanliness, or developer experience)

### Describe the suggested feature and problem it solves.

Currently the V4 singleton has two "extsload" style methods:
1. `extsload(bytes32 slot)` which reads a single slot
2. `extsload(bytes32 startSlot, uint256 nSlots)` which reads `nSlots` continuous `startSlots` from the given slot

Most of the storage relationships in the contract rely on mappings, which use the `keccak256` function to derive their storage slots making storage very sparse. This means that the only use case for the continuous `extsload` is for reading structs (although this could change with #623 ). Nevertheless, most storage is sparse and calling getters multiple times is quite expensive (~1.4k of call overhead) therefore having a method that takes a list of keys `bytes32[]` and returns their values could aid contracts such as external fee/staking contracts to efficiently retrieve data from the contract.

### Describe the desired implementation.

Add a third `extsload` method that takes a single `bytes32[]` array as an input and returns the values at the desired storage slots.

shuhuiluo:
Suggested here.
https://github.com/Uniswap/v4-core/pull/559#pullrequestreview-1982030657

Philogy:
> Suggested here. [#559 (review)](https://github.com/Uniswap/v4-core/pull/559#pullrequestreview-1982030657)

Oh yeah, missed that although my implementation in #625 is more optimized than the suggested implementation in #559

shuhuiluo:
> > Suggested here. [#559 (review)](https://github.com/Uniswap/v4-core/pull/559#pullrequestreview-1982030657)
> 
> Oh yeah, missed that although my implementation in #625 is more optimized than the suggested implementation in #559

As far as I test, the [impl in my comment](https://github.com/Uniswap/v4-core/pull/559#pullrequestreview-1982030657) is 154 gas cheaper on your test.

PR #625, opened 2024-04-29
Add sparse extsload

Philogy:
## Related Issue
#624

## Description of changes
Add an optimized implementation for a function that takes an array of storage keys, loads their values from storage and returns them.


saucepoint (changes requested):
can you add the new function signature to `IExtsload` so its available through `IPoolManager`

Philogy:
@saucepoint done

hensha256:
Please can you run `forge test --isolate` ? bytecode size has increased here but missing the snapshot for it

Philogy:
Updated the gas snapshot @hensha256

gretzke (approved):

PR #633, opened 2024-05-07
Before swap returns unspecified delta

hensha256:
## Related Issue

## Description of changes


shuhuiluo (changes requested):

src/interfaces/IHooks.sol
```diff
@@ -54,14 +55,15 @@ interface IHooks {
...
+    /// @return HookReturnDelta The hook's delta in token0 and token1 as a BalanceDelta.
+    /// Positive: the hook is owed/took currency, negative: the hook owes/sent currency
     function afterAddLiquidity(
         address sender,
         PoolKey calldata key,
         IPoolManager.ModifyLiquidityParams calldata params,
         BalanceDelta delta,
         bytes calldata hookData
-    ) external returns (bytes4, BalanceDelta);
+    ) external returns (bytes4, HookReturnDelta);
```
saucepoint, line 66: i think the liquidity return delta should stay as BalanceDelta

its quite weird that swap deltas are packed `(specified, unspecified)` while liquidity deltas are packed `(amount0, amount1)` and named differently than BalanceDelta

src/test/CustomCurveHook.sol
```diff
@@ -36,26 +37,17 @@ contract CustomCurveHook is BaseTestHooks {
...
-        address, /* sender **/
-        PoolKey calldata, /* key **/
-        IPoolManager.SwapParams calldata params,
-        BalanceDelta, /* delta **/
-        bytes calldata /* hookData **/
-    ) external view override onlyPoolManager returns (bytes4, int128) {
-        return (IHooks.afterSwap.selector, int128(params.amountSpecified));
+        // return -amountSpecified as specified to no-op the concentrated liquidity swap
+        HookReturnDelta hookDelta = toHookReturnDelta(int128(-params.amountSpecified), int128(params.amountSpecified));
+        return (IHooks.beforeSwap.selector, hookDelta);
```
saucepoint, line 50: curious of the gas savings for custom curves 🤔 

but overall i like how we dont need afterSwap now!

src/types/HookReturnDelta.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {BalanceDelta} from "./BalanceDelta.sol";
+
+type HookReturnDelta is bytes32;
+
+// Creates a hookReturnDelta from specified and unspecified, or amount0 and amount1
+function toHookReturnDelta(int128 upperDelta, int128 lowerDelta) pure returns (HookReturnDelta hookReturnDelta) {
```
saucepoint, line 9: for consistent naming:

```suggestion
function toHookReturnDelta(int128 deltaSpecified, int128 deltaUnspecified) pure returns (HookReturnDelta hookReturnDelta) {
```

src/types/HookReturnDelta.sol
```diff
@@ -0,0 +1,43 @@
...
+library HookReturnDeltaLibrary {
+    HookReturnDelta public constant ZERO_DELTA = HookReturnDelta.wrap(bytes32(0));
+
+    /// converts a HookReturnDelta type to a BalanceDelta type
+    /// returned by afterAddLiquidity and afterRemoveLiquidity
+    function toBalanceDelta(HookReturnDelta delta) internal pure returns (BalanceDelta balanceDelta) {
+        assembly {
+            balanceDelta := delta
+        }
+    }
```
saucepoint, line 25: wont be needed if liquidity deltas return as BalanceDelta 👍

src/test/DynamicFeesTestHook.sol
```diff
@@ -30,10 +31,10 @@ contract DynamicFeesTestHook is BaseTestHooks {
     function beforeSwap(address, PoolKey calldata key, IPoolManager.SwapParams calldata, bytes calldata)
         external
         override
-        returns (bytes4, int128)
+        returns (bytes4, HookReturnDelta)
     {
         manager.updateDynamicLPFee(key, fee);
-        return (IHooks.beforeSwap.selector, 0);
+        return (IHooks.beforeSwap.selector, HookReturnDeltaLibrary.ZERO_DELTA);
```
saucepoint, line 37: feels a lot cleaner and clearer to have a constant here 😄

src/libraries/Hooks.sol
```diff
@@ -259,27 +258,30 @@ library Hooks {
...
+        int128 hookDeltaSpecified = beforeSwapHookReturn.getSpecifiedDelta();
+        int128 hookDeltaUnspecified = beforeSwapHookReturn.getUnspecifiedDelta();
+
         if (self.hasPermission(AFTER_SWAP_FLAG)) {
-            hookDeltaUnspecified = self.callHookWithReturnDelta(
+            hookDeltaUnspecified += self.callHookWithReturnDelta(
                 abi.encodeWithSelector(IHooks.afterSwap.selector, msg.sender, key, params, swapDelta, hookData),
                 self.hasPermission(AFTER_SWAP_RETURNS_DELTA_FLAG)
-            ).toInt128();
+            ).getUnspecifiedDelta();
```
saucepoint, line 272: is a case where afterSwap takes a fee from specified currency?

src/interfaces/IHooks.sol
saucepoint: if we decide to make liquidity-return-delta a BalanceDelta, we'll probably want to rename HookReturnDelta to HookSwapDelta?

hensha256: We're thinking about this differently but its an interesting point of view. I feel like as ever I'm thinking about this from the point of view of inside of core, and you're thinking about integrators which is perfect

What I wanted was to make this function easier to handle. Previously it would just generically work with an `int256` even though none of the hooks were actually working with an int256 which felt wrong. So I gave the name `HookReturnDelta` as a generic type for any delta returned by any of the custom accounting hooks. 

```
function callHookWithReturnDelta(IHooks self, bytes memory data, bool parseReturn)
        internal
-        returns (int256 delta)
+        returns (HookReturnDelta delta)
```

Then if you look inside the actual hook calls in `Hooks.sol` it then turns each `HookReturnDelta` into the type that is actually correct for that hook. Thats why the variable names for the function to create a `HookReturnDelta` were `int128 upperDelta, int128 lowerDelta` because those deltas could either be amount0 and amount1, OR specified and unspecified, depending whether the function in question needs a BalanceDelta or 2 int128s.

However I do see that while what I've created makes things simpler internally... it makes things more confusing for integrators. So I'm going to try to find a middle ground that works for everyone!

PR #635, opened 2024-05-08
Add exttload to read transient storage from outside the contract

gretzke:
## Related Issue

## Description of changes

Fixes PROTO-133


src/test/PoolStateViewer.sol
saucepoint, line 1: hmm do we need this file? I think tests can just use the library directly

`PoolStateLibrary.getReserves(manager, currency0)` (or the syntactic sugar version manager.getReserves())

src/Exttload.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
+
+import {IExttload} from "./interfaces/IExttload.sol";
+import {Extsload} from "./Extsload.sol";
+
+/// @notice Enables public transient storage access for efficient state retrieval by external contracts.
+/// https://eips.ethereum.org/EIPS/eip-2330#rationale
+abstract contract Exttload is IExttload, Extsload {
```
saucepoint, line 9: hmm i think Exttload/IExttload should be decoupled from Extsload

while they are functionally similar, one of the reasons why we moved Extsload to a different file (with MIT) was so other projects can use the same logic

by decoupling it, other projects can pick and choose if they need Exttload, Extsload, or both

saucepoint (commented):
thank you for the PR! some minor feedback and:

i think you need to run the regenerate snapshots with `forge test --isolate`

src/test/PoolStateViewer.sol
gretzke: We cannot use the library directly, `vm.expectRevert` uses the next call, which in that case would be `manager.exttload` within the library. So the expected revert is not thrown in the next call but within the library, that's why I introduced the viewer contract

src/Exttload.sol
gretzke: fair point that a project might only want to use Exttload, but not Extsload, should I also add `exttload(bytes32 startSlot, uint256 nSlots) external view returns (bytes memory)` even though we are not using it internally?

saucepoint: since tstore is so new, im not sure if other projects will be using tstore in a sequential manner?

is it possible tstore a struct that exceeds a word / 256 bits 🤔 

i will say that sparse exttload would be interesting (i.e. read multiple reserve amounts in a single call)

i.e. the exttload equilvalent to https://github.com/Uniswap/v4-core/pull/625

src/test/PoolStateViewer.sol
saucepoint: ahh youre right, overlooked that

hensha256:
are you running snaps with `--isolate` ? Need to run `forge test --isolate`

gretzke:
Yes, the tests failing are the same failing on the base branch

src/Exttload.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
```
hensha256, line 2: nit: I think most of ours we have `^0.8.24`. Do you have a strong feeling about this?

gretzke: `Extsload` is using `>=0.6.0` so I wanted to keep that convention but adjust it to a version where `tload` exists

src/PoolManager.sol
```diff
@@ -280,11 +281,6 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
         return pools[id].getPoolBitmapInfo(word);
     }
 
-    /// @notice Temporary view function. Replaceable by transient EXTSLOAD.
-    function getReserves(Currency currency) external view returns (uint256 balance) {
```
hensha256, line 284: you should also be able to remove the following functions: `getNonzeroDeltaCount`, `currencyDelta`, `isUnlocked`

src/interfaces/IExttload.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
+
+interface IExttload {
+    /// @notice Called by external contracts to access granular pool state
+    /// @param slot Key of slot to sload
+    /// @return value The value of the slot as bytes32
```
hensha256, line 7: This natspec needs updating for transient 👯

src/Exttload.sol
hensha256: Yah makes sense

test/Sync.t.sol
```diff
@@ -125,7 +130,7 @@ contract SyncTest is Test, Deployers, GasSnapshot {
 
         // Sync has not been called.
         vm.expectRevert(Reserves.ReservesMustBeSynced.selector);
-        manager.getReserves(currency0);
+        poolStateViewer.getReserves(currency0);
 
         manager.sync(currency0);
         assertEq(manager.getReserves(currency0), currency0Balance);
```
hensha256, line 136: why is `poolStateViewer.getReserves` needed for reverts but `manager.getReserves` needed for seeing the return value?

hensha256 (commented):
few thoughts!

src/libraries/PoolStateLibrary.sol
```diff
@@ -328,6 +337,19 @@ library PoolStateLibrary {
...
+    function getReserves(IPoolManager manager, Currency currency) internal view returns (uint256 value) {
+        uint256 slot = RESERVES_OF_SLOT;
+        bytes32 key;
+        assembly {
+            mstore(0, slot)
+            mstore(32, currency)
+            key := keccak256(0, 64)
+        }
+        value = uint256(manager.exttload(key));
+        if (value == 0) revert ReservesMustBeSynced();
```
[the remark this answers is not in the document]
snreynolds: I am leaning towards having this external function should just return whatever is actually in transient storage - so 0 if not synced and ZERO_BALANCE constant if not synced and the balance of the reserves is 0

test/Sync.t.sol
gretzke: > We cannot use the library directly, vm.expectRevert uses the next call, which in that case would be manager.exttload within the library. So the expected revert is not thrown in the next call but within the library, that's why I introduced the viewer contract

saucepoint:
thoughts on moving the exttload's (getReserves, currencyDelta, etc) to `TransientStateLibrary`

vs renaming `PoolStateLibrary` to `StateLibrary`?

---

A single import is nice, but it feels overloaded/monolithic

src/libraries/PoolStateLibrary.sol
```diff
@@ -358,6 +376,46 @@ library PoolStateLibrary {
...
+            mstore(0, caller_)
+            mstore(32, currency)
+            key := keccak256(0, 64)
+        }
+        return int256(uint256(manager.exttload(key)));
+    }
+
+    /// @notice Returns whether the contract is unlocked or not
+    function isUnlocked(IPoolManager manager) external view returns (bool) {
+        return manager.exttload(bytes32(IS_UNLOCKED_SLOT)) != 0x0;
```
hensha256, line 416: do you think it would make sense to either have our global vars as `bytes32`, or have `exttload` accept a uint256? feels wrong to cast so often

gretzke: I'd prefer having global vars bytes32, I think bytes32 for arguments and return data for `exttload` is the most neutral option

snreynolds: like lets change IS_UNLOCKED_SLOT and any other slot variables to `bytes32`

src/libraries/TransientStateLibrary.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.21;
+
+import {PoolId} from "../types/PoolId.sol";
+import {IPoolManager} from "../interfaces/IPoolManager.sol";
+import {Currency} from "../types/Currency.sol";
+import {Position} from "./Position.sol";
+
+library TransientStateLibrary {
+    /// uint256(keccak256("ReservesOf")) - 1
```
hensha256, line 10: these comments should no longer say `uint256(...)`

saucepoint (approved):

src/libraries/StateLibrary.sol
```diff
@@ -3,9 +3,10 @@ pragma solidity ^0.8.21;
 
 import {PoolId} from "../types/PoolId.sol";
 import {IPoolManager} from "../interfaces/IPoolManager.sol";
+import {Currency} from "../types/Currency.sol";
 import {Position} from "./Position.sol";
 
-library PoolStateLibrary {
+library StateLibrary {
```
snreynolds, line 9: maybe we actually rename this back to PoolStateLibrary? I think there may end up being more libraries we need for various view function in PoolManager (ie for some view functions in ProtocolFee.sol) and we might want them in their own lib?

snreynolds (approved):
one tiny comment, but lmk if you disagree and think the other view functions should live together in a combined StateLibrary.sol file

snreynolds (approved):

src/libraries/StateLibrary.sol
snreynolds: will just leave for now

snreynolds (approved):
🔥

PR #637, opened 2024-05-09
Swapmath tests

dianakocsis:
## Related Issue

## Description of changes

related to PROTO-140


test/libraries/SwapMath.t.sol
```diff
@@ -183,8 +183,8 @@ contract SwapMathTest is Test, GasSnapshot {
     ) public pure {
         vm.assume(sqrtPriceRaw > 0);
         vm.assume(sqrtPriceTargetRaw > 0);
-        vm.assume(feePips > 0);
-        vm.assume(feePips < 1e6);
+        vm.assume(feePips >= 0);
+        vm.assume(feePips <= 1e6);
```
snreynolds, line 187: doesnt work for exact out

snreynolds: might need to expect something different for exact out and 100%

snreynolds (commented):
lookin good

test/libraries/SwapMath.t.sol
```diff
@@ -208,8 +213,14 @@ contract SwapMathTest is Test, GasSnapshot {
 
         // didn't reach price target, entire amount must be consumed
         if (sqrtQ != sqrtPriceTargetRaw) {
+            uint256 absAmtRemaining;
+            if (amountRemaining == type(int256).min) {
+                absAmtRemaining = uint256(type(int256).max) + 1;
+            } else if (amountRemaining < 0) {
+                absAmtRemaining = uint256(-amountRemaining);
+            }
```
hensha256, line 221: does it not need an `else absAmtRemaining = amountRemaining` ?

dianakocsis: yes! its because type(int256).min and type(int256).max do not have the same absolute value

dianakocsis: so i had to convert it to uint256 first then add 1

dianakocsis: but itll still do the else statement so its more clear!

hensha256 (approved):

PR #639, opened 2024-05-10
beforeSwap Returns Dynamic Fee

### LINKED ISSUE #554: Return swapFee in beforeSwap for hooks with relevant permission

saucepoint:
## Related Issue
Closes #554

- [ ] define the sentinel _somewhere_ (LPFeeLibrary?)

## Description of changes

---


src/PoolManager.sol
```diff
@@ -267,19 +267,19 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     {
         if (params.amountSpecified == 0) revert SwapAmountCannotBeZero();
 
-        PoolId id = key.toId();
-        _checkPoolInitialized(id);
+        _checkPoolInitialized(key.toId());
```
saucepoint, line 270: stack-too-deep means we have to hash PoolId twice 😭

.forge-snapshots/swap with return dynamic fee.snap
```diff
@@ -0,0 +1 @@
+158080
```
saucepoint, line 1: hmm 2275 gas saved vs using `manager.updateDynamicLPFee()` in beforeSwap

.forge-snapshots/swap with return dynamic fee.snap
```diff
@@ -0,0 +1 @@
+158016
```
saucepoint, line 1: 2339 gas saved vs using `updateDynamicLPFee()` in beforeSwap

src/libraries/Hooks.sol
```diff
@@ -193,6 +207,7 @@ library Hooks {
     }
 
     /// @notice calls afterModifyLiquidity hook if permissioned and validates return value
+    /// @dev if noSelfCall bypasses the call, hookDelta will be 0
```
snreynolds, line 210: i think maybe this is out of date? this one doesnt have the noSelfCall modifier

src/libraries/Hooks.sol
```diff
@@ -233,15 +248,19 @@ library Hooks {
...
+        if (msg.sender == address(self)) return (amountToSwap, hookDeltaSpecified, swapFee);
         if (self.hasPermission(BEFORE_SWAP_FLAG)) {
-            hookDeltaSpecified = self.callHookWithReturnDelta(
+            (hookDeltaSpecified, _swapFee) = self.callHookWithReturnDeltaAndFee(
                 abi.encodeWithSelector(IHooks.beforeSwap.selector, msg.sender, key, params, hookData),
                 self.hasPermission(BEFORE_SWAP_RETURNS_DELTA_FLAG)
-            ).toInt128();
+            );
+
+            if (key.fee.isDynamicFee()) swapFee = _swapFee;
```
snreynolds, line 263: could you do something similar to how we do hookDeltaSpecified? where we pass in a bool into `callHookWithReturnDeltaAndFee` for if the hook has permission to set the dynamic fee otherwise return 0

saucepoint:
replaced by #648

PR #641, opened 2024-05-11
Make `BalanceDelta` great

shuhuiluo:
## Related Issue
Which issue does this pull request resolve?

## Description of changes
Optimize `eq`, `neq`, `add` and `sub` for `BalanceDelta`. Refactor `amount0` and `amount1` for correctness.

PR #586, opened 2024-04-14
fix: Replace `shr` with `sar` for integer types

shuhuiluo:
## Related Issue
Which issue does this pull request resolve?

## Description of changes
This commit updates the `shr` instruction to `sar` in the BalanceDelta and Pool contracts, as `sar` is the correct way to shift right when dealing with signed integers. This change ensures that the sign is preserved during bit shifting operations, preventing any potential bugs related to negative amounts or liquidity values.

PR #642, opened 2024-05-11
Before swap returns unspecified delta

hensha256:
Replacing #633 due to a complex merge.

Adds an unspecified delta returned by beforeSwap, so that hooks that want to edit both deltas need not use both beforeSwap and afterSwap


src/interfaces/IHooks.sol
```diff
@@ -103,7 +104,7 @@ interface IHooks {
         PoolKey calldata key,
         IPoolManager.SwapParams calldata params,
         bytes calldata hookData
-    ) external returns (bytes4, int128);
+    ) external returns (bytes4, BeforeSwapDelta);
```
qiwu7, line 107: nit: should probably update the @return comment for this

src/libraries/Hooks.sol
```diff
@@ -259,27 +258,30 @@ library Hooks {
         IPoolManager.SwapParams memory params,
         BalanceDelta swapDelta,
         bytes calldata hookData,
-        int128 hookDeltaSpecified
-    ) internal returns (BalanceDelta swapperDelta, BalanceDelta hookDelta) {
+        HookReturnDelta beforeSwapHookReturn
+    ) internal returns (BalanceDelta, BalanceDelta) {
```
qiwu7, line 262: Can you elaborate why `beforeSwap` can modify both `specified` and `unspecified` balances, while `afterSwap` can only modify the `unspecified` balance?

What are some of the possible scenarios where `afterSwap` only modify `unspecified` while `beforeSwap` modified both? Will it be better off if both before/after can modify both?

qiwu7: In addition, is there a way to know the unspecified delta amount in the `beforeSwap` hook? If i need to take a portion (say 0.1%) of the unspecified, is it possible to do it in the beforeSwap hook?

src/libraries/Hooks.sol
```diff
@@ -233,15 +229,18 @@ library Hooks {
...
-        returns (int256 amountToSwap, int128 hookDeltaSpecified)
+        returns (int256 amountToSwap, HookReturnDelta hookReturn)
     {
         amountToSwap = params.amountSpecified;
-        if (msg.sender == address(self)) return (amountToSwap, hookDeltaSpecified);
+        if (msg.sender == address(self)) return (amountToSwap, HookReturnDeltaLibrary.ZERO_DELTA);
+
         if (self.hasPermission(BEFORE_SWAP_FLAG)) {
-            hookDeltaSpecified = self.callHookWithReturnDelta(
+            hookReturn = self.callHookWithReturnDelta(
```
qiwu7, line 238: Is there a way to know the unspecified delta amount in the beforeSwap hook? If i need to take a portion (say 0.1%) of the unspecified, is it possible to do it in the beforeSwap hook?

src/libraries/Hooks.sol
hensha256: hi! i think theres a few key points to explain this:
- the `specified` amount is the amount that defines the exact amount that a user wants to swap - whether input or output. No matter what the hook returns in the `beforeSwap` hook to alter the amount actually swapped on the pool, the user will see a difference of `amountSpecified` in their wallet to maintain expected behaviour.
- After the swap has happened, its therefore impossible to alter this `specified` amount because its already been swapped on the pool and that cannot be reversed. That is why there is no `deltaSpecified` in the `afterSwap` hook. `deltaSpecified` is to alter how much is to be swapped on the pool - so in `afterSwap` it is too late
- For `deltaUnspecified` there are 2 main scenarios:
1. Hooks wanting to take a % (or some formula) of the unspecified currency of a trade to take a fee. The unspecified amount of a swap on a pool is not known until the actual swap has taken place (unless you simulate the full swap and liquidity amounts to calculate it elsewhere). Hooks like this will want to therefore apply this delta in the `afterSwap` hook as they cannot know how much to take until after the swap has taken place
2. Hooks that implement a "custom curve". This means that the hooks actually do not allow swaps to happen in the V4 Concentrated Liquidity pool and instead implement custom swap logic internal to the hook. These hooks will return `-amountSpecified` as their `specified` delta, so that the V4 pool will trade `amountSpecified-amountSpecified = 0` on the V4 pool, aka no swap occurs. Instead they then calculate their own custom unspecified currency amount for the trade. These hooks _do not_ need to wait for `afterSwap` hook to know their `deltaSpecified` because their formula is independent of the swap logic in the pool manager. Before this pull request such hooks had to return this deltaUnspecified in the `afterSwap` hook which is unnecessarily inefficient. This PR lets those hooks return both values in `beforeSwap`.

I do not envision there being a good reason for a hook to need to return a `deltaUnspecified` in _both_ the before and after swap hooks. Most should just choose 1 - depending if their deltaUnspecified is independent of the unspecified of the swap on V4, or not. However it is technically possible to return both because the code does not stop it.

hensha256: Does that make sense? 😅 I get this question a lot so definitely want to find an explanation that makes sense for us to put in the docs haha

qiwu7: Thanks for the detailed explanation! This makes lots of sense!

What I am trying to do is to define a `preferred currency` to take `hookFee` on. So this `preferred currency` could either be a `specified currency` or an `unspecified currency`. I think in that case I will have to have both before&after hooks, and then depends on the direction and the token specified to decide where to take the hook fee.

.forge-snapshots/swap CA custom curve + swap noop.snap
```diff
@@ -1 +1 @@
-139399
\ No newline at end of file
+135875
```
snreynolds, line 1: this seems pretty compelling

src/interfaces/IHooks.sol
snreynolds: think this still needs updating

src/libraries/Hooks.sol
```diff
@@ -233,21 +236,31 @@ library Hooks {
...
-        if (msg.sender == address(self)) return (amountToSwap, hookDeltaSpecified);
+        if (msg.sender == address(self)) return (amountToSwap, BeforeSwapDeltaLibrary.ZERO_DELTA);
+
         if (self.hasPermission(BEFORE_SWAP_FLAG)) {
-            hookDeltaSpecified = self.callHookWithReturnDelta(
-                abi.encodeWithSelector(IHooks.beforeSwap.selector, msg.sender, key, params, hookData),
-                self.hasPermission(BEFORE_SWAP_RETURNS_DELTA_FLAG)
-            ).toInt128();
+            bool canReturnDelta = self.hasPermission(BEFORE_SWAP_RETURNS_DELTA_FLAG);
+            hookReturn = BeforeSwapDelta.wrap(
```
snreynolds, line 246: Instead of wrapping, can you just have callHookWithReturnDelta already return the type `BeforeSwapDelta`

src/libraries/Hooks.sol
```diff
@@ -141,6 +143,7 @@ library Hooks {
     {
         bytes memory result = callHook(self, data);
 
+        // If this hook wasnt meant to return something, default to 0 delta
```
snreynolds, line 146: So here do `BeforeSwapDelta.ZERO_DELTA` and below decode it into BeforeSwapDelta type?

src/libraries/Hooks.sol
```diff
@@ -233,21 +236,31 @@ library Hooks {
...
 
-            // Update the swap amount according to the hook's return, and check that the swap type doesnt change (exact input/output)
-            if (hookDeltaSpecified != 0) {
-                bool exactInput = amountToSwap < 0;
-                amountToSwap += hookDeltaSpecified;
-                if (exactInput ? amountToSwap > 0 : amountToSwap < 0) revert HookDeltaExceedsSwapAmount();
+            // skip this logic for the case where the hook return is 0
+            if (canReturnDelta) {
+                // any return in unspecified is passed to the afterSwap hook for handling
+                int128 hookDeltaSpecified = hookReturn.getSpecifiedDelta();
```
snreynolds, line 256: Can this logic be moved into `callHookWithReturnDelta` cause we check canReturnDelta twice now?

src/libraries/Hooks.sol
hensha256: If your fee is a % of the input (or output) then yes I think youll need to implement both hooks. However you could save double-calculating things by saving intermediary results in beforeSwap in transient storage to then reference in the afterSwap hook :)

src/libraries/Hooks.sol
```diff
@@ -259,27 +272,30 @@ library Hooks {
...
         if (self.hasPermission(AFTER_SWAP_FLAG)) {
-            hookDeltaUnspecified = self.callHookWithReturnDelta(
+            hookDeltaUnspecified += self.callHookWithReturnDelta(
                 abi.encodeWithSelector(IHooks.afterSwap.selector, msg.sender, key, params, swapDelta, hookData),
                 self.hasPermission(AFTER_SWAP_RETURNS_DELTA_FLAG)
             ).toInt128();
         }
 
+        BalanceDelta hookDelta;
         if (hookDeltaUnspecified != 0 || hookDeltaSpecified != 0) {
```
snreynolds, line 290: Just making sure that in the case hookDeltaUnspecified is nonzero and then on line 283 they return something to make it 0 again thats OK. Ie we don't actually use hookDeltaUnspecified in any way before or during the internal swap call?

src/test/CustomCurveHook.sol
```diff
@@ -36,26 +37,17 @@ contract CustomCurveHook is BaseTestHooks {
...
         // this "custom curve" is a line, 1-1
         // take the full input amount, and give the full output amount
         manager.take(inputCurrency, address(this), amount);
         outputCurrency.settle(manager, address(this), amount, false);
 
-        // return -amountSpecified to no-op the concentrated liquidity swap
-        return (IHooks.beforeSwap.selector, int128(-params.amountSpecified));
-    }
-
-    function afterSwap(
```
snreynolds, line 51: sick!

src/test/DeltaReturningHook.sol
```diff
@@ -36,20 +38,28 @@ contract DeltaReturningHook is BaseTestHooks {
...
         IPoolManager.SwapParams calldata params,
         bytes calldata /* hookData **/
-    ) external override onlyPoolManager returns (bytes4, int128) {
-        (Currency specifiedCurrency,) = _sortCurrencies(key, params);
-        _settleOrTake(specifiedCurrency, deltaSpecified);
+    ) external override onlyPoolManager returns (bytes4, BeforeSwapDelta) {
+        (Currency specifiedCurrency, Currency unspecifiedCurrency) = _sortCurrencies(key, params);
+
+        if (deltaSpecified != 0) _settleOrTake(specifiedCurrency, deltaSpecified);
+        if (deltaUnspecifiedBeforeSwap != 0) _settleOrTake(unspecifiedCurrency, deltaUnspecifiedBeforeSwap);
```
snreynolds, line 58: Would it be more optimal to settle deltaUnspecified always afterSwap if deltaUnspecifiedAfterSwap > 0 ? Otherwise you have to settle/take twice and they may net.

src/test/DynamicFeesTestHook.sol
```diff
@@ -30,10 +31,10 @@ contract DynamicFeesTestHook is BaseTestHooks {
     function beforeSwap(address, PoolKey calldata key, IPoolManager.SwapParams calldata, bytes calldata)
         external
         override
-        returns (bytes4, int128)
+        returns (bytes4, BeforeSwapDelta)
     {
         manager.updateDynamicLPFee(key, fee);
-        return (IHooks.beforeSwap.selector, 0);
+        return (IHooks.beforeSwap.selector, BeforeSwapDelta.wrap(0));
```
snreynolds, line 37: use constant

snreynolds: & i think theres a few more places that could just use BeforeSwapDelta.ZERO_DELTA

src/types/BeforeSwapDelta.sol
```diff
@@ -0,0 +1,40 @@
...
+}
+
+library BeforeSwapDeltaLibrary {
+    BeforeSwapDelta public constant ZERO_DELTA = BeforeSwapDelta.wrap(0);
+
+    /// extracts int128 from the upper 128 bits of the BeforeSwapDelta
+    /// returned by beforeSwap
+    function getSpecifiedDelta(BeforeSwapDelta delta) internal pure returns (int128 deltaSpecified) {
+        assembly {
+            deltaSpecified := shr(128, delta)
```
snreynolds, line 28: `sar` ? Referencing this fix here I think for signed ints we want to shift right with sar https://github.com/Uniswap/v4-core/pull/586/files#diff-96131175180a7d11a6b3d18a431973bc72a649426f8ecf735f9e7ff973412c5a

snreynolds: Maybe we need to fix this also in BalanceDelta ?

test/PoolManager.t.sol
```diff
@@ -876,12 +876,7 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
     }
 
     function test_swap_beforeSwapNoOpsSwap_exactInput() public {
-        address hookAddr = address(
-            uint160(
-                Hooks.AFTER_SWAP_FLAG | Hooks.AFTER_SWAP_RETURNS_DELTA_FLAG | Hooks.BEFORE_SWAP_FLAG
-                    | Hooks.BEFORE_SWAP_RETURNS_DELTA_FLAG
-            )
-        );
+        address hookAddr = address(uint160(Hooks.BEFORE_SWAP_FLAG | Hooks.BEFORE_SWAP_RETURNS_DELTA_FLAG));
```
snreynolds, line 879: Weeee I love this!!!

src/libraries/Hooks.sol
hensha256: yeah nice call out, but hookDeltaUnspecified is only used after the afterswap hook so if they cancel out that was just a waste of gas

src/types/BeforeSwapDelta.sol
hensha256: good point. BalanceDelta was updated by @shuhuiluo in #641 so have used the same logic as that lib

src/test/DeltaReturningHook.sol
hensha256: Currently kept this hook like this so that our tests can just enable 1 hook or the other for flexibility

src/libraries/Hooks.sol
hensha256: Cannot do this as `callHookWithReturnDelta` is shared by different hooks

src/libraries/Hooks.sol
hensha256: Cannot do this as callHookWithReturnDelta is shared by different hooks

src/libraries/Hooks.sol
hensha256: Cannot do this as callHookWithReturnDelta is shared by different hooks

src/libraries/Hooks.sol
snreynolds: ah I see yes

src/libraries/Hooks.sol
snreynolds: hmm yeah, i tried playing around so we're not double checking canReturnDelta twice but nothing super clean :/

snreynolds (approved):
i think just fix linting and good to go!

snreynolds (approved):

PR #645, opened 2024-05-13
Accurate swap gas

hensha256:


test/PoolManager.t.sol
```diff
@@ -642,21 +642,15 @@ contract PoolManagerTest is Test, Deployers, GasSnapshot {
         IPoolManager.SwapParams memory swapParams =
             IPoolManager.SwapParams({zeroForOne: true, amountSpecified: -100, sqrtPriceLimitX96: SQRT_PRICE_1_2});
 
-        PoolSwapTest.TestSettings memory testSettings =
-            PoolSwapTest.TestSettings({takeClaims: false, settleUsingBurn: false});
-
-        swapRouter.swap(key, swapParams, testSettings, ZERO_BYTES);
+        swapRouterNoChecks.swap(key, swapParams);
```
snreynolds, line 645: Can you keep these same tests with the checks as well, and just add two that are only for gas? Don't want to lose testing coverage

gretzke (dismissed):

snreynolds (approved):

PR #647, opened 2024-05-13
Sparse extsload

gretzke:
## Related Issue

## Description of changes


src/Extsload.sol
```diff
@@ -25,4 +25,28 @@ abstract contract Extsload is IExtsload {
...
+            let len := shl(5, slots.length)
+            let srcEndOffset := add(slots.offset, len)
+            // Iterate through the calldata array.
+            for { let srcOffset := slots.offset } lt(srcOffset, srcEndOffset) { srcOffset := add(srcOffset, 32) } {
+                // Compute memory offset using the computed relative offset and store loaded value.
+                mstore(add(srcOffset, relativeOffset), sload(calldataload(srcOffset)))
+            }
+            // Directly returning avoids Solidity doing it which would trigger a re-encode, almost
+            // doubling the total cost of the method.
+            return(free, add(len, 0x40))
```
hensha256, line 49: would love consistency of hex vs decimal

src/Extsload.sol
```diff
@@ -25,4 +25,28 @@ abstract contract Extsload is IExtsload {
...
+    function extsload(bytes32[] calldata slots) external view returns (bytes32[] memory) {
+        /// @solidity memory-safe-assembly
+        assembly {
+            let free := mload(64)
+            // The offset of our array in the returndata is 32.
+            mstore(free, 32)
+            mstore(add(free, 32), slots.length)
+            // Return values will start at `free + 64`. We compute the difference between the memory
+            // and calldata offset so we don't have to track and update a separate memory offset.
+            let relativeOffset := sub(add(free, 64), slots.offset)
```
hensha256, line 38: how do you know `free + 64` is definitely bigger than slots.offset

gretzke: `slots.offset` will always be 68, 4 bytes for the function signature, 1 word for the offset, 1 word for the length

shuhuiluo: Since this function is external and everything is in assembly, the Solidity memory convention can be discarded, meaning we can use memory starting from 0 and override `0x40`. An implementation was proposed in the following comment.
https://github.com/Uniswap/v4-core/pull/559#pullrequestreview-1982030657

shuhuiluo: I appreciate the recognition. But I didn't mean to override your impl with mine. The comments you wrote are very useful and the way you used the relative offset instead of incrementing two variables should be more efficient. I was suggesting changing the `free` variable to a hardcode `0`. Thank you.

snreynolds (dismissed):
lgtm

snreynolds:
> lgtm

altho i hate that syntax for infinite for loop :P lol

src/Extsload.sol
```diff
@@ -25,4 +25,23 @@ abstract contract Extsload is IExtsload {
 
         return value;
     }
+
+    /// @dev since the function is external and enters a new call context and exits right after execution, Solidity's memeory management convention can be disregarded and a direct slice of memory can be returned
```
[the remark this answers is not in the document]
snreynolds: srry one small nit

test/Extsload.t.sol
```diff
@@ -0,0 +1,26 @@
...
+    Loadable loadable = new Loadable();
+
+    function test_load10_sparse() public {
+        bytes32[] memory keys = new bytes32[](10);
+        for (uint256 i = 0; i < keys.length; i++) {
+            keys[i] = keccak256(abi.encode(i));
+            vm.store(address(loadable), keys[i], bytes32(i));
+        }
+
+        bytes32[] memory values = loadable.extsload(keys);
```
hensha256, line 20: can you add a line after this
```
snapLastCall('sparse external sload');
```

youll need to import/inherit `GasSnapshot`

src/Extsload.sol
hensha256: @shuhuiluo i did a gas comparison of those 2 algorithms and this one is cheaper. For loading 10 sparse slots (the test in `Extsload.t.sol`), the implementation is 2023, and the old one but using `0` instead of `fmp` is 2145 :)

shuhuiluo: @hensha256 I took the best of both worlds. Could you try this? I got 1958.
```solidity
        assembly {
            // The abi offset of dynamic array in the returndata is 32.
            mstore(0, 0x20)
            mstore(0x20, slots.length)
            // A left bit-shift of 5 is equivalent to multiplying by 32 but costs less gas.
            let end := add(0x40, shl(5, slots.length))
            // Return values will start at 64 while calldata offset is 68.
            for { let memptr := 0x40 } 1 {} {
                // Compute calldata offset using the memory offset and store loaded value.
                mstore(memptr, sload(calldataload(add(memptr, 0x04))))
                memptr := add(memptr, 0x20)
                if iszero(lt(memptr, end)) { break }
            }
            // The end offset is also the length of the returndata.
            return(0, end)
        }
```

PR #648, opened 2024-05-14
beforeSwap returns dynamic fee

### LINKED ISSUE #554: Return swapFee in beforeSwap for hooks with relevant permission

saucepoint:
## Related Issue
Closes #554

- [x] define the sentinel _somewhere_ (LPFeeLibrary?)

## Description of changes

---


.forge-snapshots/swap with return dynamic fee.snap
```diff
@@ -0,0 +1 @@
+157979
```
saucepoint, line 1: 2,261 gas less than the previous implementation (no return fee, and have beforeSwap call `updateDynamicLPFee`)

saucepoint:
there's an alternative implementation where `key.fee.isDynamic()` is **not** passed to `callHookWithReturnDeltaAndFee` (to avoid writing `lpFee = type(uint24).max` _twice_)

it's not as clean, but saves 9 gas lul

https://github.com/Uniswap/v4-core/compare/before-return-swap-fee...temp-before-return-swap-fee?expand=1

src/libraries/Hooks.sol
```diff
@@ -148,6 +148,25 @@ library Hooks {
...
+    /// @return delta The delta returned by the hook
+    /// @return fee The fee returned by the hook
+    function callHookWithReturnDeltaAndFee(IHooks self, bytes memory data, bool parseReturn, bool parseFee)
+        internal
+        returns (int256 delta, uint24 fee)
+    {
+        bytes memory result = callHook(self, data);
+        (, delta, fee) = abi.decode(result, (bytes4, int256, uint24));
+
+        if (!parseReturn) {
```
snreynolds, line 161: I think now that there is more than 1 thing that is returned and needs to be parsed we can change this var name to parseDelta

src/libraries/Hooks.sol
```diff
@@ -236,20 +255,23 @@ library Hooks {
     /// @notice calls beforeSwap hook if permissioned and validates return value
     function beforeSwap(IHooks self, PoolKey memory key, IPoolManager.SwapParams memory params, bytes calldata hookData)
         internal
-        returns (int256 amountToSwap, BeforeSwapDelta hookReturn)
+        returns (int256 amountToSwap, BeforeSwapDelta hookReturn, uint24 lpFee)
     {
         amountToSwap = params.amountSpecified;
-        if (msg.sender == address(self)) return (amountToSwap, BeforeSwapDeltaLibrary.ZERO_DELTA);
+        lpFee = type(uint24).max;
+        if (msg.sender == address(self)) return (amountToSwap, BeforeSwapDeltaLibrary.ZERO_DELTA, type(uint24).max);
```
snreynolds, line 262:
```suggestion
        if (msg.sender == address(self)) return (amountToSwap, BeforeSwapDeltaLibrary.ZERO_DELTA, lpFee);
```

.forge-snapshots/swap with return dynamic fee.snap
snreynolds: Im seeing 2,818 difference.. comparing this number to the snapshot below

snreynolds (commented):
Im a bit torn on this one - it definitely makes the hook callsites more messy, and I don't actually love exposing multiple ways of doing the same thing but it is a significant gas difference if a hook does want to change the fee on every swap...

snreynolds:
> there's an alternative implementation where `key.fee.isDynamic()` is **not** passed to `callHookWithReturnDeltaAndFee` (to avoid writing `lpFee = type(uint24).max` _twice_)
> 
> it's not as clean, but saves 9 gas lul
> 
> https://github.com/Uniswap/v4-core/compare/before-return-swap-fee...temp-before-return-swap-fee?expand=1

I actually dont mind this impl now that we still have to check canReturnDelta after the callsite... if that makes sense. Like this way was cleaner before Alice's PR

hensha256 (commented):
Also torn... its making some other non-standard cases more expensive too sadly

.forge-snapshots/swap with return dynamic fee.snap
hensha256: no i agree with sauce - you have to take the _old_ snapshot of the snapshot below, not the new one

.forge-snapshots/swap with lp fee and protocol fee.snap
```diff
@@ -1 +1 @@
-181949
\ No newline at end of file
+182500
```
hensha256, line 1: this one is calling `updateDynamicLPFee` in the beforeSwap hook

hensha256: (just commenting for my own sanity for the ones that have gone up more)

.forge-snapshots/swap with return dynamic fee.snap
snreynolds: oh sorry yes that makes sense haha

saucepoint:
> I actually dont mind this impl now that we still have to check canReturnDelta after the callsite... if that makes sense. Like the way I suggested was cleaner before Alice's PR

went ahead and made the change because 9 gas 😎

.forge-snapshots/swap CA custom curve + swap noop.snap
```diff
@@ -1 +1 @@
-135892
\ No newline at end of file
+135805
```
hensha256, line 1: HAHA the dream

src/interfaces/IHooks.sol
```diff
@@ -99,12 +99,13 @@ interface IHooks {
     /// @param hookData Arbitrary data handed into the PoolManager by the swapper to be be passed on to the hook
     /// @return bytes4 The function selector for the hook
     /// @return BeforeSwapDelta The hook's delta in specified and unspecified currencies. Positive: the hook is owed/took currency, negative: the hook owes/sent currency
+    /// @return uint24 An optional swap fee, only used if the Pool has a dynamic fee and the value is less than of equal to LPFeeLibrary.MAX_LP_FEE
```
hensha256, line 102:
```suggestion
    /// @return uint24 An optional swap fee, only used if the Pool has a dynamic fee and the value is less than or equal to LPFeeLibrary.MAX_LP_FEE
```

src/libraries/Pool.sol
```diff
@@ -317,8 +318,8 @@ library Pool {
         state.feeGrowthGlobalX128 = zeroForOne ? self.feeGrowthGlobal0X128 : self.feeGrowthGlobal1X128;
         state.liquidity = cache.liquidityStart;
 
-        swapFee =
-            cache.protocolFee == 0 ? slot0Start.lpFee : uint24(cache.protocolFee).calculateSwapFee(slot0Start.lpFee);
+        uint24 _lpFee = params.fee <= LPFeeLibrary.MAX_LP_FEE ? params.fee : slot0Start.lpFee;
```
hensha256, line 321: is the _ needed?

src/libraries/Pool.sol
```diff
@@ -317,8 +318,8 @@ library Pool {
         state.feeGrowthGlobalX128 = zeroForOne ? self.feeGrowthGlobal0X128 : self.feeGrowthGlobal1X128;
         state.liquidity = cache.liquidityStart;
 
-        swapFee =
-            cache.protocolFee == 0 ? slot0Start.lpFee : uint24(cache.protocolFee).calculateSwapFee(slot0Start.lpFee);
+        uint24 lpFee = params.fee <= LPFeeLibrary.MAX_LP_FEE ? params.fee : slot0Start.lpFee;
```
hensha256, line 321: maybe 1 comment above this just saying 
```
// if the beforeSwap hook returned a valid fee, use that as the LP fee, otherwise load from storage
```
or something?

`Pool.sol` feels quite removed from the hook interaction

hensha256: this line makes me want to do the following:
update `LPFeeLib.validate` to be `isValid returns bool` then other places that call it can check `if(!isValid) revert` and this place can check `if (isValid) params.fee : slot0.lpFee`

this actually follows the model of `ProtocolFeeLibrary` too - currently they behave differently which is weird. Then we should also rename to `ProtocolFeeLibrary.isValid` too hahaali

src/PoolManager.sol
```diff
@@ -271,19 +271,25 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
...
-            Pool.SwapParams({
-                tickSpacing: key.tickSpacing,
-                zeroForOne: params.zeroForOne,
-                amountSpecified: amountToSwap,
-                sqrtPriceLimitX96: params.sqrtPriceLimitX96
-            }),
-            params.zeroForOne ? key.currency0 : key.currency1 // input token
-        );
+        BeforeSwapDelta beforeSwapDelta;
+        {
```
hensha256, line 275: 😭

src/libraries/Hooks.sol
```diff
@@ -125,14 +127,8 @@ library Hooks {
...
-        bytes4 expectedSelector;
-        bytes4 selector;
-        assembly {
-            expectedSelector := mload(add(data, 0x20))
-            selector := mload(add(result, 0x20))
-        }
-
-        if (selector != expectedSelector) revert InvalidHookResponse();
+        // Check expected selector and returned selector match.
+        if (result.parseSelector() != data.parseSelector()) revert InvalidHookResponse();
```
hensha256, line 131: this library makes me so happy @snreynolds 👑

src/libraries/LPFeeLibrary.sol
```diff
@@ -14,6 +14,7 @@ library LPFeeLibrary {
 
     // the lp fee is represented in hundredths of a bip, so the max is 100%
     uint24 public constant MAX_LP_FEE = 1000000;
+    uint24 public constant MAX_UINT24 = type(uint24).max;
```
hensha256, line 17: maybe a name like `NO_FEE_CHANGE_FLAG` or something would be clearer? or `NO_FEE_UPDATE_FLAG`? not sure

test/DynamicReturnFees.t.sol
```diff
@@ -0,0 +1,141 @@
...
+            currency0,
+            currency1,
+            IHooks(address(dynamicReturnFeesHook)),
+            LPFeeLibrary.DYNAMIC_FEE_FLAG,
+            SQRT_PRICE_1_1,
+            ZERO_BYTES
+        );
+    }
+
+    function test_dynamicReturnSwapFee(uint24 fee) public {
```
hensha256, line 59: `test_fuzz_...`

test/DynamicReturnFees.t.sol
```diff
@@ -0,0 +1,141 @@
...
+        dynamicReturnFeesHook.setFee(fee);
+
+        int256 amountSpecified = -10000;
+        BalanceDelta result = swap(key, true, amountSpecified, ZERO_BYTES);
+
+        // after swapping ~1:1, the amount out (amount1) should be approximately 0.30% less than the amount specified
+        assertEq(result.amount0(), amountSpecified);
+
+        if (fee > LPFeeLibrary.MAX_LP_FEE) {
+            // if the fee is too large, the fee is not used
```
hensha256, line 69: pls can you add that the fallback fee is 0? thats what initialize sets it to

test/DynamicReturnFees.t.sol
```diff
@@ -0,0 +1,141 @@
...
+        );
+    }
+
+    function test_dynamicReturnSwapFee(uint24 fee) public {
+        dynamicReturnFeesHook.setFee(fee);
+
+        int256 amountSpecified = -10000;
+        BalanceDelta result = swap(key, true, amountSpecified, ZERO_BYTES);
+
+        // after swapping ~1:1, the amount out (amount1) should be approximately 0.30% less than the amount specified
```
hensha256, line 65: is this comment right? i think its a copy paste from another test

hensha256 (commented):
Few small things!

src/libraries/LPFeeLibrary.sol
saucepoint: yep i like NO_FEE_UPDATE_FLAG

src/libraries/Pool.sol
saucepoint: yeah i like the suggestion -- updated for both fee libraries

src/libraries/ProtocolFeeLibrary.sol
```diff
@@ -18,12 +18,16 @@ library ProtocolFeeLibrary {
...
+        return self <= MAX_PROTOCOL_FEE;
+    }
+
     function validate(uint24 self) internal pure returns (bool) {
         if (self != 0) {
             uint16 fee0 = getZeroForOneFee(self);
             uint16 fee1 = getOneForZeroFee(self);
             // The fee is represented in pips and it cannot be greater than the MAX_PROTOCOL_FEE.
-            if ((fee0 > MAX_PROTOCOL_FEE) || (fee1 > MAX_PROTOCOL_FEE)) {
+            if ((!isValid(fee0)) || (!isValid(fee1))) {
```
hensha256, line 30: nit: add `using ProtocolFeeLibrary for uint24` and write `fee.isValid()`

src/libraries/Hooks.sol
```diff
@@ -236,22 +232,22 @@ library Hooks {
     /// @notice calls beforeSwap hook if permissioned and validates return value
     function beforeSwap(IHooks self, PoolKey memory key, IPoolManager.SwapParams memory params, bytes calldata hookData)
         internal
-        returns (int256 amountToSwap, BeforeSwapDelta hookReturn)
+        returns (int256 amountToSwap, BeforeSwapDelta hookReturn, uint24 lpFee)
     {
         amountToSwap = params.amountSpecified;
-        if (msg.sender == address(self)) return (amountToSwap, BeforeSwapDeltaLibrary.ZERO_DELTA);
+        lpFee = LPFeeLibrary.NO_FEE_UPDATE_FLAG;
```
snreynolds, line 238: maybe no fee override is a more appropriate name? cause it doenst actually update the cache

snreynolds (commented):
I think its good, the one thing I don't love is that hooks default to return 0 if they dont need to use that fee field EXCEPT if you're a dynamic fee hook and you dont want to override - you shouldn't return 0, you need to return the flag. Feels footgunny but don't have a great workaround

src/libraries/ProtocolFeeLibrary.sol
saucepoint: so i went ahead and renamed the original function to `isValidProtocolFee`, because the function behaves like `isValid` (returns bool) while `validate` is for reverting

Would've aligned on making both ProtocolFeeLibrary.isValid and LPFeeLibrary.isValid, but there's a file that uses both and there was a naming collision

src/libraries/ParseBytes.sol
```diff
@@ -0,0 +1,30 @@
+import {LPFeeLibrary} from "./LPFeeLibrary.sol";
```
hensha256, line 1: oh this can be removed now

src/libraries/Pool.sol
```diff
@@ -317,8 +319,16 @@ library Pool {
...
 
-        swapFee =
-            cache.protocolFee == 0 ? slot0Start.lpFee : uint24(cache.protocolFee).calculateSwapFee(slot0Start.lpFee);
+        // if the beforeSwap hook returned a valid fee override, use that as the LP fee, otherwise load from storage
+        uint24 lpFee;
+        if (!params.lpFeeOverride.isOverride()) {
+            lpFee = slot0Start.lpFee;
+        } else {
+            lpFee = params.lpFeeOverride.removeOverrideFlag();
+            if (!lpFee.isValid()) lpFee = slot0Start.lpFee;
```
hensha256, line 328: do you think we shoulda ctually just revert here? like should this be `lpFee.validate()` and if the hook returned an incorrect LP fee then thats an error and should revert?

hensha256: do you have a test for this case? where the hook returns a value >max?

hensha256: found it 😂

src/libraries/LPFeeLibrary.sol
```diff
@@ -9,8 +9,9 @@ library LPFeeLibrary {
     /// @notice Thrown when the static or dynamic fee on a pool exceeds 100%.
     error FeeTooLarge();
 
-    uint24 public constant STATIC_FEE_MASK = 0x7FFFFF;
+    uint24 public constant FEE_MASK = 0x7FFFFF;
     uint24 public constant DYNAMIC_FEE_FLAG = 0x800000;
+    uint24 public constant BEFORE_SWAP_FEE_OVERRIDE_FLAG = 0x800000;
```
snreynolds, line 14: If we are going to use the same flag - we can just name this `FEE_FLAG` and have it defined once

src/libraries/LPFeeLibrary.sol
```diff
@@ -19,14 +20,33 @@ library LPFeeLibrary {
...
     function getInitialLPFee(uint24 self) internal pure returns (uint24 lpFee) {
         // the initial fee for a dynamic fee pool is 0
         if (self.isDynamicFee()) return 0;
-        lpFee = self & STATIC_FEE_MASK;
+        lpFee = self & FEE_MASK;
         lpFee.validate();
     }
+
+    /// @notice converts a fee to an override fee (to be returned by beforeSwap) by setting the top bit of the uint24
+    function asOverrideFee(uint24 self) internal pure returns (uint24) {
```
snreynolds, line 39: I dont think we should keep this in this library bc core doesnt actually use it

src/libraries/LPFeeLibrary.sol
hensha256: yeeah i did think that - just wasnt sure if the double naming makes the code more readable in other places

src/libraries/LPFeeLibrary.sol
saucepoint: its used in our test hook 🤔 

another periphery vs core debate 🤔 

```solidity
    function beforeSwap(address, PoolKey calldata, IPoolManager.SwapParams calldata, bytes calldata)
        external
        override
        returns (bytes4, BeforeSwapDelta, uint24)
    {
        return (IHooks.beforeSwap.selector, BeforeSwapDeltaLibrary.ZERO_DELTA, fee.asOverrideFee());
    }
```

src/libraries/LPFeeLibrary.sol
saucepoint: hmm, i think if we want to use one, `DYNAMIC_FEE_FLAG` works

used for the poolkey and for fee-overrides, which makes sense to me

saucepoint: consolidating to one constant, we might have 2 redundant functions:

```solidity
    function isDynamicFee(uint24 self) internal pure returns (bool) {
        return self & DYNAMIC_FEE_FLAG != 0;
    }
    
    /// @notice returns true if the fee has the override flag set (top bit of the uint24)
    function isOverride(uint24 self) internal pure returns (bool) {
        return self & DYNAMIC_FEE_FLAG != 0;
    }
 ```
 
 kind of like having `isOverride`, feels a lot clearer for
 
 ```solidity
         // if the beforeSwap hook returned a valid fee override, use that as the LP fee, otherwise load from storage
        uint24 lpFee =
            params.lpFeeOverride.isOverride() ? params.lpFeeOverride.removeOverrideAndValidate() : slot0Start.lpFee;
 ```

File: src/libraries/Pool.sol [hunk 21]
```diff
@@ -320,13 +320,8 @@ library Pool {
...
         // if the beforeSwap hook returned a valid fee override, use that as the LP fee, otherwise load from storage
-        uint24 lpFee;
-        if (!params.lpFeeOverride.isOverride()) {
-            lpFee = slot0Start.lpFee;
-        } else {
-            lpFee = params.lpFeeOverride.removeOverrideFlag();
-            if (!lpFee.isValid()) lpFee = slot0Start.lpFee;
-        }
+        uint24 lpFee =
+            params.lpFeeOverride.isOverride() ? params.lpFeeOverride.removeOverrideAndValidate() : slot0Start.lpFee;
```
hensha256, line 324: sooo much better ❤️

hensha256 (dismissed):
🚀

src/libraries/LPFeeLibrary.sol
```diff
@@ -9,24 +9,51 @@ library LPFeeLibrary {
     /// @notice Thrown when the static or dynamic fee on a pool exceeds 100%.
     error FeeTooLarge();
 
-    uint24 public constant STATIC_FEE_MASK = 0x7FFFFF;
+    uint24 public constant FEE_MASK = 0x7FFFFF;
+    uint24 public constant OVERRIDE_MASK = 0xBFFFFF;
+
+    // the top bit of the fee is used to signal if a Pool's is dynamic
```
hensha256, line 15:
```suggestion
    // the top bit of the fee is used to signal if a Pool's lp fee is dynamic
```

src/libraries/LPFeeLibrary.sol
```diff
@@ -9,24 +9,51 @@ library LPFeeLibrary {
...
     uint24 public constant DYNAMIC_FEE_FLAG = 0x800000;
 
+    // the second bit of the fee (returned by beforeSwap) is used to signal if the stored fee should be overridden
```
hensha256, line 18:
```suggestion
    // the second bit of the fee returned by beforeSwap is used to signal if the stored LP fee should be overridden in this swap
```

hensha256: maybe then 1 more comment on a line below saying only dynamic-fee pools can return a fee from the beforeSwap hook

hensha256 (approved):
lfg

PR #650, opened 2024-05-14
Optimise before swap return fee

hensha256:
Potential optimisation to #648


.forge-snapshots/swap skips hook call if hook is caller.snap
```diff
@@ -1 +1 @@
-225454
\ No newline at end of file
+225012
```
hensha256, line 1: key gas to look at

.forge-snapshots/swap with lp fee and protocol fee.snap
```diff
@@ -1 +1 @@
-182375
\ No newline at end of file
+181943
```
hensha256, line 1: key gas to look at

.forge-snapshots/update dynamic fee in before swap.snap
```diff
@@ -1 +1 @@
-160612
\ No newline at end of file
+160180
```
hensha256, line 1: key gas to look at

.forge-snapshots/swap with return dynamic fee.snap
```diff
@@ -1 +1 @@
-157794
\ No newline at end of file
+157355
```
hensha256, line 1: key gas to look at

.forge-snapshots/swap CA custom curve + swap noop.snap
```diff
@@ -1 +1 @@
-136229
\ No newline at end of file
+135805
```
hensha256, line 1: https://github.com/Uniswap/v4-core/pull/650/files

hensha256, line 1: key gas to look at

.forge-snapshots/swap with return dynamic fee.snap
saucepoint: 3,257 😮‍💨

saucepoint (approved):
i like it -- was one my main issues with the original impl. felt like the one-off handler `callHookWithReturnDeltaAndFee` wasnt very clean

PR #651, opened 2024-05-14
Optimisation - remove decode

hensha256:
We dont use the bytes4 of the decoding, we only use the int256. This just fetches it directly


hensha256:
closing in favour of #652 🫡

PR #652, opened 2024-05-14
Try shared parse lib

snreynolds:
## Related Issue

## Description of changes


src/libraries/Hooks.sol
```diff
@@ -125,12 +127,8 @@ library Hooks {
...
-        bytes4 expectedSelector;
-        bytes4 selector;
-        assembly {
-            expectedSelector := mload(add(data, 0x20))
-            selector := mload(add(result, 0x20))
-        }
+        bytes4 expectedSelector = data.parseSelector();
+        bytes4 selector = result.parseSelector();
 
         if (selector != expectedSelector) revert InvalidHookResponse();
```
hensha256, line 133: slightly less readable but it removes most of the gas that this change added... can i suggest

```
        // Check expected selector and returned selector match
        if (result.parseSelector() != data.parseSelector()) revert InvalidHookResponse();
```

src/libraries/ParseReturn.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.20;
+
+library ParseReturn {
```
hensha256, line 4: maybe 1 comment here saying that all hooks return
`bytes4`, or `bytes4, 32-byte-delta`, or `bytes4, 32-byte-delta, uint24`

hensha256 (approved):

PR #653, opened 2024-05-15
Optimize `TickBitmap.flipTick` using assembly

shuhuiluo:
## Related Issue

## Description of changes
This pull request is a part of #560 in an effort to ease the review process. It focuses on the `flipTick` function in the `TickBitmap` library.


test/libraries/TickBitmap.t.sol
```diff
@@ -92,8 +92,8 @@ contract TickBitmapTest is Test, GasSnapshot {
     }
 
     function test_flipTick_flippingATickThatResultsInDeletingAWord_gas() public {
-        flipTick(SOLO_INITIALIZED_TICK_IN_WORD);
         snapStart("flipTick_flippingATickThatResultsInDeletingAWord");
+        flipTick(SOLO_INITIALIZED_TICK_IN_WORD);
```
hensha256, line 96: HAHAH omg 🤦‍♀️ thank you

src/libraries/TickBitmap.sol
```diff
@@ -28,11 +28,25 @@ library TickBitmap {
...
     function flipTick(mapping(int16 => uint256) storage self, int24 tick, int24 tickSpacing) internal {
-        unchecked {
-            if (tick % tickSpacing != 0) revert TickMisaligned(tick, tickSpacing); // ensure that the tick is spaced
-            (int16 wordPos, uint8 bitPos) = position(tick / tickSpacing);
-            uint256 mask = 1 << bitPos;
-            self[wordPos] ^= mask;
+        /// @solidity memory-safe-assembly
+        assembly {
+            // ensure that the tick is spaced
+            if smod(tick, tickSpacing) {
```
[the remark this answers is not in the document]
hensha256: i actually think a comment of just the whole old function might be good - so that its apparent what wordPos and bitPos are in this function given it no longer calls `position`

test/libraries/TickBitmap.t.sol
```diff
@@ -167,8 +167,8 @@ contract TickBitmapTest is Test, GasSnapshot {
     }
 
     function test_nextInitializedTickWithinOneWord_lteFalse_onBoundary_gas() public {
-        bitmap.nextInitializedTickWithinOneWord(255, 1, false);
         snapStart("nextInitializedTickWithinOneWord_lteFalse_onBoundary");
+        bitmap.nextInitializedTickWithinOneWord(255, 1, false);
```
hensha256, line 171: i see all the tests run with a tickSpacing of 1... would you be able to add a test with more of a range of tick vs tickSpacing? maybe a fuzz test would be ideal

hensha256 (commented):
Love it! Just a couple of things :)

test/libraries/TickBitmap.t.sol
shuhuiluo: I'm afraid not.

shuhuiluo: Added a fuzz test.

src/libraries/TickBitmap.sol
```diff
@@ -28,11 +28,32 @@ library TickBitmap {
...
+         *     if (tick % tickSpacing != 0) revert TickMisaligned(tick, tickSpacing);
+         *     (int16 wordPos, uint8 bitPos) = position(tick / tickSpacing);
+         *     uint256 mask = 1 << bitPos;
+         *     self[wordPos] ^= mask;
+         */
+        /// @solidity memory-safe-assembly
+        assembly {
+            // ensure that the tick is spaced
+            if smod(tick, tickSpacing) {
+                mstore(0, 0xd4d8f3e6) // selector for TickMisaligned(int24,int24)
```
hensha256, line 42: i see we have no tests that revert for this in TickBitmap.t.sol. Please can you add one?
```
vm.expectRevert(abi.encodeWithSelector(TickBitmap.TickMisaligned.selector, tickLower, tickSpacing));
```

That way we always verify this selector is right too

shuhuiluo: It's included in the fuzz test I added. Do you still want a separate one?
https://github.com/Uniswap/v4-core/pull/653/files#diff-4583698d7e46719feb044f179651f012d50645cd55e43405a959c2dea73ae751R104-R105

test/libraries/TickBitmap.t.sol
```diff
@@ -92,11 +92,24 @@ contract TickBitmapTest is Test, GasSnapshot {
...
     }
 
+    function test_fuzz_flipTick(int24 tick, int24 tickSpacing) public {
+        tickSpacing = int24(bound(tickSpacing, 1, type(int24).max));
+
+        if (tick % tickSpacing != 0) {
+            vm.expectRevert(abi.encodeWithSelector(TickBitmap.TickMisaligned.selector, tick, tickSpacing));
+            bitmap.flipTick(tick, tickSpacing);
+        } else {
+            bool initialized = isInitialized(tick, tickSpacing);
```
hensha256, line 107: i think `initialized` is always going to be false because the bitmap is empty at the beginning. So this test is only testing false->true flipping.
Should be enough just to add another 2 lines where you call flip again and check it

src/libraries/TickBitmap.sol
hensha256: yeah youre right thats fine, idk why the fuzz wasnt loading earlier haha

test/libraries/TickBitmap.t.sol
```diff
@@ -288,9 +301,17 @@ contract TickBitmapTest is Test, GasSnapshot {
         }
     }
 
+    function isInitialized(int24 tick, int24 tickSpacing) internal view returns (bool) {
+        unchecked {
+            int24 compressed = tick / tickSpacing;
+            if (tick < 0 && tick % tickSpacing != 0) compressed--; // round towards negative infinity
```
hensha256, line 307: why do you have to subtract 1? why cant it just be the solidity that was previously in `TickBitmap`?
```
(int16 wordPos, uint8 bitPos) = position(tick / tickSpacing);
```

shuhuiluo: That's the first two lines of `nextInitializedTickWithinOneWord`, which I factored out to a `compress` function in #654.

hensha256: but for `flipTick` theres a requirement that (tick % tickSpacing != 0) otherwise it reverts, so this line feels out of place here? Not sure what I'm missing

test/libraries/TickBitmap.t.sol
shuhuiluo: Added.

test/libraries/TickBitmap.t.sol
shuhuiluo: Do you mean `isInitialized` should return `false` instead of rounding if `tick % tickSpacing != 0`?

hensha256 (approved):

PR #656, opened 2024-05-15
Prevent tick calculation for small swaps

0xVolosnikov:
## Related Issue
A significant proportion of swaps are small swaps that do not even cross the boundaries of one tick. This is especially common for stablecoin pools with tickspacing 1.

At the same time, in almost every swap, the current tick is recalculated at the end. `getTickAtSqrtPrice` is used for this calculation and it is rather costly function. On the other side, we can quite simply determine situations in which tick recalculation is definitely not needed and avoid using this function.

## Description of changes

Added a check to avoid tick recalculation for such small swaps. It should be noted that this check will slightly increase the gas for large swaps. Judging by the gas snapshots, at the moment all swaps in tests in which gas is measured are small swaps.


src/libraries/TickMath.sol
```diff
@@ -39,6 +39,21 @@ library TickMath {
...
+    /// @param tick Tick ​​for which the check is carried out
+    /// @param nextNotCrossedTick The next active tick that is not crossed
+    /// @param zeroForOne The direction of price movement
+    function isSqrtPriceInTick(uint160 sqrtPriceX96, int24 tick, int24 nextNotCrossedTick, bool zeroForOne)
+        internal
+        pure
+        returns (bool)
+    {
+        if (zeroForOne) return nextNotCrossedTick == tick || sqrtPriceX96 >= getSqrtPriceAtTick(tick);
+        return nextNotCrossedTick == tick + 1 || sqrtPriceX96 < getSqrtPriceAtTick(tick + 1);
```
shuhuiluo, line 54:
```suggestion
        if (zeroForOne) return nextNotCrossedTick == tick ? true : sqrtPriceX96 >= getSqrtPriceAtTick(tick);
        unchecked {
            return nextNotCrossedTick == tick + 1 ? true : sqrtPriceX96 < getSqrtPriceAtTick(tick + 1);
        }
```
nit: make the short-circuit more explicit in case the compiler fails to optimize it.

src/libraries/Pool.sol
```diff
@@ -427,7 +427,10 @@ library Pool {
                 }
             } else if (state.sqrtPriceX96 != step.sqrtPriceStartX96) {
                 // recompute unless we're on a lower tick boundary (i.e. already transitioned ticks), and haven't moved
-                state.tick = TickMath.getTickAtSqrtPrice(state.sqrtPriceX96);
+                if (!TickMath.isSqrtPriceInTick(state.sqrtPriceX96, state.tick, step.tickNext, zeroForOne)) {
+                    // recomputing does not occur if we obviously did not change the tick
+                    state.tick = TickMath.getTickAtSqrtPrice(state.sqrtPriceX96);
+                }
```
shuhuiluo, line 433: It's a nice arbitrage between `getSqrtPriceAtTick` and `getTickAtSqrtPrice`. But in order for the protocol team to adopt it, it could use more explanations. Because this part of code is where KyberSwap messed up if I remember correctly.

src/libraries/TickMath.sol
```diff
@@ -39,6 +39,21 @@ library TickMath {
         }
     }
 
+    /// @notice Check if sqrtPriceX96 still in `tick` after price movement
+    /// @dev Used to prevent costly `getTickAtSqrtPrice` calculation for small swaps
+    /// @param sqrtPriceX96 The fixed point Q64.96 value of the sqrt price after price movement
+    /// @param tick Tick ​​for which the check is carried out
+    /// @param nextNotCrossedTick The next active tick that is not crossed
```
shuhuiluo, line 46:
```suggestion
    /// @param tickNext The next active tick that is not crossed
```
I think it should simply be `tickNext` for readability.

src/libraries/Pool.sol
shuhuiluo: It could use formal verification like:

For the case `zeroForOne == true`, `state.sqrtPriceX96 >= step.sqrtPriceNextX96`. So if `state.tick == step.tickNext`, `state.sqrtPriceX96 >= getSqrtPriceAtTick(state.tick)`. Because `zeroForOne == true`, `getTickAtSqrtPrice(state.sqrtPriceX96) <= state.tick`. So if `state.sqrtPriceX96 >= getSqrtPriceAtTick(state.tick)`, `state.tick` remains the same.

Still, the above proof is sloppy.

hensha256:

PR #658, opened 2024-05-16
v3 fuzzer

gretzke:
Compares v3 and v4 outputs


src/test/SwapRouterNoChecks.sol
```diff
@@ -43,5 +43,7 @@ contract SwapRouterNoChecks is PoolTestBase {
             data.key.currency1.settle(manager, data.sender, uint256(int256(-delta.amount1())), false);
             data.key.currency0.take(manager, data.sender, uint256(int256(delta.amount0())), false);
         }
+
+        return "";
```
hensha256, line 47: what happens to gas if instead of this you just name the return param?
like `(bytes memory returnData)` and then leave it unassigned?

hensha256: just intrigued haha

gretzke: +37 gas, I just wanted to get rid of the compiler warning

test/PoolManager.swap.t.sol
```diff
@@ -0,0 +1,114 @@
...
+        // add liquidity
+        v3Pool.mint(
+            address(this),
+            v4LiquidityParams.tickLower,
+            v4LiquidityParams.tickUpper,
+            uint128(int128(v4LiquidityParams.liquidityDelta)),
+            ""
+        );
+
+        modifyLiquidityRouter.modifyLiquidity{value: msg.value}(key_, v4LiquidityParams, "");
```
hensha256, line 68: this function isnt payable so there cant be a msg.value

test/PoolManager.swap.t.sol
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    function uniswapV3SwapCallback(int256 amount0Delta, int256 amount1Delta, bytes calldata) external {
+        if (amount0Delta > 0) currency0.transfer(msg.sender, uint256(amount0Delta));
+        if (amount1Delta > 0) currency1.transfer(msg.sender, uint256(amount1Delta));
+    }
+}
+
+contract V3SwapTests is V3Fuzzer {
+    function test_shouldSwapEqual(int24 lowerTick, int24 upperTick, int128 liquidityDelta, int256 sqrtPriceX96seed)
```
snreynolds, line 105: nit: could we call these params unsanitized or unbounded? when the fuzzer fails on inputs here I thought the test was actually failing with these inputs but we sanitize them later on so its a little misleading

shuhuiluo: As far as I experimented, input args of external functions are abi-decoded from calldata, which are considered sanitized, since abi decoding includes sanity checks. But it's worth constructing dirty calldata and making a low-level call to see if abi decoding reverts.

test/PoolManager.swap.t.sol
```diff
@@ -0,0 +1,176 @@
...
+        console2.log("tickUpper:", tickUpper);
+        vm.assume(liquidityDelta > 0);
+        console2.log("liquidityDelta:", liquidityDelta);
+        if (tick < tickLower) {
+            amount0Delta = SqrtPriceMath.getAmount0Delta(
+                TickMath.getSqrtPriceAtTick(tickLower), TickMath.getSqrtPriceAtTick(tickUpper), liquidityDelta
+            ).toInt128();
+        } else if (tick < tickUpper) {
+            amount0Delta = SqrtPriceMath.getAmount0Delta(
+                TickMath.getSqrtPriceAtTick(tickLower), sqrtPriceX96, liquidityDelta
```
shuhuiluo, line 154:
```suggestion
                sqrtPriceX96, TickMath.getSqrtPriceAtTick(tickUpper), liquidityDelta
```

hensha256 (dismissed):

hensha256 (approved):

PR #661, opened 2024-05-16
remove reddundant line

hensha256:
This line doesnt do anything - because of line 28 `if (self.isDynamicFee()) return 0;` we already know that the first bit is 0.


saucepoint (dismissed):

src/libraries/LPFeeLibrary.sol
```diff
@@ -52,9 +52,8 @@ library LPFeeLibrary {
     }
 
     /// @notice Removes the override flag and validates the fee (reverts if the fee is too large)
-    function removeOverrideAndValidate(uint24 self) internal pure returns (uint24) {
-        uint24 fee = self.removeOverrideFlag();
+    function removeOverrideAndValidate(uint24 self) internal pure returns (uint24 fee) {
+        fee = self.removeOverrideFlag();
         fee.validate();
-        return fee;
```
hensha256, line 58: this removed 4 gas 😆

saucepoint (approved):

PR #273, opened 2023-06-18
Optimize `TickMath`

shuhuiluo:
## Related Issue
Which issue does this pull request resolve?

## Description of Changes

The `TickMath` library has undergone extensive annotation and optimization, with the implementation ported from [Aperture-Finance/uni-v3-lib](https://github.com/Aperture-Finance/uni-v3-lib). The optimizations primarily focus on eliminating branching code that involves conditional statements. An inlined and more efficient version of the `mostSignificantBit` function from [Solady::LibBit](https://github.com/Vectorized/solady/blob/37b4fd04dd82977eb0161839389187abb7afdb58/src/utils/LibBit.sol) has been incorporated, following its introduction in #257. Additional amendments have been made to reduce the number of opcodes and stack/memory operations.

New Forge tests for `TickMath` have been added to affirm equivalence with the reference implementation and to allow comparisons of gas usage. Gas snapshots were taken both before and after the changes to evaluate the effectiveness of these optimizations. The results are as follows:
## Gas Usage Comparison

| Library  | Test               | Before | After  | Gas Efficiency |
|----------|--------------------|--------|--------|----------------|
| TickMath | getSqrtRatioAtTick | 87496  | 74032  | 15.39%         |
| TickMath | getTickAtSqrtRatio | 233945 | 188818 | 23.90%         |

*Note: Each test runs the corresponding function 100 times with deterministic inputs.*

PR #663, opened 2024-05-17
Optimize `absTick` calculation in `TickMath.getSqrtPriceAtTick`

shuhuiluo:
Part of #273.

## Related Issue

## Description of changes
The getSqrtPriceAtTick function in the TickMath library has been refactored with a more efficient assembly implementation. This optimization significantly reduces the gas usage by using bitwise operations for computing absolute value of a tick.


src/libraries/TickMath.sol
```diff
@@ -46,7 +46,15 @@ library TickMath {
     /// at the given tick
     function getSqrtPriceAtTick(int24 tick) internal pure returns (uint160 sqrtPriceX96) {
         unchecked {
-            uint256 absTick = tick < 0 ? uint256(-int256(tick)) : uint256(int256(tick));
```
hensha256, line 49: could you add 1 test to show that the behaviour for type(int24).min is unchanged - just bc thats the 1 number that performs `-1` weirdly. Something like

```

    function test_getSqrtPriceAtTick_throwsForMinIn24() public {
        vm.expectRevert(TickMath.InvalidTick.selector);
        tickMath.getSqrtPriceAtTick(type(int24).min);
    }
```

hensha256 (dismissed):

hensha256 (approved):

PR #664, opened 2024-05-17
Optimize `price` calculation in `TickMath.getSqrtPriceAtTick`

shuhuiluo:
Part of #273.

## Related Issue

## Description of changes
The way the 'price' variable is calculated within the TickMath library has been significantly refactored. Previously, the variable was set through a ternary conditional statement. Now, a more intricate assembly logic is used, which involves shifting, AND operations, and masking. Apart from the improved readability, this change also aims to optimize the math operations for 'price' calculation.


src/libraries/TickMath.sol
```diff
@@ -49,8 +49,13 @@ library TickMath {
...
             if (absTick > uint256(int256(MAX_TICK))) revert InvalidTick();
 
-            uint256 price =
-                absTick & 0x1 != 0 ? 0xfffcb933bd6fad37aa2d162d1a594001 : 0x100000000000000000000000000000000;
+            // Equivalent to:
+            //     price = absTick & 0x1 != 0 ? 0xfffcb933bd6fad37aa2d162d1a594001 : 0x100000000000000000000000000000000;
+            //     or price = int(2**128 / sqrt(1.0001)) if (absTick & 0x1) else 1 << 128
+            uint256 price;
+            assembly {
+                price := xor(shl(128, 1), mul(xor(shl(128, 1), 0xfffcb933bd6fad37aa2d162d1a594001), and(absTick, 0x1)))
```
hensha256, line 57: does the optimiser deal with cutting down this line? theres lots of things it knows it advance
`xor(shl(128, 1), 0xfffcb933bd6fad37aa2d162d1a594001)`
`shl(128, 1)`

shuhuiluo: Yes. The compiler optimizer is able to reduce literal math expressions.

hensha256 (approved):

PR #667, opened 2024-05-18
Optimize overflow check in `mulDivRoundingUp`

shuhuiluo:
Factored out of #601.

## Related Issue

## Description of changes
This commit optimizes the overflow check within the `mulDivRoundingUp` function of FullMath library. This is achieved by combining two steps: the increment of the `result` and the overflow check, into a single step using the `++` operator. This simplifies the code, improves its readability and brings down the gas usage slightly, as reflected in updated snapshot files.


src/libraries/FullMath.sol
```diff
@@ -109,9 +109,8 @@ library FullMath {
     function mulDivRoundingUp(uint256 a, uint256 b, uint256 denominator) internal pure returns (uint256 result) {
         unchecked {
             result = mulDiv(a, b, denominator);
-            if (mulmod(a, b, denominator) > 0) {
-                require(result < type(uint256).max);
-                result++;
+            if (mulmod(a, b, denominator) != 0) {
+                require(++result > 0);
```
hensha256, line 113: looks good. I see that the test `test_fuzz_mulDivRoundingUp` isn't very good because it checks 
```
assertTrue(result == numerator / d || result == numerator / d + 1);
```
🥲 
please could you update it to correctly identify whether it should have `+1` or not with a mod?

shuhuiluo: Updated.

hensha256 (approved):

PR #669, opened 2024-05-18
Optimize `TickMath` bounds check in assembly

shuhuiluo:
Factored out of #273.

## Related Issue

## Description of changes
A trick to perform bounds check on fixed bounds has been introduced in `TickMath`. A new constant 
```solidity
MAX_SQRT_PRICE_MINUS_MIN_SQRT_PRICE_MINUS_ONE = MAX_SQRT_PRICE - MIN_SQRT_PRICE - 1;
```
was also introduced to facilitate the check. In `getTickAtSqrtPrice`, it is required that `MIN_SQRT_PRICE <= sqrtPriceX96 < MAX_SQRT_PRICE`.

- If `sqrtPriceX96 < MIN_SQRT_PRICE`, `sqrtPriceX96 - MIN_SQRT_PRICE` underflows so `sqrtPriceX96 - MIN_SQRT_PRICE > MAX_SQRT_PRICE_MINUS_MIN_SQRT_PRICE_MINUS_ONE` is `true`.
- If `sqrtPriceX96 >= MAX_SQRT_PRICE`, `sqrtPriceX96 - MIN_SQRT_PRICE > MAX_SQRT_PRICE_MINUS_MIN_SQRT_PRICE_MINUS_ONE` is also `true`.

Therefore, the bounds check reduces to
```solidity
gt(sub(sqrtPriceX96, MIN_SQRT_PRICE), MAX_SQRT_PRICE_MINUS_MIN_SQRT_PRICE_MINUS_ONE)
```


src/libraries/TickMath.sol
```diff
@@ -24,6 +24,9 @@ library TickMath {
     uint160 internal constant MIN_SQRT_PRICE = 4295128739;
     /// @dev The maximum value that can be returned from #getSqrtPriceAtTick. Equivalent to getSqrtPriceAtTick(MAX_TICK)
     uint160 internal constant MAX_SQRT_PRICE = 1461446703485210103287273052203988822378723970342;
+    /// @dev A threshold used for optimized bounds check, equals `MAX_SQRT_PRICE - MIN_SQRT_PRICE - 1`
+    uint160 internal constant MAX_SQRT_PRICE_MINUS_MIN_SQRT_PRICE_MINUS_ONE =
+        1461446703485210103287273052203988822378723970342 - 4295128739 - 1;
```
hensha256, line 29: dont constants get evaluated at runtime not compile time? so cheaper to hardcode the actual value?

shuhuiluo: They should be evaluated at compile time. If it's not, it should be a bug reported to https://github.com/ethereum/solidity.

But I recall your comment about the quirks on `keccak256` constants. I tried replacing it with a hardcoded literal and verified there's no difference in gas.

hensha256: the docs say (though the docs arent always right lol)
> For a constant variable, the expression assigned to it is copied to all the places where it is accessed and also re-evaluated each time. This allows for local optimizations. Immutable variables are evaluated once at construction time and their value is copied to all the places in the code where they are accessed.

Thanks for checking!

hensha256 (approved):

PR #670, opened 2024-05-18
Refactor nested ternary assignment in `TickMath.getTickAtSqrtPrice`

shuhuiluo:
Factored out of #273.

## Related Issue

## Description of changes
The formula to calculate the 'tick' value in the TickMath library has been refactored for potentially better clarity. Instead of using the conditional operator for this calculation, it has been replaced with an equivalent 'if-else' structure which may provide clearer semantic understanding.


hensha256:

PR #258, opened 2023-06-16
Optimize `SqrtPriceMath`

shuhuiluo:
## Related Issue
Which issue does this pull request resolve?

## Description of changes

The functions within `SqrtPriceMath` have been comprehensively rewritten and optimized to various degrees, with implementation ported from [Aperture-Finance/uni-v3-lib](https://github.com/Aperture-Finance/uni-v3-lib). Heuristics for replacing conditional statements with bitwise operations have previously been discussed in https://github.com/Uniswap/universal-router/pull/291. Additional changes primarily revolve around mathematical enhancements.

A new function, `FullMath.mulDivQ96`, is introduced for more efficient execution of `mulDiv(a, b, FixedPoint96.Q96)`. Helper functions such as `sort2` and `absDiff` implemented in inline assembly are also introduced to reduce gas which may later be placed in a separate library.

Gas snapshots have been taken after each commit to incrementally validate the effectiveness of these optimizations. The results are as follows:

## Gas Usage Comparison

| Function                   | Before | After | Gas Savings |
|----------------------------|--------|-------|-------------|
| getNextSqrtPriceFromInput  | 685    | 648.5   | 5.33%      |
| getNextSqrtPriceFromOutput | 673  | 657   | 2.38%      |
| getAmount0Delta            | 590.5    | 515.5   | 12.70%      |
| getAmount1Delta            | 589    | 355.5   | 39.64%      |

*Note: The provided gas numbers are averages calculated from Foundry gas snapshots, taking into account both `roundUp = true` and `roundUp = false` scenarios.* 

In addition to these, `SwapMath.computeSwapStep` has also seen improvements ranging from 9.1% to 15.6%.

PR #675, opened 2024-05-19
Add `UnsafeMath.div` and optimize `SqrtPriceMath.getAmount0Delta`

shuhuiluo:
Factored out of #258.

## Related Issue

## Description of changes
Added a new `div` function in `UnsafeMath` library that skips checking for zero divisor and a function `sort2` for sorting two uint values in `SqrtPriceMath`. Refactored code in assembly to avoid implicit casting. Optimize `SqrtPriceMath.getAmount0Delta` appreciably.

Saves almost **100** in both runtime gas and bytecode size.


src/libraries/SqrtPriceMath.sol
```diff
@@ -35,29 +35,34 @@ library SqrtPriceMath {
...
+            numerator1 := shl(96, liquidity)
+            _sqrtPX96 := sqrtPX96
+        }
 
         if (add) {
             unchecked {
-                uint256 product = amount * sqrtPX96;
-                if (product / amount == sqrtPX96) {
+                uint256 product = amount * _sqrtPX96;
+                if (UnsafeMath.div(product, amount) == _sqrtPX96) {
```
hensha256, line 48: why is it beneficial to call `UnsafeMath.div(product, amount)` given `product / amount` was already inside an unchecked block? whats the benefit?

shuhuiluo: > It is not possible to disable the check for division by zero or modulo by zero using the unchecked block.

https://docs.soliditylang.org/en/latest/control-structures.html#checked-or-unchecked-arithmetic

PR #676, opened 2024-05-19
Optimize `SqrtPriceMath.getAmount1Delta`

shuhuiluo:
Factored out of #258.

## Related Issue

## Description of changes
Added a new function `absDiff` to `SqrtPriceMath` library which calculates the absolute difference between two numbers. Refactored the way price delta is computed between two prices in the `getAmount1Delta` function. The calculation is now more efficient.

Saves as much as **300** in runtime gas and **97** in bytecode size.


src/libraries/SqrtPriceMath.sol
```diff
@@ -169,6 +169,15 @@ library SqrtPriceMath {
         }
     }
 
+    /// @notice Equivalent to: `a > b ? a - b : b - a`
+    function absDiff(uint160 a, uint160 b) internal pure returns (uint256 res) {
+        assembly {
+            let diff := sub(a, b)
+            let mask := sar(255, diff)
```
hensha256, line 176: could you add a similar comment to absTick? just something like
mask is all 1s if b > a or all 0s when a >= b
b-a = -(a-b) = invert (a-b-1)

hensha256 (approved):

PR #682, opened 2024-05-20
import v3-sdk to ffi

dianakocsis:
## Related Issue
importing v3-sdk to ffi

## Description of changes
To Run tests:
`foundryup --pr 7941`
`forge test --no-match-path 'test/js-scripts/**/*'`


test/libraries/FullMath.t.sol
```diff
@@ -205,4 +206,40 @@ contract FullMathTest is Test {
 
         return mulDivResultOverflows || mulDivRoundingUpResultOverflows;
     }
+
+    function test_mulDiv_matchesJavaScriptImpl(uint256 a, uint256 b, uint256 denominator) public {
```
snreynolds, line 210: anything that uses ffi let's put in the test name so we can filter out these tests if we need to when we run locally

`test_ffi_fuzz_...`

test/js-scripts/package.json
```diff
@@ -28,6 +28,12 @@
...
   "scripts": {
     "forge-test-getSqrtPriceAtTick": "npx ts-node src/getSqrtPriceAtTick.ts",
-    "forge-test-getTickAtSqrtPrice": "npx ts-node src/getTickAtSqrtPrice.ts"
+    "forge-test-getTickAtSqrtPrice": "npx ts-node src/getTickAtSqrtPrice.ts",
+    "forge-test-mulDiv": "npx ts-node src/mulDiv.ts",
+    "forge-test-mulDivRoundingUp": "npx ts-node src/mulDivRoundingUp.ts"
+  },
+  "dependencies": {
+    "@uniswap/v3-sdk": "^3.11.2",
+    "jsbi": "^4.3.0"
```
snreynolds, line 37: did you manually install this? Don't think we need to

dianakocsis: oh ty!

snreynolds (commented):
looks good! would just add that in-line config.

& looks like we need to fix github actions

PR #688, opened 2024-05-22
use better bounds for liquidity

snreynolds:
## Related Issue

## Description of changes


.forge-snapshots/removeLiquidity.snap
```diff
@@ -1 +0,0 @@
-120637
```
hensha256, line 1: whys this gone?

src/test/Fuzzers.sol
```diff
@@ -10,35 +10,95 @@ import {BalanceDelta} from "../types/BalanceDelta.sol";
...
+    function boundLiquidityDelta(PoolKey memory key, int256 liquidityDeltaUnbounded, int256 liquidityMaxByAmount)
+        internal
+        pure
+        returns (int256)
+    {
+        int256 liquidityMaxPerTick = int256(uint256(Pool.tickSpacingToMaxLiquidityPerTick(key.tickSpacing)));
+
+        // Finally bound the seeded liquidity by either the max per tick, or by the amount allowed in the position range.
+        int256 liquidityMax = liquidityMaxByAmount > liquidityMaxPerTick ? liquidityMaxPerTick : liquidityMaxByAmount;
+        return bound(liquidityDeltaUnbounded, 0.0000001e18, liquidityMax);
```
hensha256, line 30: whats the reason for bounding by `0.0000001e18`?

src/test/Fuzzers.sol
```diff
@@ -10,35 +10,95 @@ import {BalanceDelta} from "../types/BalanceDelta.sol";
...
+    }
+
+    function boundLiquidityDeltaStrict(PoolKey memory key, int256 liquidityDeltaUnbounded, int256 liquidityMaxByAmount)
```
hensha256, line 33: can you add some comments explaining when one should use `boundLiquidityDeltaStrict` vs `boundLiquidityDelta`

hensha256 (commented):
a few thoughts

src/test/Fuzzers.sol
hensha256: like this one just seems to divide by 2..? i dont understand what makes it "strict"

.forge-snapshots/removeLiquidity.snap
snreynolds: its just that someone didn't take it out when they merged something into main... 

we really should be cleaning snapshots on every merge but people dont delete the folder so they get leftover. 

i run `rm -rf ./forge-snapshots` and then rerun them and so it catches things like this

snreynolds: looks like at some point this was probably renamed to "simple removeLiquidity"

src/test/Fuzzers.sol
snreynolds: This is the old version (ie how we were doing it on all of the StateLibrary tests.) 

We need a "strict" version for bounding when you want to add liquidity at potentially overlapping ticks. The liquidity boundaries that I added are for vanilla one off positions (ie it bounds it at the max for one tick) which will lead to reverts if there is already liquidity at that tick.

The 2 is carried over from before - it's not really exact and maybe could be better for multiple positions but it works for now

src/test/Fuzzers.sol
snreynolds: We just don't want 0 - because thats collecting fees. So we want the smallest amount of liquidity we can add without it being 0

hensha256: and why not just `1`?

src/test/Fuzzers.sol
hensha256: I see I see. I feel like the function names could be changed to reflect this more clearly. Maybe just `boundLiquidityDeltaByMax` and `boundLiquidityDeltaForMultiplePositions` or `boundLiquidityDeltaByHalfMax` or something.

src/test/Fuzzers.sol
```diff
@@ -10,35 +10,97 @@ import {BalanceDelta} from "../types/BalanceDelta.sol";
...
+            TickMath.getSqrtPriceAtTick(tickLower),
+            TickMath.getSqrtPriceAtTick(tickUpper),
+            uint128(type(int128).max)
         );
+
+        // Compare the max amounts (defined by the range of the position) to the max amount constrained by the type container.
+        // The true maximum should be the minimum of the two.
+        // (ie If the position range allows a deposit of more then int128.max in any token, then here we cap it at int128.max.)
+
+        uint256 amount0 = uint256(type(uint128).max / 2);
```
hensha256, line 66: why is this being divided by 2?

hensha256: why not just do type(int128).max?

snreynolds: type(uint128).max / 2 == type(int128).max

src/test/Fuzzers.sol
snreynolds: 1 would work, this code is just taking the most extremes for minimum and more extremes for maximum

src/test/Fuzzers.sol
snreynolds: sure can do that!

src/test/Fuzzers.sol
gretzke: I'd suggest using 1, the tests still pass and we have a broader range of values

gretzke (approved):

PR #689, opened 2024-05-22
Fixing hook address constants

hensha256:
Quick PR to fix the hook address constants. Some tests were doing the wrong thing, constant names didnt make sense etc.


test/PoolManagerInitialize.t.sol
```diff
@@ -163,7 +163,7 @@ contract PoolManagerInitializeTest is Test, Deployers, GasSnapshot {
         // Assumptions tested in Pool.t.sol
         sqrtPriceX96 = uint160(bound(sqrtPriceX96, TickMath.MIN_SQRT_PRICE, TickMath.MAX_SQRT_PRICE - 1));
 
-        address hookEmptyAddr = Constants.EMPTY_HOOKS;
+        address hookEmptyAddr = Constants.ALL_HOOKS;
```
snreynolds, line 166: The old EMPTY_HOOKS address is different than ALL_HOOKS - just checking thats intended here? 
also change var name.. hookEmptyAddress is misleading

.forge-snapshots/swap with hooks.snap
```diff
@@ -1 +1 @@
-123297
\ No newline at end of file
+142010
```
snreynolds, line 1: guessing this is because our all hooks address has less zeros now?

hensha256: no its because this gas test for "swap with hooks" wasnt calling any hooks 🙃

test/PoolManagerInitialize.t.sol
hensha256: ok changed the variable name. I think it was just bad naming tbh but im definitely confused. All the `EMPTY_HOOKS` uses were to create a gas test for `swap with hook` or `add liq with hook` so really "empty" is just wrong lol

.forge-snapshots/swap with hooks.snap
snreynolds: aha and now it is yes?

snreynolds (approved):

PR #690, opened 2024-05-22
add justfile

snreynolds:
## Related Issue

Commands are getting a little long to run tests with --isolate and soon with --no-match-path when we have the v3-sdk dependency. This documents that and allows you to run a few commands:

`just test` : Run all forge tests.

`just prep`: Runs linter.

`just test-mt {regex}`: Runs forge test --mt 

## Description of changes

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,10 +1,10 @@
 
 test *args: (test-forge args)
 build *args: (build-forge args)
-prep *args: fix (test args)
+prep *args: fix (test-forge args)
 
 
-test-forge *args: install-forge build-forge
+test-forge *args: build-forge
     forge test --isolate --no-match-path 'test/js-scripts/**/*' {{args}}
 
 
--- a/justfile
+++ b/justfile
@@ -1,15 +1,15 @@
-test: test-forge
-prep: fix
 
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -0,0 +1,19 @@
+
+test *args: (test-forge args)
+build *args: (build-forge args)
+prep *args: fix (test args)
+
+
+test-forge *args: install-forge build-forge
+    forge test --isolate --no-match-path 'test/js-scripts/**/*' {{args}}
```
saucepoint, line 8: not exactly sure how nested commands work with just, but looked like we were calling `forge install` twice if you can `just test`

```suggestion
prep *args: fix (test-forge args)


test-forge *args: build-forge
    forge test --isolate --no-match-path 'test/js-scripts/**/*' {{args}}
```

saucepoint (approved):
lgtm and works on my end!

optional nit picks below. but not a big deal to me since i think this file wont be touched all that often

justfile
snreynolds: yeah ur prob right good catch

hensha256 (approved):

PR #691, opened 2024-05-22
Protocolfees tests

dianakocsis:
## Related Issue

## Description of changes


src/PoolManager.sol
```diff
@@ -93,12 +93,12 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     /// @inheritdoc IPoolManager
     int24 public constant MIN_TICK_SPACING = TickMath.MIN_TICK_SPACING;
 
-    mapping(PoolId id => Pool.State) internal pools;
+    mapping(PoolId id => Pool.State) internal _pools;
 
     constructor(uint256 controllerGasLimit) ProtocolFees(controllerGasLimit) {}
 
     function _getPool(PoolId id) internal view override returns (Pool.State storage) {
```
hensha256, line 100: why do we have this function 😂 can you remove it? i feel like it should just be `_pools[id]` everywhere?

src/ProtocolFees.sol
```diff
@@ -59,7 +59,7 @@ abstract contract ProtocolFees is IProtocolFees, Owned {
     /// @dev to prevent an invalid protocol fee controller from blocking pools from being initialized
     ///      the success of this function is NOT checked on initialize and if the call fails, the protocol fees are set to 0.
     /// @dev the success of this function must be checked when called in setProtocolFee
-    function _fetchProtocolFee(PoolKey memory key) internal returns (bool success, uint24 protocolFees) {
+    function _fetchProtocolFee(PoolKey memory key) internal returns (bool success, uint24 protocolFee) {
```
hensha256, line 62: so i belieeeeve it was plural because it contains 2 fees, one for each direction. But I do agree with this change it sounds better

src/test/ProtocolFeesImplementation.sol
```diff
@@ -0,0 +1,36 @@
...
+    using PoolIdLibrary for PoolKey;
+
+    mapping(PoolId id => Pool.State) internal _pools;
+
+    constructor(uint256 _controllerGasLimit) ProtocolFees(_controllerGasLimit) {}
+
+    function setPrice(PoolKey memory key, uint160 sqrtPriceX96) public {
+        Pool.State storage pool = _getPool(key.toId());
+        Slot0 newSlot = pool.slot0.setSqrtPriceX96(sqrtPriceX96);
+        pool.slot0 = newSlot;
```
hensha256, line 22: you should just be able to do 
`pool.slot0 = pool.slot0.setSqrtPriceX96(sqrtPriceX96)`

src/PoolManager.sol
dianakocsis: it's because when we moved all the protocol fees logic into ProtocolFees.sol, it needed to have `_getPool(id)` to set the protocol fee on the pool. And since PoolManager inherits from ProtocolFees, it needs to implement that function

src/ProtocolFees.sol
dianakocsis: ohh i see. we could do `_fetchProtocolFees`? they should either be both singular or both plural

src/PoolManager.sol
hensha256: ohhhhh nice, thanks for reminding me

src/test/ProtocolFeesImplementation.sol
```diff
@@ -0,0 +1,36 @@
...
+import {Slot0} from "../types/Slot0.sol";
+
+contract ProtocolFeesImplementation is ProtocolFees {
+    using PoolIdLibrary for PoolKey;
+
+    mapping(PoolId id => Pool.State) internal _pools;
+
+    constructor(uint256 _controllerGasLimit) ProtocolFees(_controllerGasLimit) {}
+
+    function setPrice(PoolKey memory key, uint160 sqrtPriceX96) public {
```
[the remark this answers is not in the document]
hensha256: oops i didnt publish this question last night

test/ProtocolFeesImplementation.t.sol
```diff
@@ -0,0 +1,202 @@
...
+        protocolFees.setProtocolFeeController(feeController);
+        vm.expectRevert(IProtocolFees.InvalidCaller.selector);
+        protocolFees.setProtocolFee(key, 1);
+    }
+
+    function test_setProtocolFee_revertsWithInvalidFee() public {
+        protocolFees.setProtocolFeeController(feeController);
+        vm.prank(address(feeController));
+        vm.expectRevert(IProtocolFees.InvalidProtocolFee.selector);
+        protocolFees.setProtocolFee(key, MAX_PROTOCOL_FEE_BOTH_TOKENS + 1);
```
hensha256, line 81: could you add 2 more lines to check it also reverts for `MAX_PROTOCOL_FEE_BOTH_TOKENS + (1 << 12)`

test/ProtocolFeesImplementation.t.sol
```diff
@@ -0,0 +1,202 @@
...
+        protocolFees.collectProtocolFees(address(this), currency0, 100);
+        assertEq(protocolFees.protocolFeesAccrued(currency0), 0);
+        assertEq(currency0.balanceOf(address(this)), 100);
+    }
+
+    function test_fuzz_collectProtocolFees(address recipient, uint256 amount, uint256 feesAccrued) public {
+        vm.assume(feesAccrued <= currency0.balanceOf(address(protocolFees)));
+        vm.assume(amount <= feesAccrued);
+        vm.assume(recipient != address(protocolFees));
+        vm.assume(recipient != address(0));
```
hensha256, line 119: does it actually revert for the 0 address?

src/test/ProtocolFeesImplementation.sol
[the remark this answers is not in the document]
dianakocsis: it's because in order to successfuly set a protocol fee on a pool, the pool needs to be initialized. and for a pool to be initialized, it needs to have a price set. i can add a comment to make that more clear

hensha256 (approved):

PR #692, opened 2024-05-22
Enable via IR compilation / memory safe issue fix

gretzke:
When compiling via ir compilation fails due to memory safeness issues, unless every assembly block from this PR are annotated as `memory-safe`. Once compilation succeeds two tests fail with a panic error (over- or underflow) that do not fail when compiling without ir.


test/libraries/Hooks.t.sol
```diff
@@ -943,12 +943,6 @@ contract HooksTest is Test, Deployers, GasSnapshot {
         );
     }
 
-    function test_gas_hookShouldCallBeforeSwap() public {
-        snapStart("HooksShouldCallBeforeSwap");
```
snreynolds, line 947: why are these gas tests being removed? 

and can you clear them out of .forge-snapshots/ if its necessary they are removed?

gretzke: they were cleared from ./forge-snapshots already, these tests are removed because they cause an underflow because via ir in the forge gas snapshot library with the gas calibration variable. These functions only measure gas, so there is no real test coverage lost in removing them

snreynolds (dismissed):
just update main!

.forge-snapshots/simple swap.snap
```diff
@@ -1 +1 @@
-130974
\ No newline at end of file
+124969
```
hensha256, line 1: SOOO GOOOOOOOOOOD

gretzke: 🎉🎉🎉

foundry.toml
```diff
@@ -1,8 +1,8 @@
 [profile.default]
-optimizer_runs = 800
-via_ir = false
+optimizer_runs = 44444444
```
hensha256, line 2: hahahaha nice

.forge-snapshots/getReserves.snap
```diff
@@ -1 +1 @@
-1167
\ No newline at end of file
+1630
```
hensha256, line 1: woah... interesting

hensha256 (approved):

PR #695, opened 2024-05-23
Revert custom errors with new `CustomRevert` library

shuhuiluo:
## Related Issue

Bytecode size.

## Description of changes
Solidity creates a new [`JUMPDEST`](https://www.evm.codes/#5b?fork=cancun) at a section of bytecode containing the revert logic every time `revert CustomError()` is used. This pull request introduces a `CustomRevert` library with a `revertWith` function as a unified `JUMPDEST` for handling the revert of custom errors without an argument hence reducing contract bytecode. It also demonstrates a syntax sugar for passing custom error selector to the function with `using CustomRevert for bytes4;`.

By replacing `revert CustomError()` with `CustomError.selector.revertWith()`, contract size is reduced by 9 bytes for each revert of custom errors without an argument, allowing for more aggressive optimizer runs.


src/NoDelegateCall.sol
```diff
@@ -1,9 +1,13 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.20;
 
+import {CustomRevert} from "./libraries/CustomRevert.sol";
```
hensha256, line 4: could do `using CustomRevert for bytes4 global`?
I actually think we should do that with more of our libs too

src/libraries/CustomRevert.sol
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /// @dev Reverts with a custom error with two int24 arguments
+    /// @dev The free memory pointer is tampered with but it is fine since the call context is exited immediately
+    function revertWith(bytes4 selector, int24 value1, int24 value2) internal pure {
+        /// @solidity memory-safe-assembly
+        assembly {
+            mstore(0, selector)
+            mstore(0x04, value1)
+            mstore(0x24, value2)
```
hensha256, line 51: dont these top 4 bytes of the fmp need to be set back to 0?

hensha256: oh its reverting 💀

src/NoDelegateCall.sol
shuhuiluo: Unfortunately the `global` keyword only works for user-defined types.
https://docs.soliditylang.org/en/latest/contracts.html#using-for

hensha256 (approved):

hensha256 (approved):

PR #704, opened 2024-05-24
Use `abi.encodeCall` in place of `abi.encodeWithSelector`

shuhuiluo:
## Related Issue

## Description of changes
Replaced some instances of `abi.encodeWithSelector` with more preferable `abi.encodeCall` as it performs a full type-check, ensuring the types match the function signature.
https://docs.soliditylang.org/en/latest/cheatsheet.html#abi-encoding-and-decoding-functions

Reduced runtime gas and bytecode size. Did not replace everything since the rest incurs regression.


hensha256 (approved):
woo great PR

PR #710, opened 2024-05-24
Optimize `CurrencyLibrary.transfer`

shuhuiluo:
## Related Issue

## Description of changes
Use Solady implementation of `safeTransferETH` and `safeTransfer` with additional comments which are cleaner and more efficient. Saves runtime gas and bytecode size. Solady SafeTransferLib has also been audited.


src/types/Currency.sol
```diff
@@ -38,42 +38,36 @@ library CurrencyLibrary {
...
+        // https://github.com/Vectorized/solady/blob/0dde2a008d917aa8076f348eac2855edbe181cc0/src/utils/SafeTransferLib.sol
 
-        bool success;
         if (currency.isNative()) {
+            /// @solidity memory-safe-assembly
             assembly {
-                // Transfer the ETH and store if it succeeded or not.
-                success := call(gas(), to, amount, 0, 0, 0, 0)
+                // Transfer the ETH and revert if it fails.
+                if iszero(call(gas(), to, amount, 0, 0, 0, 0)) {
```
gretzke, line 47: this implementation is different from Vectorized implementation, why do you set `argsOffset` and `retOffset` to 0 instead of `codesize()`?

shuhuiluo: They are equivalent as long as the calldata length and returndata length are 0. Using `codesize()` is an optimization for chains without `PUSH0`.

gretzke (approved):

gretzke:
changed the base branch so we can fix merge conflicts

PR #711, opened 2024-05-24
Prevent duplicate dynamic fee pools

saucepoint:
## Related Issue

When initializing a dynamic fee pool, there is an opportunity to create "duplicate" pools. The following PoolKeys were valid and would initialize successfully

```solidity
PoolKey(currency0, currency1, 0x800000, 60, IHooks(0xABC))
PoolKey(currency0, currency1, 0x800001, 60, IHooks(0xABC))
PoolKey(currency0, currency1, 0x800002, 60, IHooks(0xABC))
```

We believe this can enable malicious/phishing behavior where the same hook contract has different behavior dependent on `key.fee`

## Description of changes

During pool initialization, if the key.fee has the dynamic fee flag and is not exactly equal to `0x800000`, revert with `FeeNotExactlyDynamic`


snreynolds (approved):

i9:
@saucepoint hi are we losing capability of setting default fee for a dynamic fee pool? We're building VIP fee discount hook based on trading volume, and depends on beforeSwap key.fee as original undiscounted fee.

saucepoint:
@i9 yes dynamic fee pools initialize to 0-fee by default. Please use afterInitialize to set a default fee!

PR #280, opened 2023-06-21
Refactor and optimize `SwapMath`

shuhuiluo:
## Related Issue
Which issue does this pull request resolve?

Obscure control flow related to `SwapMath`.

## Description of changes

Improvements to the `SwapMath` library have been implemented. A new function, `getSqrtRatioTarget`, written in `Yul` using bitwise operations, has been introduced to replace a nested ternary operation within `Pool.swap`:
```Solidity
(
    params.zeroForOne
        ? step.sqrtPriceNextX96 < params.sqrtPriceLimitX96
        : step.sqrtPriceNextX96 > params.sqrtPriceLimitX96
) ? params.sqrtPriceLimitX96 : step.sqrtPriceNextX96
```
which basically was doing
```Solidity
zeroForOne ? max(sqrtPriceNextX96, sqrtPriceLimitX96) : min(sqrtPriceNextX96, sqrtPriceLimitX96)
```

Furthermore, the `computeSwapStep` function has been refactored, resulting in a cleaner control flow and significant gas savings. The gas savings achieved range from 10% to 28%, as can be seen in `SwapMath.spec.ts.snap`. In conjunction with the changes introduced in pull request #258, the maximum savings on `computeSwapStep` can reach up to 40%.

All tests pass except for `SwapMath::#computeSwapStep::entire input amount taken as fee`, where `amountIn` changes from `0` to `9`, and `feeAmount` changes from `10` to `1`.

I wonder if "entire input amount taken as fee when output is zero" is an intended feature and the invariant `amountIn == 0 && feeAmount == 10` should be kept. Adjusting the invariant to `amountIn + feeAmount == 10` appears to be acceptable and reasonable.

PR #718, opened 2024-05-26
Avoid recomputing `amountIn` when `exactIn` and `amountRemainingLessF…

shuhuiluo:
…ee < amountIn`

Originally part of #280.

## Related Issue

## Description of changes
Simplify the flow in SwapMath library by removing unnecessary calculations related to input amounts. The removed calculations are replaced with a direct assignment of the remaining amount which simplifies the logic. Moreover, the potential for any rounding issues is significantly reduced.

In the case of `exactIn` and `amountRemainingLessFee < amountIn`, the original logic computes `sqrtPriceNextX96` from `amountRemainingLessFee`. It then uses `sqrtPriceNextX96` to compute `amountIn`, which shouldn't differs much from `amountRemainingLessFee`, while `amountIn` can simply exhaust `amountRemainingLessFee`.


hensha256 (dismissed):

hensha256:
@snreynolds is also going to review this. Please dont merge without approvals from both of us 🙏

snreynolds (dismissed):
makes sense to me! that edge case I believe is really just a rounding error and with this change it rounds the fee down which I believe is more accurate behavior?

hensha256 (approved):

snreynolds (approved):

PR #720, opened 2024-05-28
add mapping key names in pool state

xyek:
this PR simply adds the key names in the mappings in the Pool.State. It makes it a bit easy to imagine what the code must be doing while reading around that piece of code.


xyek:
Cannot change base branch. #734

PR #721, opened 2024-05-28
add getPoolState method to the StateLibrary

xyek:
This util is helpful when someone wants to read all the global pool states. It is better than making multiple extsload calls and also it does not reallocate memory apart from the returndata being copied to memory through the call.


test/libraries/StateLibrary.t.sol
```diff
@@ -37,6 +38,38 @@ contract StateLibraryTest is Test, Deployers, Fuzzers, GasSnapshot {
...
+
+        // swap to create fees, crossing a tick
+        uint256 swapAmount = 100 ether;
+        swap(key, true, -int256(swapAmount), ZERO_BYTES);
+
+        (Slot0 slot0, uint256 feeGrowthGlobal0X128, uint256 feeGrowthGlobal1X128, uint128 liquidity) =
+            StateLibrary.getPoolState(manager, poolId);
+        snapLastCall("extsload getPoolState");
+
+        assertEq(slot0.sqrtPriceX96(), 78680104762184586858280382455);
```
hensha256, line 59: can you instead test these against `getSlot0`?

src/libraries/StateLibrary.sol
```diff
@@ -30,6 +31,31 @@ library StateLibrary {
...
+     * @return feeGrowthGlobal1X128 The global fee growth for token1.
+     * @return liquidity The liquidity of the pool.
+     */
+    function getPoolState(IPoolManager manager, PoolId poolId)
+        internal
+        view
+        returns (Slot0 slot0, uint256 feeGrowthGlobal0X128, uint256 feeGrowthGlobal1X128, uint128 liquidity)
+    {
+        bytes32 stateSlot = _getPoolStateSlot(poolId);
+        bytes memory data = manager.extsload(stateSlot, 4);
```
hensha256, line 50: this needs to be updated to an array of bytes32 that gets returned

src/libraries/StateLibrary.sol
```diff
@@ -30,6 +31,31 @@ library StateLibrary {
...
+        assembly {
+            slot0 := mload(add(data, 32))
+            feeGrowthGlobal0X128 := mload(add(data, 64))
+            feeGrowthGlobal1X128 := mload(add(data, 96))
+            liquidity := mload(add(data, 128))
```
hensha256, line 55: needs to be masked to remove the other 128 bits of data from this data type

hensha256 (changes requested):

PR #722, opened 2024-05-28
Remove variable assignment for constant access in assembly

gretzke:
See #719, solved merge conflicts


src/libraries/TransientStateLibrary.sol
```diff
@@ -24,10 +24,9 @@ library TransientStateLibrary {
     /// @dev returns 0 if the reserves are not synced
     /// @dev returns type(uint256).max if the reserves are synced but the value is 0
     function getReserves(IPoolManager manager, Currency currency) internal view returns (uint256) {
-        bytes32 slot = RESERVES_OF_SLOT;
         bytes32 key;
         assembly {
```
hensha256, line 28: shouldnt this be memory safe too?

hensha256: i see a few others in this contract too

gretzke: weirdly enough this compiles without it being memory safe

hensha256 (approved):

PR #723, opened 2024-05-28
Annotate memory safe assembly for further gas optimizations

gretzke:


src/libraries/TransientStateLibrary.sol
```diff
@@ -25,7 +25,7 @@ library TransientStateLibrary {
     /// @dev returns type(uint256).max if the reserves are synced but the value is 0
     function getReserves(IPoolManager manager, Currency currency) internal view returns (uint256) {
         bytes32 key;
-        assembly {
+        assembly ("memory-safe") {
```
shuhuiluo, line 28: Just FYI, I think the gas savings come from missing "memory-safe" annotation in `TransientStateLibrary`. Using
```solidity
    assembly {
```
or
```solidity
    assembly ("memory-safe") {
```
shouldn't make a difference. But it's nice to unify the style.

gretzke: it does make a difference when compiling via IR, the optimizer behaves differently when it can assume the assembly to be memory safe, that's why gas usage is also different

snreynolds (approved):

PR #724, opened 2024-05-28
Optimize memory in get position

havi-kim:
## Description of changes
This PR optimizes the memory usage in the get function for retrieving position information by changing the order of encoding owner and salt. 

### Benefits:
- Gas Optimization:
  - Small reduction in gas usage.
- Code Clarity:
  - By not interfering with the free memory pointer at 0x40, the code now avoids any subtle confusion for future reviewers and maintainers, even though modifying the free memory pointer does not currently cause any issues.

### Drawbacks:
- ABI Encode Order Change:
  - The order of params in the ABI encoding has been altered.
  
  Please feel free to close this PR if you find it unnecessary.


hensha256:
Hi! We appreciate your contribution. Unfortunately it came after audits had already begun so we will not be able to accept it 🙏

PR #725, opened 2024-05-28
use bytes32

snreynolds:
## Related Issue


src/libraries/StateLibrary.sol
```diff
@@ -11,7 +11,7 @@ library StateLibrary {
     // | Name                  | Type                                    | Slot | Offset | Bytes | Contract                        |
     // |-----------------------|-----------------------------------------|------|--------|-------|---------------------------------|
     // | pools                 | mapping(PoolId => struct Pool.State)    | 6    | 0      | 32    | src/PoolManager.sol:PoolManager |
-    uint256 public constant POOLS_SLOT = 6;
+    bytes32 public constant POOLS_SLOT = bytes32(uint256(6));
```
gretzke, line 14: `bytes32(0x06)`?

gretzke (approved):

src/libraries/StateLibrary.sol
snreynolds: going to leave as is

PR #728, opened 2024-05-29
tests for fee controller gas limit

dianakocsis:
## Related Issue


src/test/ProtocolFeeControllerTest.sol
```diff
@@ -59,3 +59,11 @@ contract InvalidReturnSizeProtocolFeeControllerTest is IProtocolFeeController {
         }
     }
 }
+
+contract GasLimitProtocolFeeControllerTest is IProtocolFeeController {
+    function protocolFeeForPool(PoolKey memory /* key */ ) external pure returns (uint24) {
+        // consume gas
+        while (true) {}
```
hensha256, line 66: lol beautiful

test/ProtocolFeesImplementation.t.sol
```diff
@@ -218,4 +219,30 @@ contract ProtocolFeesTest is Test, GasSnapshot, Deployers {
...
+        gasLimitFeeController = new GasLimitProtocolFeeControllerTest();
+        protocolFees.setProtocolFeeController(gasLimitFeeController);
+        vm.prank(address(gasLimitFeeController));
+        (bool success, uint24 protocolFee) = protocolFees.fetchProtocolFee(key);
+        assertFalse(success);
+        assertEq(protocolFee, 0);
+    }
+
+    function test_fetchProtocolFee_revertsWithProtocolFeeCannotBeFetched() public {
+        protocolFees = new ProtocolFeesImplementation(9079256829993496519);
```
hensha256, line 233: Would be great to have this as a constant or something

this file can have a `GAS_LIMIT` and then set it as an immutable in the implementation contract. And then that contract can just access it too?

test/ProtocolFeesImplementation.t.sol
```diff
@@ -218,4 +219,30 @@ contract ProtocolFeesTest is Test, GasSnapshot, Deployers {
         assertFalse(success);
         assertEq(protocolFee, 0);
     }
+
+    function test_fetchProtocolFee_gasLimit() public {
+        gasLimitFeeController = new GasLimitProtocolFeeControllerTest();
```
hensha256, line 224: would love a clearer name - maybe `ConsumesGasLimit...`?

test/ProtocolFeesImplementation.t.sol
```diff
@@ -218,4 +219,30 @@ contract ProtocolFeesTest is Test, GasSnapshot, Deployers {
...
+        vm.expectRevert(IProtocolFees.ProtocolFeeCannotBeFetched.selector);
+        // consumes too much gas before fetching the fee
+        protocolFees.consumeGasLimitAndFetchFee(key);
+    }
+
+    function test_fetchProtocolFee_revertsWithGasLimitExceeded() public {
+        protocolFees = new ProtocolFeesImplementation(9079256829993496519);
+        protocolFees.setProtocolFeeController(feeController);
+        vm.prank(address(feeController));
+        vm.expectRevert();
```
hensha256, line 245: please can you fill in what the revert message should be for this case

hensha256: I cant see why this one will be any different to the previous one? surely it also reverts with `ProtocolFeeCannotBeFetched` because its consuming even more gas and the amount left is even lower?

hensha256 (changes requested):
Some thoughts

test/ProtocolFeesImplementation.t.sol
dianakocsis: it consumes less!

dianakocsis: the limit is the same, but in `consumeGasAndFetchFee`, its consuming less gas

src/test/ProtocolFeesImplementation.sol
```diff
@@ -33,4 +33,29 @@ contract ProtocolFeesImplementation is ProtocolFees {
     function updateProtocolFees(Currency currency, uint256 amount) public {
         ProtocolFees._updateProtocolFees(currency, amount);
     }
+
+    function consumeGasLimitAndFetchFee(PoolKey memory key) public {
+        // consume gas before calling fetchProtocolFee / getting the protocolFeeForPool from the controller
+        while (true) {
```
snreynolds, line 39: Seems like these functions are the same -- could just write one and have that value as an input i think

test/ProtocolFeesImplementation.t.sol
snreynolds: would agree here - not sure what the difference is if there is one besides just consuming a different amount of gas before calling fetchProtocolFee

You could fuzz that amount even

dianakocsis: oh yeah i messed up the numbers

src/test/ProtocolFeesImplementation.sol
havi-kim: Hello, I'm not part of your team, but I have a question. Doesn't this method increase the testing time?

test/ProtocolFeesImplementation.t.sol
havi-kim:
```
    function test_fetchProtocolFee_revertsWithProtocolFeeCannotBeFetched() public {
        uint256 gasLimit = 5000;
        protocolFees = new ProtocolFeesImplementation(gasLimit);
        protocolFees.setProtocolFeeController(feeController);
        vm.prank(address(feeController));
        vm.expectRevert(IProtocolFees.ProtocolFeeCannotBeFetched.selector);
        protocolFees.fetchProtocolFee{gas: gasLimit - 1}(key);
    }
```
I'm suggesting this because I'm concerned that the current method might increase the testing time.
Thank you for considering my suggestion.

PR #743, opened 2024-06-17
noDelegateCall on locked functions

jeffishjeff:
## Related Issue

Only protect initialize() and unlock() with noDelegateCall isn't enough as potential exploiter can just modify a proxy contract's lock (e.g. call unlock() on the proxy) then make delegate calls to Uniswap's pool manager, getting around the business license

## Description of changes

- add noDelegateCall to functions with onlyWhenUnlocked in PoolManager
- for test
    - a mock ProxyPoolManager under src/test
    - a set of delegate call tests
    - minor changes to Deployer and PoolClaimsTest to not reinvent the wheel


test/PoolManagerDelegateCall.t.sol
```diff
@@ -0,0 +1,64 @@
...
+import {IPoolManager} from "../src/interfaces/IPoolManager.sol";
+import {NoDelegateCall} from "../src/NoDelegateCall.sol";
+import {PoolManager} from "../src/PoolManager.sol";
+import {PoolSwapTest} from "../src/test/PoolSwapTest.sol";
+import {ProxyPoolManager} from "../src/test/ProxyPoolManager.sol";
+import {Deployers} from "./utils/Deployers.sol";
+
+contract PoolManagerDelegateCall is Test, Deployers {
+    // override to use ProxyPoolManager
+    function deployFreshManager() internal virtual override {
```
snreynolds, line 15: Could you add these tests to NoDelegateCall.t.sol?

snreynolds (commented):
Hi thanks so much for contributing this! 

Few comments:
- Could you run forge fmt? Seems like some formatting diffs.
- I think this means we don't need noDelegateCall on unlock yes?

test/PoolManagerDelegateCall.t.sol
jeffishjeff: Thanks for the feedback, will have these changes in shortly.

Want to pick your brain on which functions to have noDelegateCall though, now that it's logically uncoupled from unlock. I see two potential approaches:

- protect just the core pool logic, namely modifyLiquidity, swap, and donate; and leave it off mint/burn and settle/take
- protect all interface functions, that would also include sync and updateDynamicLFFee

Somehow I feel the line currently drawn somewhat arbitrary...

jeffishjeff: Changes re comment checked in

jeffishjeff: forge fmt made difference though, not sure what the formatting diffs may be..

snreynolds: I do agree that the line is somewhat arbitrary currently. I think core functionality swap/modify/donate makes sense. I'd probably opt for it to also be on initialize. But, I don't feel strongly it needs to be on mint/burn and take/settle though. @saucepoint / @gretzke thoughts?

snreynolds: for reference this was what we previously had changed https://github.com/Uniswap/v4-core/pull/549/files#diff-84227e9327ad53c4ad37657de3ed4ee30143b70e038ddabd7e6214eba4f0865c

snreynolds:
I think you may need to update the snapshot file also

if you see gas snapshots change when you run `forge test --isolate` you can commit those changes! I'm seeing some failing tests bc of gas mismatchees

test/PoolManagerDelegateCall.t.sol
snreynolds: Ok I think my final/full opinion here re:exact functions for this modifier. We should include it on:

- initialize
- modifyLiquidity
- swap
- donate

Justification: Want to keep this overhead minimal & those are the core swap lifecycle functions. I don't see any harm with someone delegatecalling mint/burn and take/settle sync/ other public/external functions.

Let me know thoughts @jeffishjeff

snreynolds (commented):
nice looks good - just need that final update to remove from mint/burn & take/settle and then I can merge

jeffishjeff:
Checked in, also removed associated tests for mint/burn and take/settle, and restored PoolClaimsTest

snreynolds (approved):

PR #748, opened 2024-06-18
add poolKeyLibrary

Jun1on:
## Related Issue

confusing code when determining which token is specified/unspecified

## Description of changes

new library for PoolKey

### DIFF
--- a/src/types/PoolKey.sol
+++ b/src/types/PoolKey.sol
@@ -22,7 +22,7 @@ struct PoolKey {
 
 library PoolKeyLibrary {
     /// @notice Returns the unspecified currency and amount for a swap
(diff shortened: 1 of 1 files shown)


src/types/PoolKey.sol
```diff
@@ -17,3 +18,25 @@ struct PoolKey {
     /// @notice The hooks of the pool
     IHooks hooks;
 }
+
+library PoolKeyLibrary {
+    /// @notice Returns the unspecified currency and amount for a swap
+    function getUnspecified(PoolKey calldata key, IPoolManager.SwapParams calldata params)
```
ConjunctiveNormalForm, line 24: it's confusing that this function returns the *unspecified* currency but *specified* amount

PR #751, opened 2024-06-21
Clarify comment transient storage Solidity and Assembly

MarcusWentz:
## Description of changes

Transient storage opcodes are known now after Ethereum's Dencun Fork Update.
Will library Lock.sol still be deleted based on the contract's comments?

### DIFF
--- a/src/libraries/Lock.sol
+++ b/src/libraries/Lock.sol
@@ -4,7 +4,8 @@ pragma solidity ^0.8.20;
 import {IHooks} from "../interfaces/IHooks.sol";
 
 /// @notice This is a temporary library that allows us to use transient storage (tstore/tload)
-/// TODO: This library can be deleted when we have the transient keyword support in solidity.
+/// TODO: This library can be deleted when we have transient storage commands in Solidity 
+/// instead of Assembly for tload(key) and tstore(key,value) documented here: https://www.evm.codes/
 library Lock {
     // The slot holding the unlocked state, transiently. bytes32(uint256(keccak256("Unlocked")) - 1)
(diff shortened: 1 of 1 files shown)


hensha256:
The `todo` is us waiting for the `transient` keyword so that one day we might be able to write
```
bool transient isLocked;
```
and not need to use assembly. However for now no such keyword exists.

MarcusWentz:
Solidity 0.8.27 Release Announcement:

https://soliditylang.org/blog/2024/09/04/solidity-0.8.27-release-announcement/

# Transient Storage

This release introduces support for transient storage variables into the parser. To clarify, the feature as a whole is not yet complete.

The compiler will now accept the syntax for marking some variables as transient and let users generate transient storage layout. However, it is still not possible to generate bytecode for contracts using such variables.

We plan to introduce high-level language support for transient storage gradually, over the next few releases. In particular, the next release will provide full support for transient state variables of value types. The support will be extended from there to cover more complex use cases.

MarcusWentz:
High-Level Language Support for Transient Storage 

https://github.com/ethereum/solidity/issues/15007

Restrict the initialization of transient storage state variables  

https://github.com/ethereum/solidity/pull/15317

Solidity transient type example from this pull request:

```solidity
// SPDX-License-Identifier: MIT
pragma solidity 0.8.27;

contract C {
	int transient x = -99;
	address transient a = address(0xABC);
	bool transient b = x > 0 ? false : true;
}
```
Remix IDE throws the expected error for each transient declaration (for now until more support is added): 
```
DeclarationError: Initialization of transient storage state variables is not supported.
```

PR #754, opened 2024-06-27
add asOverrideFee

Jun1on:
## Related Issue
dynamic fee must have 2nd highest bit set

## Description of changes
add asOverrideFee in LPFeeLibrary

### DIFF
--- a/src/libraries/LPFeeLibrary.sol
+++ b/src/libraries/LPFeeLibrary.sol
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity ^0.8.20;
 
-import {PoolKey} from "../types/PoolKey.sol";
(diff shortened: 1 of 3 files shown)


src/libraries/LPFeeLibrary.sol
```diff
@@ -63,4 +62,8 @@ library LPFeeLibrary {
         fee = self.removeOverrideFlag();
         fee.validate();
     }
+
+    function asOverrideFee(uint24 self) internal pure returns (uint24) {
```
marktoda, line 66: natspec

saucepoint:
iirc we originally had this helper function in LPFeeLibrary, but we deferred to putting it in periphery. The reasoning was anything defined in `v4-core/src` should also be used/reference by other `v4-core/src` files

PR #759, opened 2024-07-09
OZ M01, Certora M01: Fix memory safe assembly

hensha256:


src/Extsload.sol
```diff
@@ -17,42 +17,46 @@ abstract contract Extsload is IExtsload {
     /// @inheritdoc IExtsload
     function extsload(bytes32 startSlot, uint256 nSlots) external view returns (bytes memory) {
         assembly ("memory-safe") {
-            // The abi offset of dynamic array in the returndata is 32.
-            mstore(0, 0x20)
+            let memptr := mload(0x40)
```
gretzke, line 20: I thought we didn't have to care about memory safety here, because the view function is external

src/libraries/Hooks.sol
```diff
@@ -135,8 +135,9 @@ library Hooks {
                     revert(0x1c, 0x04)
                 }
                 // bubble up revert
-                returndatacopy(0, 0, returndatasize())
-                revert(0, returndatasize())
+                let fmp := mload(0x40)
+                returndatacopy(fmp, 0, returndatasize())
```
gretzke, line 139: Does memory safety matter if the function reverts?

src/types/Currency.sol
```diff
@@ -36,36 +36,44 @@ library CurrencyLibrary {
...
         if (currency.isNative()) {
             assembly ("memory-safe") {
                 // Transfer the ETH and revert if it fails.
-                if iszero(call(gas(), to, amount, 0x00, 0x00, 0x00, 0x00)) {
-                    mstore(0x00, 0xf4b3b1bc) // `NativeTransferFailed()`.
-                    revert(0x1c, 0x04)
-                }
+                success := call(gas(), to, amount, 0, 0, 0, 0)
             }
+            if (!success) revert NativeTransferFailed();
```
gretzke, line 48: `NativeTransferFailed.selector.revertWith()`?

src/types/Currency.sol
```diff
@@ -36,36 +36,44 @@ library CurrencyLibrary {
...
-                    revert(0x1c, 0x04)
-                }
-                mstore(0x34, 0) // Restore the part of the free memory pointer that was overwritten.
+
+                // Now clean the memory we used
+                mstore(fmp, 0) // 4 byte `selector` and 28 bytes of `to` were stored here
+                mstore(add(fmp, 0x20), 0) // 4 bytes of `to` and 28 bytes of `amount` were stored here
+                mstore(add(fmp, 0x40), 0) // 4 bytes of `amount` were stored here
             }
+            if (!success) revert ERC20TransferFailed();
```
gretzke, line 76: `ERC20TransferFailed.selector.revertWith()`?

src/libraries/CustomRevert.sol
```diff
@@ -45,20 +45,22 @@ library CustomRevert {
...
         assembly ("memory-safe") {
-            mstore(0, selector)
-            mstore(0x04, value1)
-            mstore(0x24, value2)
-            revert(0, 0x44)
+            let fmp := mload(0x40)
+            mstore(fmp, selector)
+            mstore(add(fmp, 0x04), value1)
+            mstore(add(fmp, 0x24), value2)
+            revert(fmp, 0x44)
```
gretzke, line 63: does it matter if we overwrite the fmp if we revert anyways?

src/Extsload.sol
gretzke: here I mean, because the function is external, it enters a new call context and it exits right after, so we should be able to disregard the memory conventions

hensha256: Yeah I had a long conversation with our auditors about this. This is the only issue that was raised by _all 3_ audit firms.
Yes it is correct that the frame ends after this external call, but the code does not follow the conventions of memory safely. While it is not currently vulnerable to an attack that any of them can find, it is not future proof - because the assembly is marked as `memory-safe` but it does not follow the rules of memory safety. So if any changes are ever made to how memory is handled, this code might become unsafe due to the fact its been labelled as safe but isnt.

src/libraries/CustomRevert.sol
hensha256: yes it does. If youre going to mark assembly as memory safe, the execution needs to end with the fmp pointing to a reasonable memory area.
Basically no one can find a problem but they all think its hacky and could lead to very undefined behaviour as execution exits with fmp holding a very exotic value.

hensha256: I'll dm you a link to the long conversation I had with openzeppelin on slack 😁

gretzke (approved):

PR #771, opened 2024-07-09
OZ M02: ProtocolFeeController Gas Griefing

dianakocsis:
## Related Issue
OpenZeppelin M-02 - ProtocolFeeController Gas Griefing

## Description of changes
using assembly to execute the call and manually copying only 32 bytes from the return data to memory


src/ProtocolFees.sol
```diff
@@ -69,21 +69,29 @@ abstract contract ProtocolFees is IProtocolFees, Owned {
...
-                returnData := mload(add(_data, 0x20))
+                success := call(gasLimit, toAddress, 0, add(data, 0x20), mload(data), 0, 0)
+
+                // success if return data size is 32 bytes
+                success := and(success, eq(returndatasize(), 32))
+
+                // load the return data if success is true
+                if success {
+                    let fmp := mload(0x40)
+                    returndatacopy(fmp, 0, returndatasize())
```
hensha256, line 86: is it cheaper to hardcode `returndatasize()` as 32?
We know its 32 from the check above `eq(returndatasize(), 32)`

dianakocsis: it actually made it go up by 1!

hensha256: lol weird

src/ProtocolFees.sol
```diff
@@ -69,21 +69,29 @@ abstract contract ProtocolFees is IProtocolFees, Owned {
...
+            address toAddress = address(protocolFeeController);
 
+            bytes memory data = abi.encodeCall(IProtocolFeeController.protocolFeeForPool, (key));
             uint256 returnData;
             assembly ("memory-safe") {
-                returnData := mload(add(_data, 0x20))
+                success := call(gasLimit, toAddress, 0, add(data, 0x20), mload(data), 0, 0)
+
+                // success if return data size is 32 bytes
+                success := and(success, eq(returndatasize(), 32))
```
hensha256, line 81: could you add another comment that just says something like
```
// we only load the return value if its the correct length to prevent gas griefing
```

hensha256 (approved):

PR #774, opened 2024-07-09
OZ N16: Create SECURITY.md

gretzke:
## Description of changes

Add file with security information

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,12 +2,12 @@
 
 ## Vulnerability Disclosure and Bug Bounty
 
-Bug bounty details can be found in https://uniswap.org/bug-bounty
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,13 @@
+# Uniswap Labs Security
```
hensha256, line 1: The audit also says
```
Consider adding a NatSpec comment containing a security contact above each contract definition. Using the @custom:security-contact convention is recommended
```

Shall we add the contact at the top of the PoolManager too?

gretzke: I'm not against it but I'm also fine with just the security file

hensha256: ok fair i'll merge

PR #779, opened 2024-07-10
C01: native ERC20 double spend

vimageDE:
## Related Issue

C01

## Description of changes

Fixes an issue where native tokens that have a native ERC20 representation (e.g., CELO) could be credited twice by syncing the ERC20 version, settling the native token and then settling the ERC20 token without an actual ERC20 transfer required.

This fix requires a strict flow where an ERC20 token needs to settle and sync directly after, settling the native token after the ERC20 settle would revert.


src/test/ActionsRouter.sol
```diff
@@ -100,8 +108,8 @@ contract ActionsRouter is IUnlockCallback, Test {
     }
 
     function _assertReservesEquals(bytes memory params) internal view {
-        (Currency currency, uint256 expectedReserves) = abi.decode(params, (Currency, uint256));
-        assertEq(manager.getReserves(currency), expectedReserves, "reserves value incorrect");
+        (, uint256 expectedReserves) = abi.decode(params, (Currency, uint256));
```
hensha256, line 111: why does it still decode a currency at all? shouldnt it just only pass in the uint256 now?

src/test/NativeERC20.sol
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+import {Test} from "forge-std/Test.sol";
+
+pragma solidity ^0.8.0;
+
+contract NativeERC20 is Test {
```
hensha256, line 7: Would love a comment above this contract explaining what it does - and that transfer changes balances of native too

test/Reserves.t.sol
```diff
@@ -45,6 +40,6 @@ contract ReservesTest is Test {
     function test_fuzz_get_set(Currency currency, uint256 value) public {
         vm.assume(value != type(uint256).max);
         currency.setReserves(value);
-        assertEq(currency.getReserves(), value);
+        assertEq(Reserves.getReserves(), value);
```
hensha256, line 43: these tests should add checking the getter for the currency!
Yu have `currency.setReserves(value)` so then we need to check
- is value set correctly (you have this)
- is currency set correctly (this needs to be added)

src/libraries/Reserves.sol
```diff
@@ -3,40 +3,48 @@ pragma solidity ^0.8.20;
 
 import {Currency} from "../types/Currency.sol";
 import {CustomRevert} from "./CustomRevert.sol";
+import {IPoolManager} from "../interfaces/IPoolManager.sol";
 
 library Reserves {
```
hensha256, line 8: Ok heres a suggestion for naming. Open to suggestions though

- `library CurrencyReserves`
- `constant RESERVES_OF_SLOT` ✅ 
- `SYNC_SLOT` -> `CURRENCY_SLOT`
- `function checkSync` -> `function checkCurrencyNotSet`
- `Currency syncing` -> `Currency currencySet`
- `function getSyncedCurrency` -> `function getCurrency`
- `function getReserves` ✅ 
- `function setReserves` -> `function setCurrencyAndReserves`

src/libraries/TickBitmap.sol
```diff
@@ -43,7 +43,7 @@ library TickBitmap {
     /// @param tick The tick to flip
     /// @param tickSpacing The spacing between usable ticks
     function flipTick(mapping(int16 => uint256) storage self, int24 tick, int24 tickSpacing) internal {
-        /**
+        /*
```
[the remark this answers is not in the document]
gretzke: fixes a compiler warning because it assumed this was natspec, I tried to sneak it in in this PR lol

src/PoolManager.sol
```diff
@@ -264,13 +261,15 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     }
 
     /// @inheritdoc IPoolManager
-    function sync(Currency currency) public returns (uint256 balance) {
-        balance = currency.balanceOfSelf();
-        currency.setReserves(balance);
+    function sync(Currency currency) external {
+        Reserves.checkSync();
+        if (currency.isNative()) return;
```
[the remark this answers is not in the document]
gretzke: natspec or inline?

src/libraries/TickBitmap.sol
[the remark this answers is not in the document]
hensha256: LOL ok fine

src/PoolManager.sol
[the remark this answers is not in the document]
hensha256: i think natspec maybe? under a @dev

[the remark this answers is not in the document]
gretzke: Poor Kenny probably being pinged 1000 times a day lol

src/interfaces/IPoolManager.sol
```diff
@@ -45,6 +45,8 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload, IExttload {
     ///@notice Thrown when native currency is passed to a non native settlement
     error NonZeroNativeValue();
 
+    error AlreadySynced();
```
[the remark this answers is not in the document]
vimageDE: fixed :)

src/libraries/Reserves.sol
vimageDE: Thanks a lot for the suggestions!
I am wondering about removing sync out of some of the names... Does it not provide more clarity to keep Synced in the naming, Since all of it is so closely connected to the process of syncing?
I do agree with the fact that there is no clear line currently though, but I personally think it might help with readability by leaning into the other direction (I have included the changes I fully agree with for clarity) :

- `library Reserves` -> `library CurrencyReserves`
- `constant RESERVES_OF_SLOT` ✅
- `SYNC_SLOT` -> `CURRENCY_SLOT`
- `function checkSync` -> `function checkNotSynced`
- `Currency syncing` -> `Currency currencySynced`
- `function getSyncedCurrency` ✅
- `function getReserves` -> `getSyncedReserves`
- `function setReserves` -> `function syncCurrencyAndReserves`

Let me know what you think, happy to stick with your naming suggestions, if you like it more that way :)

src/PoolManager.sol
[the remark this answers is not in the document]
vimageDE: Just want to let Kenny know that I added the comment in the IPoolManager! 😂

src/libraries/TransientStateLibrary.sol
```diff
@@ -17,20 +17,13 @@ library TransientStateLibrary {
...
     /// @dev returns type(uint256).max if the reserves are synced but the value is 0
-    function getReserves(IPoolManager manager, Currency currency) internal view returns (uint256) {
-        bytes32 key;
-        assembly ("memory-safe") {
-            mstore(0, RESERVES_OF_SLOT)
-            mstore(32, currency)
-            key := keccak256(0, 64)
-        }
-        return uint256(manager.exttload(key));
+    function getReserves(IPoolManager manager) internal view returns (uint256) {
```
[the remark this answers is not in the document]
vimageDE: Added view function for the currency! 
Also removed the duplications of the slots, imported them instead from the libraries, as discussed.

src/test/ActionsRouter.sol
vimageDE: does not need it, good catch!

test/Reserves.t.sol
vimageDE: Added checks for the currency in the tests and added a new test to confirm the functionality of `Reserves.reset()`

src/PoolManager.sol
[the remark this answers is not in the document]
snreynolds: just letting kenny know natspec comment looks good

src/libraries/Reserves.sol
vimageDE: _-For documentation purposes-_
After an internal discussion, we decided on these namings:
- `library Reserves` -> `library CurrencyReserves`
- `constant RESERVES_OF_SLOT` ✅
- `SYNC_SLOT` -> `CURRENCY_SLOT`
- `function checkSync` -> `function requireNotSynced`
- `Currency syncing` -> `Currency currencySynced`
- `function getSyncedCurrency` ✅
- `function getReserves` -> `getSyncedReserves`
- `function setReserves` -> `function syncCurrencyAndReserves`

src/PoolManager.sol
```diff
@@ -279,21 +278,24 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
...
-            uint256 reservesBefore = currency.getReserves();
-            uint256 reservesNow = sync(currency);
+            // Reserves are guaranteed to be set, because currency and reserves are always set together
+            uint256 reservesBefore = Reserves.getReserves();
+            uint256 reservesNow = currency.balanceOfSelf();
             paid = reservesNow - reservesBefore;
         }
 
         _accountDelta(currency, paid.toInt128(), msg.sender);
+        Reserves.reset();
```
snreynolds: I had a question - if we move this up into the else clause.. Does that basically re introduce the issue?

src/libraries/Reserves.sol
```diff
@@ -7,36 +7,46 @@ import {CustomRevert} from "./CustomRevert.sol";
...
+    function getSyncedCurrency() internal view returns (Currency currency) {
         assembly {
-            tstore(key, value)
+            currency := tload(SYNC_SLOT)
         }
     }
 
-    function getReserves(Currency currency) internal view returns (uint256 value) {
-        bytes32 key = _getKey(currency);
+    function reset() internal {
```
snreynolds: can you add a comment here that the reserves of slot will be dirtied, but it is safe because it is always overridden before the next settle... or just to be extra safe we should wipe it even in reset (in which case you can use Reserves.setReserves(0, 0);

src/PoolManager.sol
vimageDE: good catch

src/PoolManager.sol
```diff
@@ -279,21 +278,23 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
...
-    function settle(Currency currency) external payable override onlyWhenUnlocked returns (uint256 paid) {
+    function settle() external payable onlyWhenUnlocked returns (uint256 paid) {
+        Currency currency = Reserves.getSyncedCurrency();
         if (currency.isNative()) {
             paid = msg.value;
         } else {
             if (msg.value > 0) NonZeroNativeValue.selector.revertWith();
-            uint256 reservesBefore = currency.getReserves();
-            uint256 reservesNow = sync(currency);
+            uint256 reservesBefore = Reserves.getReserves();
```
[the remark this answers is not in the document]
vimageDE: If you just call settle without sync, you won't have a currency set and the settle function therefore expects a native token settle, so should not reach this else part... or did I miss out on something?

src/PoolManager.sol
```diff
@@ -264,13 +261,15 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     }
 
     /// @inheritdoc IPoolManager
-    function sync(Currency currency) public returns (uint256 balance) {
-        balance = currency.balanceOfSelf();
-        currency.setReserves(balance);
+    function sync(Currency currency) external {
+        Reserves.checkNotSynced();
+        if (currency.isNative()) return;
```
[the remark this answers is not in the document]
vimageDE: This could make it easier for people to work with, since they can stick to the sync->settle flow, even for native tokens.
But if you think it should better revert in that scenario, let me know!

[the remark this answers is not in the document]
hensha256: Im happy with it returning instead. Theres no need for it to revert imo

src/PoolManager.sol
```diff
@@ -279,21 +278,25 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     }
 
     /// @inheritdoc IPoolManager
-    function settle(Currency currency) external payable override onlyWhenUnlocked returns (uint256 paid) {
+    function settle() external payable onlyWhenUnlocked returns (uint256 paid) {
+        Currency currency = CurrencyReserves.getSyncedCurrency();
+        // If not previously synced, expects native currency to be settled because currency.isNative() == address(0)
```
hensha256, line 283:
```suggestion
        // If not previously synced, expects native currency to be settled because CurrencyLibrary.NATIVE == address(0)
```

hensha256: nit... but `currency.isNative()` returns a boolean, so will never be equal to address(0)

src/libraries/CurrencyReserves.sol
```diff
@@ -0,0 +1,52 @@
...
+
+    function getSyncedCurrency() internal view returns (Currency currency) {
+        assembly {
+            currency := tload(CURRENCY_SLOT)
+        }
+    }
+
+    function reset() internal {
+        assembly {
+            tstore(CURRENCY_SLOT, 0)
```
hensha256, line 36: would rather this also reset the reserves of. I know its not strictly needed, but it would be weird that `getSyncedReserves` returns a non-0 value, when `getSyncedCurrency` returns 0.

src/test/NativeERC20.sol
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+import {Test} from "forge-std/Test.sol";
+
+pragma solidity ^0.8.0;
```
hensha256, line 5: pragma should be above imports

test/CurrencyReserves.t.sol
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    function test_reset_currency() public {
+        currency0.syncCurrencyAndReserves(100);
+        uint256 value = CurrencyReserves.getSyncedReserves();
+        assertEq(value, 100);
+        assertEq(Currency.unwrap(CurrencyReserves.getSyncedCurrency()), Currency.unwrap(currency0));
+        CurrencyReserves.reset();
+        uint256 valueAfterReset = CurrencyReserves.getSyncedReserves();
+        assertEq(valueAfterReset, 100);
```
hensha256, line 39: yeah definitely dont like this - would love it to also be reset to 0 please

test/Sync.t.sol
```diff
@@ -263,4 +180,45 @@ contract SyncTest is Test, Deployers, GasSnapshot {
...
+
+    function test_settle_failsWithNativeERC20IfNotSyncedInOrder(uint256 value) public noIsolate {
+        value = bound(value, 1, uint256(int256(type(int128).max / 2)));
+        vm.deal(address(this), value);
+        vm.deal(address(manager), value);
+        NativeERC20 nativeERC20 = new NativeERC20();
+
+        manager.sync(Currency.wrap(address(nativeERC20)));
+
+        // uint256 balanceBefore = address(this).balance;
```
hensha256, line 192: delete comment or move down to be part of the OZ attack comment

src/PoolManager.sol
vimageDE: good catch!

src/libraries/CurrencyReserves.sol
vimageDE: Are you fine with the extra gas costs? Will effect every non native settle...

test/Sync.t.sol
vimageDE: Are you happe with the way it is handled at the bottom of the test?

https://github.com/vimageDE/v4-core/blob/1241d0c37c73ba766151c8fea4fdd30e8c5f49cb/test/Sync.t.sol#L221

src/libraries/CurrencyReserves.sol
hensha256: yes i'm ok with the extra cost

test/Sync.t.sol
hensha256: yeah i basically just think the entire commented out test should be together, and not interspersed among the new test

src/PoolManager.sol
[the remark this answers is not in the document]
hensha256: yeah exactly, the fact that youre in the `else` branch means that `currency` was non-0. This means `sync` must have happened. This means that `reserves` must also have been correctly synced.

test/CurrencyReserves.t.sol
snreynolds: Its strictly never used by the contract and 100 extra gas

snreynolds: We should just be checking that that value is not used after another call to sync

test/Sync.t.sol
```diff
@@ -137,95 +129,18 @@ contract SyncTest is Test, Deployers, GasSnapshot {
         router.executeActions(actions, params);
     }
 
-    /// @notice When there is no actual balance in the pool, the ZERO_BALANCE stored in transient reserves should never actually used in calculating the amount paid in settle.
```
snreynolds, line 140: RIP this test, but i think its a great side effect of this PR that this footgunny behaviour is not possible now but

test/Sync.t.sol
```diff
@@ -239,22 +154,24 @@ contract SyncTest is Test, Deployers, GasSnapshot {
...
+        actions[3] = Actions.SETTLE; // Since reserves now == reserves, paid = 0 and the delta owed by the user will still be -10 after settle.
 
         actions[4] = Actions.ASSERT_DELTA_EQUALS;
         params[4] = abi.encode(currency0, address(router), -10);
 
+        actions[5] = Actions.SYNC;
+        params[5] = abi.encode(currency0);
+
         // To now settle the delta, the user owes 10 to the pool.
         // Because sync is called in settle we can transfer + settle.
```
snreynolds, line 166: this comment is no longer true

test/Sync.t.sol
```diff
@@ -137,95 +129,18 @@ contract SyncTest is Test, Deployers, GasSnapshot {
...
-        actions[9] = Actions.SETTLE;
-        params[9] = abi.encode(currency3);
-
-        // Expect an underflow/overflow because reservesBefore > reservesNow since sync() had not been called before settle.
-        vm.expectRevert(abi.encodeWithSignature("Panic(uint256)", 0x11));
-
-        router.executeActions(actions, params);
-    }
-
     // @notice This tests expected behavior if you DO NOT call sync. (ie. Do not interact with the pool manager properly. You can lose funds.)
```
snreynolds, line 132: Can you update the comment and test name - I feel this is out of date? And it would be nice to have a new summary of whats being tested

src/libraries/CurrencyReserves.sol
vimageDE: I am happy with both... I really like the gas save, but I agree with the fact, that it will be confusing to get an outdated value with the getter function.

test/Sync.t.sol
vimageDE: Done

src/libraries/CurrencyReserves.sol
```diff
@@ -0,0 +1,52 @@
...
+
+library CurrencyReserves {
+    using CustomRevert for bytes4;
+
+    /// @notice Thrown when a user has already synced a currency, but not yet settled
+    error AlreadySynced();
+
+    /// bytes32(uint256(keccak256("ReservesOf")) - 1)
+    bytes32 constant RESERVES_OF_SLOT = 0x1e0745a7db1623981f0b2a5d4232364c00787266eb75ad546f190e6cebe9bd95;
+    /// bytes32(uint256(keccak256("Sync")) - 1)
```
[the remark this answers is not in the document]
vimageDE: fully agree, fixed!

src/libraries/CurrencyReserves.sol
```diff
@@ -0,0 +1,52 @@
...
+    bytes32 constant CURRENCY_SLOT = 0xf0e14a408baf7f453312eec68e9b7d728ec5337fbdf671f917ee8c80f3255231;
+
+    function requireNotSynced() internal view {
+        Currency currencySynced;
+        assembly {
```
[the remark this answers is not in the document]
vimageDE: True, good spot!

test/Sync.t.sol
vimageDE: You are right, this test is not worth to have at all anymore, removed

test/Sync.t.sol
vimageDE: Removed

test/Sync.t.sol
vimageDE: Updated the comment and the name, also added some comments in the test to clarify whats is happening!

test/CurrencyReserves.t.sol
gretzke: I also don't love the extra gas, but I also don't love we leave it unset. What if we read both in the transient state library and if currency == 0, we return 0 for reserves?

vimageDE: Agree, that would make it clear for anyone using the TransientStateLibrary library.
And either the extra gas is used in the settle for deleting the currency, or alternative in the TransientStateLibrary read, but that one should probably have lower prio then the settle?
Also for the CurrendyReserves `getReserves` function, it was renamed to `getSyncedReserves`, so the name helps to indicates the sync condition I would say.

vimageDE: Plus anyone using the TransientStateLibrary could decide themselves to remove it if they want to save the gas in the read :)

snreynolds: I like this suggestion for TransientStateLibrary

snreynolds (approved):
Awesome work 👏

snreynolds (changes requested):
Oops, see that the TransientStateLibrary still needs an update so it overrides the reserves balance. I think I'd just say to make @gretzke's suggestion & update tests and then I'll re-approve.

snreynolds (approved):

PR #782, opened 2024-07-12
ABDK CVF8 & 10

hensha256:
Add comments to warn that code-changes could lead to overflow problems that currently dont exist


snreynolds (approved):
Are these prompted by an audit change?

hensha256:
Yes! By ABDK CVF8 & 10 :)

snreynolds (commented):
I think this looks good, just want to make sure ABDK are good with the comment and also agree that constraints on the protocolFee address their phantom overflow concern

PR #785, opened 2024-07-12
Various ABDK audit fixes

hensha256:
Issue numbers in commit messages


src/libraries/BitMath.sol
```diff
@@ -3,96 +3,47 @@ pragma solidity ^0.8.20;
...
-            if (x >= 0x2) r += 1;
+        assembly ("memory-safe") {
+            r := or(shl(8, iszero(x)), shl(7, lt(0xffffffffffffffffffffffffffffffff, x)))
+            r := or(r, shl(6, lt(0xffffffffffffffff, shr(r, x))))
+            r := or(r, shl(5, lt(0xffffffff, shr(r, x))))
+            r := or(r, shl(4, lt(0xffff, shr(r, x))))
+            r := or(r, shl(3, lt(0xff, shr(r, x))))
+            // forgefmt: disable-next-item
+            r := or(r, byte(and(0x1f, shr(shr(r, x), 0x8421084210842108cc6318c6db6d54be)),
+                0x0706060506020504060203020504030106050205030304010505030400000000))
```
wjmelements, line 23: should this also mention De Bruijn?

src/interfaces/IPoolManager.sol
```diff
@@ -53,7 +53,12 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload, IExttload {
     /// @param tickSpacing The minimum number of ticks between initialized ticks
```
snreynolds, line 53: Questions: 

Why did you implement and revert CVF92?

src/libraries/BitMath.sol
```diff
@@ -3,96 +3,47 @@ pragma solidity ^0.8.20;
...
 /// @dev This library provides functionality for computing bit properties of an unsigned integer
+/// @author Solady (https://github.com/Vectorized/solady/blob/8200a70e8dc2a77ecb074fc2e99a2a0d36547522/src/utils/LibBit.sol)
 library BitMath {
     /// @notice Returns the index of the most significant bit of the number,
     ///     where the least significant bit is at index 0 and the most significant bit is at index 255
-    /// @dev The function satisfies the property:
-    ///     x >= 2**mostSignificantBit(x) and x < 2**(mostSignificantBit(x)+1)
     /// @param x the value for which to compute the most significant bit, must be greater than 0
     /// @return r the index of the most significant bit
     function mostSignificantBit(uint256 x) internal pure returns (uint8 r) {
```
snreynolds, line 12: solmates version returns a uint256 - is this valid if we truncate to uint8?

src/libraries/SwapMath.sol
```diff
@@ -42,7 +42,8 @@ library SwapMath {
     /// @return sqrtPriceNextX96 The price after swapping the amount in/out, not to exceed the price target
     /// @return amountIn The amount to be swapped in, of either currency0 or currency1, based on the direction of the swap
     /// @return amountOut The amount to be received, of either currency0 or currency1, based on the direction of the swap
-    /// @return feeAmount The amount of input that will be taken as a fee
+    /// @return feeAmount The amount of input that will be taken as a fee/
+    /// @dev feePips must be no larger than MAX_FEE_PIPS
```
snreynolds, line 46: It seems like 39 is saying its not guaranteed to be less than MAX_FEE_PIPS. Can we make this comment say it is guranteed because feePips is capped before computeSwapStep? 

And then I think we need a fuzz test to ensure that a) calculateSwapFee < MAX and b) setLpFee never allows setting greater than MAX

src/interfaces/IPoolManager.sol
```diff
@@ -102,7 +107,8 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload, IExttload {
         external
         returns (int24 tick);
 
-    /// @notice All operations go through this function
+    /// @notice All interactions with this contract go through this function. A caller that calls `unlock` must implement
```
snreynolds, line 110: This is actually not true - Initialize and updateDynamicFee can be called without the unlockCallback.

src/interfaces/IPoolManager.sol
```diff
@@ -102,7 +107,8 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload, IExttload {
...
+    /// `IUnlockCallback(msg.sender).unlockCallback(data)`, where they interact with the remaining functions on this contract.
```
snreynolds, line 111:
```suggestion
    /// @notice All interactions on the contract that account deltas require unlocking. A caller that calls `unlock` must implement
    /// `IUnlockCallback(msg.sender).unlockCallback(data)`, where they interact with the remaining functions on this contract.
    /// @dev The only functions callable without an unlocking are `initialize` and `updateDynamicLPFee`
```

src/interfaces/IPoolManager.sol
```diff
@@ -126,7 +132,7 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload, IExttload {
     /// @return feeDelta The balance delta of the fees generated in the liquidity range. Returned for informational purposes.
     function modifyLiquidity(PoolKey memory key, ModifyLiquidityParams memory params, bytes calldata hookData)
         external
-        returns (BalanceDelta, BalanceDelta);
+        returns (BalanceDelta callerDelta, BalanceDelta feeDelta);
```
snreynolds, line 135: I really dont like this name, but we can fix in a different PR. I have an open issue on core tracking conventional names here.

src/libraries/Pool.sol
```diff
@@ -307,12 +307,12 @@ library Pool {
...
                 : slot0Start.lpFee();
 
-            swapFee = protocolFee == 0 ? lpFee : uint24(protocolFee).calculateSwapFee(lpFee);
+            swapFee = protocolFee == 0 ? lpFee : uint16(protocolFee).calculateSwapFee(lpFee);
         }
 
         bool exactInput = params.amountSpecified < 0;
 
-        if (!exactInput && (swapFee == LPFeeLibrary.MAX_LP_FEE)) {
+        if (swapFee == LPFeeLibrary.MAX_LP_FEE && !exactInput) {
```
snreynolds, line 315: nice, so just early skip in most cases

src/libraries/Pool.sol
```diff
@@ -191,7 +191,7 @@ library Pool {
                 (uint256 feesOwed0, uint256 feesOwed1) =
                     position.update(liquidityDelta, feeGrowthInside0X128, feeGrowthInside1X128);
 
-                // Fees earned from LPing are added to the user's currency delta.
+                // Fees earned from LPing are calculated, and returned
```
snreynolds, line 194: Can we still comment somewhere that fees will be added to the final delta or when we do the addition, comment why fees get added

src/types/Slot0.sol
```diff
@@ -21,7 +21,7 @@ pragma solidity ^0.8.20;
  * the protocolFee is taken from the input first, then the lpFee is taken from the remaining input
  * uint24 protocolFee;
  *
- * Used for the lp fee, either static at initialize or dynamic via hook
+ * The current LP fee of the pool. If the pool is dynamic, this does not include the dynamic fee flag.
```
snreynolds, line 24: unrelated but why does MASK_160_BITS have leading 0s unlike MASK_24_BITS ?

snreynolds (changes requested):
Verifying that this PR closes the following issues from ABDK:
- 24, 25, 26, 27, 29, 30, 39, 161, 162, 163, 143, 141, 157, 158, 116, 151, 82, 42, 56, 40, 61, 62

& 1 question:
Why did you implement and revert CVF92?

hensha256:
>Why did you implement and revert CVF92?
Because it doesn't compile due to the fact that the constants are referrenced in assembly 😆 i stupidly didnt check it lol

src/libraries/Pool.sol
hensha256: yeah saved a good amount of gas

src/types/Slot0.sol
hensha256: Because forge complains. A 160 bit flag of 0xfff... looks like an address so forge gets confused. The extra 00 just stops that.

src/libraries/Pool.sol
hensha256: adding a comment where they get added together

src/libraries/BitMath.sol
hensha256: the input is a uint256, so the maximum possible "most significant bit" is in index 255 (because we index from 0). The largest number that can be held in a uint8 is 255 - so theres enough bits of response

hensha256: The only difference is that solady returns 256 if no bits are set (which wouldnt fit in a uint8). However we check that in advance by having `require(x>0)`

src/libraries/SwapMath.sol
hensha256: We already have the following tests:
- `test_fuzz_calculateSwapFee` ensures that the swap fee never exceeds max
- `test_updateDynamicLPFee_beforeSwap_failsWithTooLargeFee` checks that updateDynamicLPFee never allows a fee>max
- `test_fuzz_dynamicReturnSwapFee` checks that the beforeSwap hook's returned dynamic fee cannot be too large
- `test_fuzz_initialize` checks that no non-dynamic fee pool can be initialized with a fee larger than max
- `test_fuzz_getInitialLPFee` checks that the initial LP fee for a pool is never larger than max

hensha256: i've added the comment to make it clearer, but let me know if theres another test you think is missing and i can code one!

snreynolds: perfect! those tests sound like they cover this

snreynolds: & comment looks good

snreynolds:
> > Why did you implement and revert CVF92?
> 
> Because it doesn't compile due to the fact that the constants are referrenced in assembly 😆 i stupidly didnt check it lol

Ah I see - so we are marking that as will not fix then yes?

snreynolds (approved):
approved for the abdk issues referenced

src/libraries/BitMath.sol
Vectorized: This is not De Brujin, but rather a custom lookup with numbers found via z3.

PR #786, opened 2024-07-13
Certora I01 - Add recipient argument to settle...

gretzke:
in order to pay on behalf of another address


src/test/ActionsRouter.sol
```diff
@@ -52,11 +53,13 @@ contract ActionsRouter is IUnlockCallback, Test {
...
             bytes memory param = params[i];
             if (action == Actions.SETTLE) {
-                _settle();
+                _settle(param);
             } else if (action == Actions.SETTLE_NATIVE) {
                 _settleNative(param);
             } else if (action == Actions.TAKE) {
                 _take(param);
+            } else if (action == Actions.TAKE_FROM) {
+                _takeFrom(param);
```
hensha256, line 62: i thought for a second youd really added a parameter to take to let you take from someone else 😅 can you rename like `PRANK_TAKE_FROM` or something?

test/Sync.t.sol
```diff
@@ -168,7 +198,7 @@ contract SyncTest is Test, Deployers, GasSnapshot {
         params[6] = abi.encode(currency0, address(this), manager, 10);
 
         actions[7] = Actions.SETTLE;
-        params[7] = abi.encode(currency0);
+        params[7] = abi.encode(address(router));
```
hensha256, line 201: mmmmm feels kind of annoying that you have to figure out who the msg.sender's address is going to be and pass it in? what if theres a scenario where yu dont know who's going to be executing your trade?
i feel like it would be nice to have a default of msg.sender somehow. Like if someone passes in address(0), default of msg.sender? Or have a separate function with no parameters.

.forge-snapshots/simple swap.snap
```diff
@@ -1 +1 @@
-124313
\ No newline at end of file
+124651
```
hensha256, line 1: !!!!!!! 350!! this makes me want 2 separate functions more 👀

gretzke: do you want to overload `settle()` and `settle(address)`?

test/Sync.t.sol
gretzke: added `settleFrom(address)`

test/Sync.t.sol
```diff
@@ -88,6 +88,35 @@ contract SyncTest is Test, Deployers, GasSnapshot {
...
+        actions[2] = Actions.SETTLE_FOR;
+        params[2] = abi.encode(recipient);
+
+        actions[3] = Actions.ASSERT_DELTA_EQUALS;
+        params[3] = abi.encode(currency2, recipient, amount);
+
+        actions[4] = Actions.ASSERT_DELTA_EQUALS;
+        params[4] = abi.encode(currency2, address(router), 0);
+
+        actions[5] = Actions.PRANK_TAKE_FROM;
```
snreynolds, line 114: I'm a little confused what this is testing.. 

We are encoding here (currency2, recipient, address(this), amount)

but then inside `_prankTakeFrom` we are decoding (currency, from, recipient, amount)

Can you add another assert after this call for what you expect to happen?

snreynolds (changes requested):

test/Sync.t.sol
gretzke: with settleFor we are settling on behalf of a third party. In order to be able to close out the delta, I'm pranking from the address that is the recipient in the settleFor call in order to successfully close out the deltas. Thus, the functionality is also implicitly tested, because the deltas are closed

hensha256: yeah +1 to what sara said. The command decodes
```
(Currency currency, address from, address recipient, uint256 amount) =
            abi.decode(params, (Currency, address, address, uint256));
```

But this call has the recipient second not third. Please can you check it and add checks 🙏

gretzke: no this is correct, the recipient is the settleFor recipient, the account that receives the positive delta during the settleFor call. So that's the account we need to prank (from). The recipient in take is the recipient of the funds

snreynolds: I think I see whats happening and IMO its just confusing because the params are the same names for different contexts. I think what you are testing is.. apply a positive delta for recipient, then prank that recipient and send it to somewhere else.

I think what I would prefer to see tested is pranking another address such that there is a negative delta open on that address then someone else transfers their own funds, and calls settleFor with recipient to pay off their negative delta

snreynolds (approved):

PR #789, opened 2024-07-15
ABDK: Custom error params

hensha256:
Adds params to custom errors

CVF 153, 156, 91, 112, 125


gretzke (approved):

gretzke:
Generally speaking I wouldn't necessarily add a parameter to the error that is passed in as a parameter to the function (e.g., new protocol fee). But because we didn't really increase gas, I'm fine with it

src/libraries/TickMath.sol
```diff
@@ -1,19 +1,24 @@
...
-    error InvalidTick();
+    error InvalidTick(int24 tick);
     /// @notice Thrown when the price passed to #getTickAtSqrtPrice does not correspond to a price between MIN_TICK and MAX_TICK
     error InvalidSqrtPrice();
 
     /// @dev The minimum tick that may be passed to #getSqrtPriceAtTick computed from log base 1.0001 of 2**-128
     int24 internal constant MIN_TICK = -887272;
     /// @dev The maximum tick that may be passed to #getSqrtPriceAtTick computed from log base 1.0001 of 2**128
     int24 internal constant MAX_TICK = 887272;
+    uint256 internal constant MAX_TICK_UINT256 = 887272;
```
snreynolds, line 21: why can't you just cast MAX_TICK instead of writing another constant?

src/PoolManager.sol
```diff
@@ -124,9 +124,13 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
...
     {
         // see TickBitmap.sol for overflow conditions that can arise from tick spacing being too large
-        if (key.tickSpacing > MAX_TICK_SPACING) TickSpacingTooLarge.selector.revertWith();
-        if (key.tickSpacing < MIN_TICK_SPACING) TickSpacingTooSmall.selector.revertWith();
-        if (key.currency0 >= key.currency1) CurrenciesOutOfOrderOrEqual.selector.revertWith();
+        if (key.tickSpacing > MAX_TICK_SPACING) TickSpacingTooLarge.selector.revertWith(key.tickSpacing);
+        if (key.tickSpacing < MIN_TICK_SPACING) TickSpacingTooSmall.selector.revertWith(key.tickSpacing);
+        if (key.currency0 >= key.currency1) {
+            CurrenciesOutOfOrderOrEqual.selector.revertWith(
+                Currency.unwrap(key.currency0), Currency.unwrap(key.currency1)
```
snreynolds, line 131: How is this passing? The `CurrenciesOutOfOrderOrEqual` selector defined in IPoolManager seems unchanged

hensha256: good catch - its passing because we do low level reverts to save gas. fixing it now :)

src/libraries/TickMath.sol
```diff
@@ -1,19 +1,24 @@
...
 /// @notice Computes sqrt price for ticks of size 1.0001, i.e. sqrt(1.0001^tick) as fixed point Q64.96 numbers. Supports
 /// prices between 2**-128 and 2**128
 library TickMath {
+    using CustomRevert for bytes4;
+
     /// @notice Thrown when the tick passed to #getSqrtPriceAtTick is not between MIN_TICK and MAX_TICK
-    error InvalidTick();
+    error InvalidTick(int24 tick);
     /// @notice Thrown when the price passed to #getTickAtSqrtPrice does not correspond to a price between MIN_TICK and MAX_TICK
     error InvalidSqrtPrice();
```
snreynolds, line 15: Why not also have a param here to be consistent with InvalidTick?

src/PoolManager.sol
snreynolds: oh i see, its the same selector. still i guess good to update in the interface

src/libraries/TickMath.sol
hensha256: thought it would increase gas but it didnt. Have reverted that

src/libraries/TickMath.sol
hensha256: oops this is another one like the currency one where i just didnt push the change to this definition, pushing

snreynolds (approved):

PR #790, opened 2024-07-15
bubble-up errors from tokens and hooks

hensha256:
When we interact with tokens or hooks information gets lost. Now
- if the contract reverts with a reason, that gets bubbled up
- otherwise we use our custom error


test/types/Currency.t.sol
```diff
@@ -140,7 +150,8 @@ contract TestCurrency is Test {
                 MockERC20(Currency.unwrap(erc20Currency)).balanceOf(address(currencyTest)), initialERC20Balance - amount
             );
         } else {
-            vm.expectRevert(CurrencyLibrary.ERC20TransferFailed.selector);
```
snreynolds, line 143: Could we bubble up the return data with our own selector as well? 

so it instead would be ERC20TransferFailed(bytes returnData)

hensha256: We do that with universal router and it means that no browsers like etherscan can parse what the return reason was. So instead of getting the token's error like "Transfer overflow" or something. You get 
`ERC20TransferFailed(hexEncodingOf("transfer overflow"))`
So just makes it impossible to read sadly.

snreynolds: Sadly I feel like that is easier for devs to decode though.. because we'd know 2 things:
- where in OUR code its failing
- where in the external code its failing

snreynolds (approved):
looks great

PR #791, opened 2024-07-15
ABDK-147 return bytes32 array

gretzke:
## Description of changes

return bytes32 array from consecutive extsload instead of bytes array


src/libraries/StateLibrary.sol
```diff
@@ -136,7 +136,7 @@ library StateLibrary {
         bytes32 slot = _getTickInfoSlot(poolId, tick);
 
         // offset by 1 word, since the first word is liquidityGross + liquidityNet
-        bytes memory data = manager.extsload(bytes32(uint256(slot) + 1), 2);
+        bytes32[] memory data = manager.extsload(bytes32(uint256(slot) + 1), 2);
         assembly ("memory-safe") {
             feeGrowthOutside0X128 := mload(add(data, 32))
```
hensha256, line 141: might be better to just have
`return data[0], data[1]`
?

hensha256: same for feeGrowthGlobal

hensha256 (approved):
lgtm!

PR #792, opened 2024-07-15
feat: add clear

snreynolds:
## Related Issue

## Description of changes

This adds a new function `clear`, which decreases any positive delta on the contract without a respective token transfer. It can be used in cases where there is dust left in the contract and it is more expensive to transfer the dust than to forfeit it to the manager.

To ensure integrators are aware of the amount being cleared I require that the amount is passed in, and is exactly the amount of the open positive delta.


test/PoolManager.clear.t.sol
```diff
@@ -0,0 +1,191 @@
...
+        actions[3] = Actions.ASSERT_DELTA_EQUALS;
+        params[3] = abi.encode(currency0, address(actionsRouter), 0);
+
+        actions[4] = Actions.ASSERT_NONZERO_DELTA_COUNT_EQUALS;
+        params[4] = abi.encode(0);
+
+        actionsRouter.executeActions(actions, params);
+    }
+
+    function test_clear_reverts_zeroDelta_inputNonZero() external {
```
snreynolds, line 137: Can I get some eyes on the behavior for this case? If the open delta is 0 and you call clear with anything nonzero it reverts with `MustClearExactBalance`.

src/PoolManager.sol
```diff
@@ -294,6 +294,14 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
         _accountDelta(currency, paid.toInt128(), msg.sender);
     }
 
+    /// @inheritdoc IPoolManager
+    function clear(Currency currency, uint256 amount) external onlyWhenUnlocked {
+        int256 current = currency.getDelta(msg.sender);
+        if (current < 0) revert OwedFundsCannotBeCleared();
+        if (amount != uint256(current)) revert MustClearExactBalance();
```
hensha256, line 301: These 2 lines feel inefficient. It checks that current isnt negative, and then it requires it be equal to `amount`.

If we know that `amount` is positive (it has to be as its a uint256), then we know that if `amount==current` then current cannot be <0 either.

I think it would be more efficient to just do

```
        int256 current = currency.getDelta(msg.sender);
        int128 amountDelta = amount.toInt128();
        if (amountDelta != current) revert MustClearExactBalance();
        _accountDelta(currency, -amountDelta, msg.sender);
```

src/interfaces/IPoolManager.sol
```diff
@@ -151,19 +152,29 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload, IExttload {
...
+    function sync(Currency currency) external returns (uint256 balance);
+
     /// @notice Called by the user to net out some value owed to the user
     /// @dev Can also be used as a mechanism for _free_ flash loans
     function take(Currency currency, address to, uint256 amount) external;
 
+    /// @notice Called by the user to pay what is owed
+    function settle(Currency currency) external payable returns (uint256 paid);
+
+    /// @notice Called to clear a positive balance without a corresponding transfer.
```
hensha256, line 167: Honestly i think this needs to be more severe in its warning. Imo it needs to state
```
Any currency that is cleared, will be non-retreivable, and locked in the contract permanently
```

Or something like that. I wont want anyone ever using this function thinking theyre clearing a debt or something by accident.

src/PoolManager.sol
gretzke, line 301: this basically only saves gas on ERC20 transfers then? because you still need to check how much is owed exactly to clear

hensha256 (commented):
couple of thoughts but generally looks great! i feel like 1 simple gas test might help us figure out my first comment though? You could just add a `bool` to the clear parameters in the actions router and then have
```

    function _clear(bytes memory params) internal {
        (Currency currency, uint256 amount, bool measureGas) = abi.decode(params, (Currency, uint256, bool));
        if measureGas snapStart()
        manager.clear(currency, amount);
        if measureGas snapEnd()
    }
```

src/PoolManager.sol
hensha256: yeah but the difference between a tload (this check) and a transfer (which is 2 sstores) is huge

src/PoolManager.sol
```diff
@@ -294,6 +294,14 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
         _accountDelta(currency, paid.toInt128(), msg.sender);
     }
 
+    /// @inheritdoc IPoolManager
+    function clear(Currency currency, uint256 amount) external onlyWhenUnlocked {
+        int256 current = currency.getDelta(msg.sender);
+        int128 amountDelta = amount.toInt128();
```
hensha256, line 300: maybe a quick comment that because amount was positive, only positive deltas can be cleared?

src/interfaces/IPoolManager.sol
```diff
@@ -151,19 +149,30 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload, IExttload {
...
+
     /// @notice Called by the user to net out some value owed to the user
     /// @dev Can also be used as a mechanism for _free_ flash loans
     function take(Currency currency, address to, uint256 amount) external;
 
+    /// @notice Called by the user to pay what is owed
+    function settle(Currency currency) external payable returns (uint256 paid);
+
+    /// @notice WARNING - Any currency that is cleared, will be non-retreivable, and locked in the contract permanently.
+    /// A callt to clear will zero out a positive balance WITHOUT a corresponding transfer.
```
hensha256, line 165:
```suggestion
    /// A call to clear will zero out a positive balance WITHOUT a corresponding transfer.
```

src/PoolManager.sol
```diff
@@ -294,6 +294,14 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
...
+        if (amountDelta != current) revert MustClearExactBalance();
```
hensha256, line 301: maybe `MustClearExactPositiveDelta`? So that when its thrown in the 0-case or the negative case it still sounds right

src/PoolManager.sol
```diff
@@ -297,8 +297,9 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     /// @inheritdoc IPoolManager
     function clear(Currency currency, uint256 amount) external onlyWhenUnlocked {
         int256 current = currency.getDelta(msg.sender);
+        // Because input is `uint256`, only positive amounts can be cleared.
         int128 amountDelta = amount.toInt128();
-        if (amountDelta != current) revert MustClearExactBalance();
+        if (amountDelta != current) revert MustClearExactPositiveDelta();
```
hensha256, line 302: oh this should be `MustClearExactPositiveDelta.revertWith()`, much cheaper reverts :)

hensha256 (dismissed):

src/PoolManager.sol
snreynolds: ah good catch!

hensha256 (approved):

PR #793, opened 2024-07-15
ABDK CVF 44

hensha256:
CVF 44


snreynolds (commented):
why dont we also need the length checks in the donate/initialize code paths?

snreynolds (approved):
i think that makes sense

PR #798, opened 2024-07-16
abdk cvf9

dianakocsis:
## Related Issue
abdk cvf9

## Description of changes
signextend


hensha256 (commented):
coming back here - why would this need sign extending if its then going to be shifted 128 to the left? its already 128 bits so that doesnt make sense to me. Imo this isnt needed and we can close this

hensha256 (approved):

PR #799, opened 2024-07-17
Rename currency settler util contract

hensha256:
## Related Issue

This contract is not secure and is just a test util. Lets name it as such, and put a better version in periphery for real contracts.

### DIFF
--- a/test/utils/CurrencySettlerTestUtil.sol
+++ b/test/utils/CurrencySettlerTestUtil.sol
@@ -9,7 +9,9 @@ import {IPoolManager} from "../../src/interfaces/IPoolManager.sol";
 /// To settle a positive delta (a credit to the user), a user may take or mint.
 /// To settle a negative delta (a debt on the user), a user make transfer or burn to pay off a debt.
 /// @dev Note that sync() is called before any erc-20 transfer in `settle`.
-library CurrencySettler {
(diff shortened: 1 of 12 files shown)


test/utils/CurrencySettlerTestUtil.sol
```diff
@@ -9,7 +9,9 @@ import {IPoolManager} from "../../src/interfaces/IPoolManager.sol";
 /// To settle a positive delta (a credit to the user), a user may take or mint.
 /// To settle a negative delta (a debt on the user), a user make transfer or burn to pay off a debt.
 /// @dev Note that sync() is called before any erc-20 transfer in `settle`.
-library CurrencySettler {
+/// @dev THIS CONTRACT IS NOT SECURE, IT IS FOR TESTING PURPOSES, DO NOT USE IT IN PROD
```
snreynolds, line 12: I feel like this warning is a bit overkill? It just doesn't use safe transfer from and we don't do this in a lot of our test contracts which should never be used in prod

snreynolds: I think just renaming it and putting it in the test/utils file is sufficient

snreynolds (changes requested):

PR #803, opened 2024-07-21
calculate positionKey with a helper function

snreynolds:
## Related Issue

PositionManager often needs to calculate the positionKey (positionId). Instead of rewriting code, we can expose this calculation in core. Likely other integrations will want something similar.

### DIFF
--- a/src/libraries/Position.sol
+++ b/src/libraries/Position.sol
@@ -37,8 +37,20 @@ library Position {
         returns (Info storage position)
     {
         // positionKey = keccak256(abi.encodePacked(owner, tickLower, tickUpper, salt))
-        bytes32 positionKey;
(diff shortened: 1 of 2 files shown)


src/libraries/Position.sol
```diff
@@ -37,8 +37,20 @@ library Position {
         returns (Info storage position)
     {
         // positionKey = keccak256(abi.encodePacked(owner, tickLower, tickUpper, salt))
```
saucepoint, line 39: can move this comment to calculatePositionKey's natspec?

saucepoint (approved):
is it useful to add a test for this helper function? i.e. one that asserts `calculatePositionKey(...) == keccak256(abi.encode(...))`

hensha256 (dismissed):

saucepoint (approved):

PR #805, opened 2024-07-23
update StateLibrary.sol

snreynolds:
updates StateLibrary.sol with :
- added natspec on getPosition (which is now getPositionInfo)
- overloads getPositionInfo
- unifies return types on the overloaded function


hensha256 (approved):

test/ModifyLiquidity.t.sol
```diff
@@ -263,45 +263,45 @@ contract ModifyLiquidityTest is Test, Logger, Deployers, JavascriptFfi, Fuzzers,
...
 
-        Position.Info memory updatedPositionSalt = manager.getPosition(
+        (uint128 updatedLiquiditySalt,,) = manager.getPositionInfo(
             simplePoolId, address(modifyLiquidityRouter), LIQ_PARAM_SALT.tickLower, LIQ_PARAM_SALT.tickUpper, SALT
         );
 
-        Position.Info memory updatedPositionSalt2 = manager.getPosition(
-            simplePoolId, address(modifyLiquidityRouter2), LIQ_PARAM_SALT.tickLower, LIQ_PARAM_SALT.tickUpper, SALT
+        (uint128 updatedLiquiditySalt2,,) = manager.getPositionInfo(
+            simplePoolId, address(modifyLiquidityRouter2), LIQ_PARAM_SALT_2.tickLower, LIQ_PARAM_SALT_2.tickUpper, SALT
```
hensha256, line 285: was this just an incorrect test before? why was it passing?

snreynolds: they are actually the same params so I do think it was a mistake from before but the test itself was still "correct"

PR #807, opened 2024-07-24
fix natspec

snreynolds:
Related to v4-periphery comments on the implementing lens contract for StateLibrary

Wanted to make sure the natspec was updated accordingly here 

https://github.com/Uniswap/v4-periphery/pull/187/commits/e2ebb94cf742bc216aad8b1eeae86392c5759b96


src/libraries/StateLibrary.sol
```diff
@@ -267,7 +267,7 @@ library StateLibrary {
 
     /**
      * @notice Retrieves the liquidity of a position.
-     * @dev Corresponds to pools[poolId].positions[positionId].liquidity. A more gas efficient version of getPositionInfo
+     * @dev Corresponds to pools[poolId].positions[positionId].liquidity. More gas efficient for just retrieiving liquidity as compared to getPositionInfo
```
hensha256, line 270: Just missing the `calulcate` change on line 216

snreynolds: ah oops good catch!

hensha256 (approved):

PR #808, opened 2024-07-24
fix: add sqrtPriceX96 and tick to initialize event

snreynolds:
## Related Issue

Its useful to have the starting price and starting tick of pools when indexing. This adds those fields on the initialize event.


matteenm (approved):

.forge-snapshots/poolManager bytecode size.snap
```diff
@@ -1 +1 @@
-24494
\ No newline at end of file
+24476
```
hensha256, line 1: wut

hensha256 (approved):

.forge-snapshots/poolManager bytecode size.snap
snreynolds: ya i know i saw that... weird

wjmelements, line 1: doesn't this number have to be below 24k?

PR #809, opened 2024-07-24
M02 tick price discrepancy

ewilz:
## Related Issue


ewilz:
I feel like this is actually bugged. I think we need a zeroForOne check on line 399

src/libraries/Pool.sol
```diff
@@ -411,14 +411,8 @@ library Pool {
...
 
-                // Equivalent to `state.tick = zeroForOne ? step.tickNext - 1 : step.tickNext;`
                 unchecked {
-                    // cannot cast a bool to an int24 in Solidity
-                    int24 _zeroForOne;
-                    assembly ("memory-safe") {
-                        _zeroForOne := and(zeroForOne, 0xff)
-                    }
-                    state.tick = step.tickNext - _zeroForOne;
+                    state.tick = (zeroForOne && state.amountSpecifiedRemaining != 0) ? step.tickNext - 1 : step.tickNext;
```
wjmelements, line 415: I think this is wrong. My understanding of this change is that the tick is not crossed when the amount specified ends exactly at the tick crossing. Here is the changed behavior:

| zeroForOne | remaining == 0 | state.tick |
| :-: | :-: | :-: |
| false | false | step.tickNext |
| false | true | step.tickNext |
| true | false | step.tickNext - 1 |
| true | true | step.tickNext |

I think, after the change to the tick crossing condition above, the condition should be (zeroForOne xor amountRemaining)
```
state.tick = (zeroForOne != (state.amountSpecifiedRemaining != 0) ? step.tickNext : step.tickNext - 1;
```

This makes it:

| zeroForOne | remaining == 0 | state.tick |
| :-: | :-: | :-: |
| false | false | step.tickNext |
| false | true | **step.tickNext - 1** |
| true | false | step.tickNext - 1 |
| true | true | step.tickNext |


The difference would be that, for zeroForOne false (one for zero; tick is increasing) and amount remaining zero, we wouldn't have crossed the tick.

wjmelements: https://github.com/Uniswap/v4-core/pull/809#issuecomment-2251103271 is an alternative fix. Either would work. I'm not sure what the intent of this change is though because there is no description. I only know that lines 415 and 399 are inconsistent.

wjmelements: (My preference is for the behavior to match UniswapV3.)

shuhuiluo: Also wonder what the rationale is. If I remember correctly, KyberSwap messed up badly on the tick transition.

ewilz: Formal verification failed because if we're trading left, we cross the tick if the price has landed right on the tick (even though the tick is technically "inclusive" to the right). V3 allowed for this small error. Not crossing it would be "more correct" but the error is so microscopic that it's really a non-issue. I don't think we've decided whether this will get merged or not, and we need one more check via my last comment.

wjmelements: @ewilz 

Your formal verification likely was given a wrong predicate. The prior behavior is sane in its own way.

In slot0, the price indicates the precise price position and the tick indicates the position with respect to the liquidity crossings. If, for ending at exact tick price, the tick is always crossed in one direction but not in the other, then the tick would be redundant with price and could be removed from slot0. Because both are defined there is a degree of freedom I presume to be intentional.

Still concerning this edge case:

* xor makes so the tick is not crossed in either direction. This has minor gas advantages if the next swap is in the opposite direction.
* zeroForOne makes so the tick is crossed in only one direction. This has the advantage of making tick redundant and allowing its removal from slot0, which likely saves the most gas.
* uniswap v3 and the prior behavior always crosses the tick. This is perhaps the worst design by gas but it is consistent with the battle-tested Uniswap V3.

The current commit of this PR is inconsistent. It picks the first approach on line 399 and the second on line 415. It seems you favor the second. I hope you will check whether it saves gas to remove tick from slot0 if you pursue this approach.

ewilz: there are technically sqrtPrice's between ticks, and you're right that there can be some degree of error, the way this code is currently on v3  & main as proof. Plus v3 has all kinds of rounding errors anyway.

- if we're getting exact though, the tick should only be crossed going right when landed on, not to the left since it's inclusive to the right. not crossing in either direction is just creating the problem in the opposite direction, this is a pass at making things exact due to some formal verification feedback.

Anyway this was feedback from an audit, we're researching it, and not necessarily committing it. In fact I'm probably in favor of not merging due to gas and triviality.

wjmelements: If the actual issue is donate, I don't think it really matters which way it goes, except at MIN_TICK, which can currently be crossed downward ("left").

wjmelements: slightly related discussion https://github.com/Uniswap/v4-core/issues/812#issuecomment-2263934980

wjmelements: > If the actual issue is donate, I don't think it really matters which way it goes, except at MIN_TICK, which can currently be crossed downward ("left").

What happens in this case is that liquidity is zero, and then donation is disallowed.
```
        uint128 liquidity = state.liquidity;
        if (liquidity == 0) NoLiquidityToReceiveFees.selector.revertWith();
```

wjmelements: Line 399 has been updated. My concern has been addressed. Can click resolve.

hensha256:
ty for your work 🫡

PR #810, opened 2024-07-25
rm public Tick Spacing limits

wjmelements:
Reviewer @hensha256 @snreynolds
There are two reasons to remove unnecessary methods.
1. to save gas on all methods by reducing the size of the 4byte abi lookup table.
2. to reduce codesize
#### Changes
Remove public constant methods MIN_TICK_SPACING and MAX_TICK_SPACING


.forge-snapshots/poolManager bytecode size.snap
```diff
@@ -1 +1 @@
-24476
\ No newline at end of file
+24339
```
wjmelements, line 1: decent progress, but still above 24k

wjmelements:
These constants have been public since https://github.com/Uniswap/v4-core/pull/71. I don't think there is a good reason for them to be public.

wjmelements:
The test failure is the same codecoverage 403 as before. I've been told it is fine and expected.

.forge-snapshots/poolManager bytecode size.snap
wjmelements: This saves 137 bytes. For comparison, using the internal function `checkNotDelegateCall` for `noDelegateCall` saves 149 bytes.

snreynolds (commented):
Nice! I like it, tagging @hensha256 also for a review

.forge-snapshots/addLiquidity CA fee.snap
```diff
@@ -1 +1 @@
-170956
\ No newline at end of file
+170604
```
hensha256, line 1: wowww nice

hensha256 (dismissed):

hensha256:
Hey @wjmelements ! Thanks for this! Please can you sign your commits so we can merge it?

wjmelements:
> Please can you sign your commits so we can merge it?

That wasn't necessary previously. Can it be skipped? Otherwise I'd have to force-push amended commits.

wjmelements:
ok I have force pushed a signed commit

wjmelements:
@hensha256 please merge

hensha256 (approved):

PR #817, opened 2024-08-05
add parameter to bubbled revert wrapper

gretzke:
## Related Issue


src/libraries/CustomRevert.sol
```diff
@@ -75,20 +75,21 @@ library CustomRevert {
     }
 
     /// @notice bubble up the revert message returned by a call and revert with the selector provided
-    /// @dev this function should only be used with custom errors of the type `CustomError(bytes revertReason)`
-    function bubbleUpAndRevertWith(bytes4 selector) internal pure {
+    /// @dev this function should only be used with custom errors of the type `CustomError(address contract_, bytes revertReason)`
```
snreynolds, line 78:
```suggestion
    /// @dev this function should only be used with custom errors of the type `CustomError(address contract, bytes revertReason)`
```

snreynolds (commented):
lgtm but I would like @hensha256 to review

src/libraries/CustomRevert.sol
```diff
@@ -75,20 +75,21 @@ library CustomRevert {
...
+    function bubbleUpAndRevertWith(bytes4 selector, address addr) internal pure {
         assembly ("memory-safe") {
             let size := returndatasize()
             let fmp := mload(0x40)
 
             // Encode selector, offset, size, data
```
hensha256, line 84: this needs updating and maybe a little more detail now
```
// Encode selector, address, data offset, data size, data
```

hensha256 (commented):
one nit

src/types/Currency.sol
```diff
@@ -31,12 +31,12 @@ library CurrencyLibrary {
...
-    /// @param revertReason bubbled up revert reason
-    error NativeTransferFailed(bytes revertReason);
+    /// @param reason bubbled up revert reason
+    error Wrap__NativeTransferFailed(address recipient, bytes reason);
 
     /// @notice Thrown when an ERC20 transfer fails
-    /// @param revertReason bubbled up revert reason
-    error ERC20TransferFailed(bytes revertReason);
+    /// @param reason bubbled up revert reason
+    error Wrap__ERC20TransferFailed(address token, address recipient, bytes reason);
```
hensha256, line 39: is this right? looks like its only passing in 1 address not two

src/libraries/CustomRevert.sol
gretzke: but `contract` is a reserved keyword 😄

src/types/Currency.sol
gretzke: ugh, thanks for catching it!

src/libraries/CustomRevert.sol
```diff
@@ -75,13 +75,13 @@ library CustomRevert {
     }
 
     /// @notice bubble up the revert message returned by a call and revert with the selector provided
-    /// @dev this function should only be used with custom errors of the type `CustomError(address contract_, bytes revertReason)`
+    /// @dev this function should only be used with custom errors of the type `CustomError(address contract, bytes revertReason)`
```
hensha256, line 78: i actually feel like `contract` is wrong? for our `Wrap__NativeTransferFailed` the `recipient` thats emitted might not be a contract. Maybe `target` is better? And that avoids the contract keyword too

gretzke: target is good, I like that

snreynolds: i like target!

PR #818, opened 2024-08-05
add salt to modLiq event

snreynolds:
## Related Issue


.forge-snapshots/addLiquidity CA fee.snap
```diff
@@ -1 +1 @@
-170646
\ No newline at end of file
+170924
```
snreynolds, line 1: 😞

saucepoint (approved):

PR #820, opened 2024-08-09
chore: pin-to-sha-hash

mr-uniswap:
## Related Issue

[Which issue does this pull request resolve?](https://linear.app/uniswap/issue/PROTO-412/pin-to-sha-hash-instead-of-version-for-github-action)

## Description of changes
Simply pins the version to a hash rather than a version number


Fendyzainol9:
Alhamdulillah subhanallah

PR #750, opened 2024-06-21
One Sixty

wjmelements:
#### Changes
* `32 * 5 = 160`

PR #824, opened 2024-08-13
Spearbit gas 03: `liquidityGrossBefore`

hensha256:
Doesnt actually improve gas, but I do think its more readable

### DIFF
--- a/src/libraries/Pool.sol
+++ b/src/libraries/Pool.sol
@@ -515,12 +515,13 @@ library Pool {
 
         uint128 liquidityGrossBefore;
         int128 liquidityNetBefore;
+
         assembly ("memory-safe") {
             // load first slot of info which contains liquidityGross and liquidityNet packed
             // where the top 128 bits are liquidityNet and the bottom 128 bits are liquidityGross
             let liquidity := sload(info.slot)
             // slice off top 128 bits of liquidity (liquidityNet) to get just liquidityGross
-            liquidityGrossBefore := shr(128, shl(128, liquidity))
+            liquidityGrossBefore := and(liquidity, 0xffffffffffffffffffffffffffffffff)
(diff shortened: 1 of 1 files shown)


src/libraries/Pool.sol
```diff
@@ -515,12 +515,13 @@ library Pool {
...
         uint128 liquidityGrossBefore;
         int128 liquidityNetBefore;
+
         assembly ("memory-safe") {
             // load first slot of info which contains liquidityGross and liquidityNet packed
             // where the top 128 bits are liquidityNet and the bottom 128 bits are liquidityGross
             let liquidity := sload(info.slot)
             // slice off top 128 bits of liquidity (liquidityNet) to get just liquidityGross
-            liquidityGrossBefore := shr(128, shl(128, liquidity))
+            liquidityGrossBefore := and(liquidity, 0xffffffffffffffffffffffffffffffff)
```
wjmelements, line 524: I think the reason it doesn't save gas is the parameter stack ordering.

I suspect if you did
```
liquidityGrossBefore := and(0xffffffffffffffffffffffffffffffff, liquidity)
```
then it would save gas

dianakocsis (approved):
i agree it looks more readable
tested it with `liquidityGrossBefore := and(0xffffffffffffffffffffffffffffffff, liquidity)` and it was still the same gas

wjmelements:
> tested it with `liquidityGrossBefore := and(0xffffffffffffffffffffffffffffffff, liquidity)` and it was still the same gas

same. thanks for checking!

src/libraries/Pool.sol
wjmelements: another explanation for it being the same could be that solc's optimization was already doing this, similar to #750

wjmelements:
I think this PR is redundant after #827

PR #825, opened 2024-08-13
Spearbit gas-05

hensha256:
given we know the returndatasize is 32, we can use scratch space instead of using the free memory pointer


src/ProtocolFees.sol
```diff
@@ -85,11 +85,10 @@ abstract contract ProtocolFees is IProtocolFees, Owned {
                 success := and(success, eq(returndatasize(), 32))
 
                 // load the return data if success is true
+                // if the returndatasize != 32, success is false, so we can load 32 bytes
                 if success {
-                    let fmp := mload(0x40)
-                    returndatacopy(fmp, 0, returndatasize())
-                    returnData := mload(fmp)
-                    mstore(fmp, 0)
+                    returndatacopy(0, 0, 32)
```
wjmelements, line 90: you can eliminate the need to do `returndatacopy` entirely by changing the call above:
```
call(gasLimit, toAddress, 0, add(data, 0x20), mload(data), 0, 32)
```
then the returndata will already be in memory

snreynolds: We do not want to load the returnData if it is greater than 32 bytes though

snreynolds (dismissed):
this looks good to me

src/ProtocolFees.sol
wjmelements: Your response signals there is something you don't understand. Running a diagnostic:

* Do you understand that the `32` parameter limits the output size, so memory corruption is impossible?
* Do you understand that UniswapV1 returns more than 32 bytes for this `CALL` but that the answer is still in the first 32 bytes?
* What do you believe is the harm in loading 32 bytes into memory if `RETURNDATASIZE` exceeds 32?

hensha256: > Do you understand that UniswapV1 returns 

What relation is there between this code and uniswap v1?

wjmelements: > What relation is there between this code and uniswap v1?

I am mistaken on that point; this code is not related to ERC20. The only remaining relation is that UniswapV1 returns extra data for all `CALL` but the intended return is still at the beginning. The extra data doesn't necessarily signal incompatibility and in cases such as UniswapV1 it can be discarded.

But it still seems harmless to load the first 32 bytes unconditionally to avoid the extra `RETURNDATACOPY`.

src/ProtocolFees.sol
```diff
@@ -78,18 +78,13 @@ abstract contract ProtocolFees is IProtocolFees, Owned {
             bytes memory data = abi.encodeCall(IProtocolFeeController.protocolFeeForPool, (key));
             uint256 returnData;
             assembly ("memory-safe") {
-                success := call(gasLimit, toAddress, 0, add(data, 0x20), mload(data), 0, 0)
+                // only load the first 32 bytes of the return data to prevent gas griefing
+                success := call(gasLimit, toAddress, 0, add(data, 0x20), mload(data), 0, 32)
```
wjmelements, line 82: awesome!

wjmelements (approved):

src/ProtocolFees.sol
hensha256: yeah i agree with your suggestion :) just wanted to clarify your point about v1 as i thought i was missing something

take a look at the latest commit and lmk your thoughts!

hensha256: oh you already did 😆 fast lol

snreynolds: I _do_ understand what you're suggesting & I misread the suggestion originally - that happens! So if you can be patient as we review tons of code every day that would be greatly appreciated. 

Thanks for the suggestion!

snreynolds (approved):

src/ProtocolFees.sol
shuhuiluo:
```suggestion
                success := call(gasLimit, toAddress, 0, add(data, 0x20), mload(data), 0, 0x20)
```
nit: use hex for calldata/memory offsets for consistency.

PR #831, opened 2024-08-13
trail of bits i02

hensha256:


src/libraries/Pool.sol
```diff
@@ -309,7 +309,7 @@ library Pool {
 
         bool exactInput = params.amountSpecified < 0;
 
-        if (swapFee == LPFeeLibrary.MAX_LP_FEE && !exactInput) {
+        if (swapFee == SwapMath.MAX_FEE_PIPS && !exactInput) {
```
saucepoint, line 312: i prefer how it is currently

the two options read:

"if swap fee is equal to maximum LP fee" vs "if swap fee is equal to the fee pip denominator", respectively

hensha256: yeah trail of bits point was that "swapFee" is the combined protocol fee and LP fee. So checking the combined fee, against the max LP fee isnt logically correct. 
Logically you should check the combined fee against the maximum allowed combined fee.

hensha256: I'll rename the swap one i think to make it read better

PR #832, opened 2024-08-14
poc: Refactor TickBitmap Reusability

saucepoint:
## Related Issue

View Quoter https://github.com/Uniswap/v4-periphery/pull/296/files is re-implementing the TickBitmap.sol library

This is an experimental change to make the library reusable in periphery

## Description of changes

Overload `nextInitializedTickWithinOneWord` such that there is an implementation that does not depend on `PoolManager` storage mapping


src/libraries/TickBitmap.sol
```diff
@@ -76,26 +76,25 @@ library TickBitmap {
...
     /// @param tickSpacing The spacing between usable ticks
     /// @param lte Whether to search for the next initialized tick to the left (less than or equal to the starting tick)
     /// @return next The next initialized or uninitialized tick up to 256 ticks away from the current tick
     /// @return initialized Whether the next tick is initialized, as the function only searches within up to 256 ticks
     function nextInitializedTickWithinOneWord(
-        mapping(int16 => uint256) storage self,
-        int24 tick,
+        uint256 bitmap,
+        uint8 bitPos,
+        int24 compressed,
```
snreynolds, line 89: why do we want to expose `compressed` at all? seems like we can just always compute compressed inside this helper? and potentially makes this function more prone to integration issues if they dont know what the correct value is for `compressed`

snreynolds (commented):
Generally I like it & not opposed to making this change.

saucepoint:
closing since it introduces operational difficulties with audits and this is cosmetic change for periphery

(sad)

PR #834, opened 2024-08-20
sb14: use isAddressZero

snreynolds:
fixes spearbit informational issue 14


src/test/CurrencyTest.sol
```diff
@@ -17,7 +17,7 @@ contract CurrencyTest {
     }
 
     function isNative(Currency currency) external pure returns (bool) {
-        return currency.isNative();
+        return currency.isAddressZero();
     }
```
saucepoint, line 21: this test file and its unit tests are the only places where `isNative()` is used in any capacity, so we should remove it entirely

saucepoint (commented):
Not entirely opposed to the consolidation, but using isAddressZero feels less readable than isNative. simply because you need to remember that address(0) is special cased for native balance handling

i might wake up tomorrow and feel differently tho

src/test/CurrencyTest.sol
hensha256: `isNative` is used in the pool manager?

saucepoint: its not, pulled it locally and searched for it

snreynolds:
If you want a different name let me know but I don't think we should have two functions that do the same thing called different things

snreynolds:
Also I very much think that since C01 this makes it way easier to know that you are really dealing with an address that is addr0 and all other addresses will follow the other path. some native tokens are not the zero address

saucepoint (approved):
woke up today and changed my mind to be in-favor:

1. on CELO, defining a pool key becomes:
```
PoolKey({
    currency0: currency.ADDRESS_ZERO
    currency1: address(token)
    ...
})
```
Feels quite explicit that the pool is using "payable" tokens, and if we use the other naming CELO-mains might feel like `Currency.NATIVE` would be the addressed token

2. our use of `syncedCurrency == address(0)` to indicate if a currency has been synced or not is much more readable with `isAddressZero()`

src/types/Currency.sol
```diff
@@ -39,14 +39,14 @@ library CurrencyLibrary {
     error Wrap__ERC20TransferFailed(address token, bytes reason);
 
     /// @notice A constant to represent the native currency
-    Currency public constant NATIVE = Currency.wrap(address(0));
```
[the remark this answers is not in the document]
snreynolds: I also like Native fwiw but we wanted to consolidate the isNative() and isZero() paths and there was a an issue open zeppelin found that makes a good case for when isNative() can be confusing on chains where the native token is also an erc20 token (ie Celo).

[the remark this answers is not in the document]
snreynolds: If you have an alternative suggestion feel free to add it. We just don't want to have both isNative() and isZero() functions

src/PoolManager.sol
```diff
@@ -268,7 +268,7 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     /// @inheritdoc IPoolManager
     function sync(Currency currency) external {
         CurrencyReserves.requireNotSynced();
-        if (currency.isNative()) return;
+        if (currency.isAddressZero()) return;
```
hensha256, line 271: Comment saying address 0 is used for the native currency?

hensha256 (approved):

src/PoolManager.sol
```diff
@@ -268,7 +268,8 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     /// @inheritdoc IPoolManager
     function sync(Currency currency) external {
         CurrencyReserves.requireNotSynced();
-        if (currency.isNative()) return;
+        /// address(0) is used for the native currency
```
hensha256, line 271:
```suggestion
        // address(0) is used for the native currency
```

hensha256 (approved):

src/types/Currency.sol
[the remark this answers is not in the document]
wjmelements: My suggestion is to remove `isZero` instead of `isNative`. I understand there are chain-specific native token precompiles though. It will be slightly unclear either way, but not to me, so I am not a great judge of this.

PR #836, opened 2024-08-20
trail of bits i01

dianakocsis:
## Related Issue
Trail of Bits I-01

## Description of changes
ensure swap fee would never be greater than 100% (even though this would never happen because of ProtocolFeeLibrary.calculateSwapFee math), but just to be sure


src/libraries/Pool.sol
```diff
@@ -309,7 +309,7 @@ library Pool {
 
         bool exactInput = params.amountSpecified < 0;
 
-        if (swapFee == LPFeeLibrary.MAX_LP_FEE && !exactInput) {
+        if (swapFee >= LPFeeLibrary.MAX_LP_FEE && !exactInput) {
             InvalidFeeForExactOut.selector.revertWith();
         }
```
saucepoint, line 314: crazy man ask -- gas should be cheaper if we put `!exactInput` as the first check (short-circuiting), because 50% of the swaps are exactInput and we dont need to revert for >100% swap fees

saucepoint: wait nvm, we want to short circuit on the fee being too large, since its false like ~100% of the time

hensha256 (dismissed):

hensha256 (approved):

snreynolds (approved):

PR #839, opened 2024-08-22
spearbit L-03

dianakocsis:
## Related Issue
Spearbit L-03
https://github.com/spearbit-audits/review-uniswap-v4/issues/37

## Description of changes
emit event when the dynamic LP fee changes


dianakocsis:
Decided against emitting an event when the dynamic fee is updated.
This is because the override possibility for individual swaps would make it hard to track all of them off chain

PR #837, opened 2024-08-21
change variable of the fee amount going to the protocol

dianakocsis:
## Related Issue
Confusing to decipher between `feeForProtocol` and `protocolFee`

## Description of changes
change the name of `feeForProtocol` to `amountToProtocol`

PR #840, opened 2024-08-22
spearbit L-04

dianakocsis:
## Related Issue
Spearbit L-04
https://github.com/spearbit-audits/review-uniswap-v4/issues/18

## Description of changes
In this previous commit (https://github.com/Uniswap/v4-core/pull/837), I changed the name `feeForProtocol` to `amountToProtocol` to make it more clear that it is the actual amount, and to avoid confusion with `protocolFee` which is in pips. In this commit, I add a comment for more clarity before returning


snreynolds (commented):
there wasn't another name change? im forgetting what we discussed

dianakocsis:
we mentioned also changing `protocolFee` to `protocolFeePips`, but I kind of like how `lpFee`, `swapFee`, and `protocolFee` are all in pips. If we change it `protocolFeePips`, I think we should also do `lpFeePips` and `swapFeePips`

src/libraries/Pool.sol
```diff
@@ -313,6 +314,8 @@ library Pool {
             InvalidFeeForExactOut.selector.revertWith();
         }
 
+        // 0 is the amountToProtocol (the actual fee amount going to the protocol)
```
hensha256, line 317:
```suggestion
        // 0 is the fee amount that should be paid to the protocol
```

hensha256 (commented):
yeah i agree i'm happy not to make that change, i feel like the comments make it clear enough

PR #842, opened 2024-08-22
spearbit I-09

dianakocsis:
## Related Issue
Spearbit I-09
https://github.com/spearbit-audits/review-uniswap-v4/issues/17

## Description of changes
add comment saying it will revert for exact out case if swap fee is the max
also just removed `InvalidFeeForExactOut` error since it will just revert in SwapMath when trying to divide by 0


hensha256:
Please can you merge main so we can get updated gas snapshots? currently gas looks like its going _up_ by removing the check. So would love to check if thats true because we remove it

dianakocsis:
seems like some of them go down, but yeah for the most part they go up...weird

hensha256:
yah then i guess we should keep it

snreynolds (commented):
lol so weird, I think we should add it back explicitly, but in this PR can you please document WHY we don't allow it to happen on exactInput

src/libraries/Pool.sol
```diff
@@ -307,7 +310,16 @@ library Pool {
...
-        if (params.amountSpecified == 0) return (BalanceDeltaLibrary.ZERO_DELTA, 0, swapFee, state);
+        // a swap fee of MAX_FEE_PIPS (100%) makes exact output swaps impossible since the input is entirely consumed by the fee
+        if (swapFee == LPFeeLibrary.MAX_LP_FEE) {
+            if (!exactInput) {
+                InvalidFeeForExactOut.selector.revertWith();
+            }
+        }
+
+        // 0 is the fee amount that should be paid to the protocol
+        // swapFee is the pool's fee in pips (LP fee + protocol fee)
```
snreynolds: I think the first line could be something more around motivation for no protocol fee in this specific case like:

// when the amount swapped is 0, there is no protocolFee applied and the fee amount paid to protocol is set to 0

snreynolds (commented):
1 nit

hensha256:
imo this comment should be repeated as a `@dev` comment on
`IPoolManager.initialize`
`IPoolManager.updateDynamicLPFee`

Thats where integrators will be calling to do these things

hensha256 (approved):

PR #844, opened 2024-08-23
Spearbit Gas-08

dianakocsis:
## Related Issue
Spearbit Gas-08
https://github.com/spearbit-audits/review-uniswap-v4/issues/15

## Description of changes
using `FullMath.mulDiv()` in `donate` and `swap` in Pool.sol is unnecessary since the amount will never overflow


src/libraries/UnsafeMath.sol
```diff
@@ -16,4 +16,10 @@ library UnsafeMath {
             }
         }
     }
+
+    function simpleMulDiv(uint256 a, uint256 b, uint256 denominator) internal pure returns (uint256 result) {
```
hensha256, line 20: simple natspec `@notice` just to say what it does

src/libraries/Pool.sol
```diff
@@ -455,10 +456,10 @@ library Pool {
             // negation safe as amount0 and amount1 are always positive
             delta = toBalanceDelta(-(amount0.toInt128()), -(amount1.toInt128()));
             if (amount0 > 0) {
```
hensha256, line 458: would love a quick comment on each use saying why its safe to do it like this. Just 1 quick sentence nothing complicated :)

hensha256: e.g.
```
// FullMath.mulDiv not needed because numerator cannot exceed `type(int128).max) * Q128`
```
or something like that?
and then something similar for swap function too

hensha256 (dismissed):

hensha256 (approved):

PR #846, opened 2024-08-26
SB 12: Fix Comments

snreynolds:
Spearbit Issue#12

Note does not fix the _fetchProtocolFee comments as I think there is more logic to be changed. Will do that in a separate PR


src/interfaces/IPoolManager.sol
```diff
@@ -136,15 +136,18 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload, IExttload {
...
-    /// @return feeDelta The balance delta of the fees generated in the liquidity range. Returned for informational purposes.
+    /// @return callerDelta The balance delta of the caller of modifyLiquidity. This is the total of both principal and fee deltas
+    /// @return feesAccrued The balance delta of the fees generated in the liquidity range. Returned for informational purposes
     function modifyLiquidity(PoolKey memory key, ModifyLiquidityParams memory params, bytes calldata hookData)
         external
-        returns (BalanceDelta callerDelta, BalanceDelta feeDelta);
+        returns (BalanceDelta callerDelta, BalanceDelta feesAccrued);
 
     struct SwapParams {
+        /// Whether to swap token zero for token one or vice versa
```
hensha256, line 146:
```suggestion
        /// Whether to swap token0 for token1 or vice versa
```

src/interfaces/IPoolManager.sol
```diff
@@ -136,15 +136,18 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload, IExttload {
...
         bool zeroForOne;
+        /// The desired input amount if negative (exactIn), or the desired output amount if positive (exactOut)
         int256 amountSpecified;
+        /// The most extreme square root price the pool may reach by the end of the swap
```
hensha256, line 150:
```suggestion
        /// The sqrt price at which, if reached, the swap will stop executing early
```

Or something like that? Open to suggestions

src/interfaces/IPoolManager.sol
```diff
@@ -136,15 +136,18 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload, IExttload {
     /// @param key The pool to modify liquidity in
     /// @param params The parameters for modifying the liquidity
     /// @param hookData The data to pass through to the add/removeLiquidity hooks
-    /// @return callerDelta The balance delta of the caller of modifyLiquidity. This is the total of both principal and fee deltas.
-    /// @return feeDelta The balance delta of the fees generated in the liquidity range. Returned for informational purposes.
+    /// @return callerDelta The balance delta of the caller of modifyLiquidity. This is the total of both principal and fee deltas
```
hensha256, line 139: I know you werent changing this comment - but i wonder if it should mention the hook delta?
Really its the total of principal, fee, and hook deltas.

snreynolds: yeah can do!

hensha256 (approved):

PR #848, opened 2024-08-26
fix: Don't overload `state` naming and only return necessary vars in swap

snreynolds:
## Related Issue


.github/workflows/coverage.yml
saucepoint, line 1: looks like we need `--via-ir` on `forge coverage` as well. lets just fix it here too

saucepoint (approved):
looks good

.github/workflows/coverage.yml
snreynolds: yep saw that

saucepoint (approved):

src/libraries/Pool.sol
```diff
@@ -237,18 +237,12 @@ library Pool {
...
-        // the amount remaining to be swapped in/out of the input/output asset
-        int256 amountSpecifiedRemaining;
-        // the amount already swapped out/in of the output/input asset
-        int256 amountCalculated;
-        // current sqrt(price)
+    // the results after a swap, written to storage and returned
+    struct SwapResult {
+        // the final current sqrt(price)
         uint160 sqrtPriceX96;
         // the tick associated with the current price
```
hensha256, line 244: `current` -> `final`?

src/libraries/Pool.sol
```diff
@@ -237,18 +237,12 @@ library Pool {
...
     }
 
-    // the top level state of the swap, the results of which are recorded in storage at the end
-    struct SwapState {
-        // the amount remaining to be swapped in/out of the input/output asset
-        int256 amountSpecifiedRemaining;
-        // the amount already swapped out/in of the output/input asset
-        int256 amountCalculated;
-        // current sqrt(price)
+    // the results after a swap, written to storage and returned
```
hensha256, line 240: i found this misleading. It doesnt just hold the results after a swap, but during the entire swap too.
I'd rather it just said like
```
Tracks the state of a pool throughout a swap, and returns the resultant values at the end of the swap
```
or something?
I do think `SwapState` is much more descriptive of what this is. `SwapResult` implies it holds the result of a swap... which would be the unspecified token and final price i guess? Really its tracking the pool's state throughout the whole swap

src/libraries/Pool.sol
```diff
@@ -237,18 +237,12 @@ library Pool {
...
+    struct SwapResult {
+        // the final current sqrt(price)
         uint160 sqrtPriceX96;
         // the tick associated with the current price
         int24 tick;
-        // the global fee growth of the input token
-        uint256 feeGrowthGlobalX128;
```
hensha256, line 251: love this simplification of only returning the values needed in the pool manager

src/libraries/Pool.sol
snreynolds: Yeah I just can't think of something better & that doesn't overload state. can update comment

PR #850, opened 2024-08-27
SafeCast int128 to uint128

saucepoint:
## Related Issue

v4-periphery has a need to convert int128 to uint128, and has a temporary library:

https://github.com/Uniswap/v4-periphery/blob/main/src/libraries/SafeCast.sol

## Description of changes

Added int128 -> uint128 safecasting

### DIFF
--- a/test/libraries/SafeCast.t.sol
+++ b/test/libraries/SafeCast.t.sol
@@ -22,7 +22,7 @@ contract SafeCastTest is Test {
         SafeCast.toUint160(type(uint160).max + uint256(1));
     }
 
-    function test_fuzz_toUint128(uint256 x) public {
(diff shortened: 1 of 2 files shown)


test/libraries/SafeCast.t.sol
```diff
@@ -31,8 +31,17 @@ contract SafeCastTest is Test {
         }
     }
 
+    function test_fuzz_toUint128(int128 x) public {
```
hensha256, line 34: can we rename these tests more explicity? now that we have 2 functions both called `toUint128`
`test_fuzz_toUint128_fromInt128`
and below
`test_toUint128_fromUint256`

and any other tests we have for those 2 functions

hensha256 (approved):

PR #852, opened 2024-08-27
Revert certora L-01

dianakocsis:
## Related Issue
revert certora L-01
https://github.com/Uniswap/v4-core/pull/795/commits/0ebb5fb8e691f002b6264fb4fad3d727a9c05642#diff-be2175206ece3af563bbb97b8b473a3f14052c7fa49d7afaa8ced1403f20b4d5

## Description of changes
Certora M02 sometimes causes the pool to be at tick n-1 when it should be at tick n. 
Certora L01, a bug we initially fixed, prevented the pool from reaching MIN_TICK except at initialization, allowing it to only reach MIN_TICK+1. 
We originally fixed L01 so that the pool could reach MIN_TICK. However, when combined with M02, it sometimes incorrectly reported being at MIN_TICK-1. 
Since we are not addressing M02, it is better to revert the fix for L01.


snreynolds (commented):
left some thoughts in slack but I think we need to add natspec (related to my 2 questions in slack) and add context also in the PR description so we can resend it to Certora and update the audit report

dianakocsis:
Added the comments plus a more detailed PR description - let me know what u think

src/libraries/Pool.sol
```diff
@@ -325,7 +325,9 @@ library Pool {
             if (params.sqrtPriceLimitX96 >= slot0Start.sqrtPriceX96()) {
                 PriceLimitAlreadyExceeded.selector.revertWith(slot0Start.sqrtPriceX96(), params.sqrtPriceLimitX96);
             }
-            if (params.sqrtPriceLimitX96 < TickMath.MIN_SQRT_PRICE) {
+            // The pool can never read MIN_TICK except at initialization; after that, it only goes as low as MIN_TICK+1
+            // Allowing MIN_TICK could potentially lead to the pool incorrectly saying it's at MIN_TICK-1, so we prevent it from reaching MIN_TICK again
```
hensha256, line 329:
```suggestion
            // Swaps can never occur at MIN_TICK, only at MIN_TICK + 1, except at initialization of a pool
            // Under certain circumstances outlined below, the tick will preemptively reach MIN_TICK without swapping there
```

snreynolds (approved):

hensha256 (approved):

hensha256 (approved):

PR #855, opened 2024-08-28
fuzz tick math property

hensha256:


.forge-snapshots/TickMathGetTickAtSqrtPrice.snap
```diff
@@ -1 +1 @@
-196153
\ No newline at end of file
+195023
```
hensha256, line 1: this is a different test it doesn tmake sense lol

test/libraries/TickMath.t.sol
```diff
@@ -205,6 +205,26 @@ contract TickMathTestTest is Test, JavascriptFfi, GasSnapshot {
         }
     }
 
+    function test_fuzz_getTickAtSqrtPrice_getSqrtPriceAtTick_relation(int24 tick) public pure {
```
hensha256, line 208: this passes with 1 million runs locally

snreynolds (commented):
nice i like it!

.forge-snapshots/TickMathGetTickAtSqrtPrice.snap
snreynolds: super odd.. I guess let's see if this becomes flaky? Not sure why this test changed this val

snreynolds (approved):

PR #856, opened 2024-08-29
ToB L01

hensha256:


test/Sync.t.sol
```diff
@@ -119,58 +157,63 @@ contract SyncTest is Test, Deployers, GasSnapshot {
...
+        actions[3] = Actions.ASSERT_DELTA_EQUALS;
+        params[3] = abi.encode(currency2, address(router), 0);
 
         router.executeActions(actions, params);
     }
 
     /// @notice When there is a balance, no delta should be applied.
-    function test_settle_balanceInPool_shouldNotApplyDelta() public noIsolate {
+    function test_settle_balanceInPool_shouldNotApplyDelta() public {
         uint256 currency0Balance = currency0.balanceOf(address(manager));
```
saucepoint, line 183: minor nit for consistency with other tests, but we should assertGt 0 right

saucepoint (commented):
looks good, also should we add `/// @dev PoolManager is not unlocked` to `IProtocolFees.collectProtocolFees`?

saucepoint (dismissed):
will defer to @snreynolds for one last look

src/PoolManager.sol
```diff
@@ -383,4 +383,9 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     function _getPool(PoolId id) internal view override returns (Pool.State storage) {
         return _pools[id];
     }
+
+    /// @notice Implementation of the _isUnlocked function defined in ProtocolFees
+    function _isUnlocked() internal view override returns (bool) {
```
snreynolds, line 388: also use this function in the modifier?

src/test/PoolSettleTest.sol
```diff
@@ -3,18 +3,22 @@ pragma solidity ^0.8.24;
...
+    function settle(Currency syncCurrency) external payable {
+        manager.unlock(abi.encode(syncCurrency));
     }
 
-    function unlockCallback(bytes calldata) external returns (bytes memory) {
+    function unlockCallback(bytes calldata data) external returns (bytes memory) {
         require(msg.sender == address(manager));
 
+        Currency syncCurrency = abi.decode(data, (Currency));
+        manager.sync(syncCurrency);
```
snreynolds, line 20: you shouldnt need to call sync here if the currency we are settling is native?

test/utils/Deployers.sol
```diff
@@ -87,15 +87,6 @@ contract Deployers {
     uint160 hookPermissionCount = 14;
     uint160 clearAllHookPermissionsMask = ~uint160(0) << (hookPermissionCount);
 
-    modifier noIsolate() {
```
snreynolds, line 90: What was this from?

hensha256: it was so the tests could call `sync` in a separate call than the actions router stuff. But now sync has to be in there to be in the lock so we dont need it

src/test/PoolSettleTest.sol
hensha256: Its only used for 1 test, and its a test where the synced currency is _different_ than whats being settled. So just had to move that inside this contract. But arguably given this contract is only used for 1 test we should just use actions router

hensha256: ok i deleted PoolSettleTest and rewrote it with actions router

snreynolds (approved):

src/PoolManager.sol
```diff
@@ -277,7 +277,7 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     }
 
     /// @inheritdoc IPoolManager
-    function sync(Currency currency) external {
+    function sync(Currency currency) external onlyWhenUnlocked {
```
wjmelements, line 280: what is the reason for this change?

wjmelements: It's unnecessary for the manager to be unlocked to `sync` before and after this change.

PR #857, opened 2024-08-29
Spearbit core issues

hensha256:
Small fixes for:
https://github.com/spearbit-audits/review-uniswap-v4/issues/58
https://github.com/spearbit-audits/review-uniswap-v4/issues/102
https://github.com/spearbit-audits/review-uniswap-v4/issues/67
https://github.com/spearbit-audits/review-uniswap-v4/issues/95
https://github.com/spearbit-audits/review-uniswap-v4/issues/57


src/libraries/SwapMath.sol
```diff
@@ -70,7 +70,7 @@ library SwapMath {
                     // `amountIn` is capped by the target price
                     sqrtPriceNextX96 = sqrtPriceTargetX96;
                     feeAmount = _feePips == MAX_SWAP_FEE
-                        ? amountIn
+                        ? 0 // 0 as amountRemainingLessFee == 0
```
dianakocsis, line 73: I think in the issue they said to keep it as amountIn and just comment that it will always be 0. But I see that it doesn't cost extra gas for us anyways
But also maybe make the comment more clear like 
// Always 0 because if amountRemainingLessFee >= amountIn with the max fee, amountIn must be 0 (making both amountRemainingLessFee and feeAmount 0).

dianakocsis (changes requested):

src/libraries/SqrtPriceMath.sol
```diff
@@ -238,11 +238,8 @@ library SqrtPriceMath {
     {
         uint256 numerator = absDiff(sqrtPriceAX96, sqrtPriceBX96);
         uint256 denominator = FixedPoint96.Q96;
-        uint256 _liquidity;
-        assembly ("memory-safe") {
-            // avoid implicit upcasting
-            _liquidity := and(liquidity, 0xffffffffffffffffffffffffffffffff)
-        }
+        uint256 _liquidity = uint256(liquidity);
```
snreynolds, line 241: hm but why do we have a comment that says "avoid implicit upcasting".. feels like this is what we/someone was trying to avoid?

hensha256: it was just done in PR #676 to save gas. But now that we've added masking and optimiser runs are higher, it doesnt save gas anymore.

dianakocsis (approved):

PR #859, opened 2024-08-30
SB-i96:move to constants file, add natspec

snreynolds:
https://github.com/spearbit-audits/review-uniswap-v4/issues/96


test/libraries/TickMath.t.sol
```diff
@@ -15,8 +15,6 @@ contract TickMathTestTest is Test, JavascriptFfi, GasSnapshot {
     uint160 constant MIN_SQRT_PRICE = 4295128739;
     uint160 constant MAX_SQRT_PRICE = 1461446703485210103287273052203988822378723970342;
```
hensha256, line 16: i feel like these constants should also just be used directly from TickMath.sol?
`MIN_SQRT_PRICE = TickMath.MIN_SQRT_PRICE`
etc

hensha256 (dismissed):
lgtm!

saucepoint (approved):
looks good, but agree with alice that the MIN/MAX SQRT PRICE should use tickmath's constants

saucepoint (approved):

PR #860, opened 2024-08-30
fix: use State for consistency, add natspec on SwapResults

snreynolds:
Some comments from #848 and I want to move the Positions struct to be consistent with Pool.State


src/libraries/Pool.sol
```diff
@@ -237,9 +237,9 @@ library Pool {
         }
     }
 
-    // the results after a swap, written to storage and returned
+    // Tracks the state of a pool throughout a swap, and bubbles these values up at the end of the swap
```
hensha256, line 240:
```suggestion
    // Tracks the state of a pool throughout a swap, and returns these values at the end of the swap
```

src/libraries/Pool.sol
```diff
@@ -237,9 +237,9 @@ library Pool {
...
     struct SwapResult {
-        // the final current sqrt(price)
+        // the final sqrt(price)
```
hensha256, line 242: i'd rather it was just `current`? Its not just final, its current throughout the swap

hensha256 (approved):

PR #861, opened 2024-08-30
gas limit as percentage

hensha256:


src/libraries/BipsLibrary.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.0;
+
+/// @title For calculating a percentage of an amount, using bips
```
hensha256, line 4: This is copied across from periphery now that its been audited.

src/ProtocolFees.sol
```diff
@@ -17,18 +18,19 @@ abstract contract ProtocolFees is IProtocolFees, Owned {
...
 
-    uint256 private immutable controllerGasLimit;
+    // a percentage of the block.gaslimit denoted in basis points, used as the gas limit for fee controller calls
+    // 100 bps is 1%, at 30M gas, the limit is 300K
+    uint256 private constant BLOCK_LIMIT_BPS = 100;
 
-    constructor(uint256 _controllerGasLimit) Owned(msg.sender) {
-        controllerGasLimit = _controllerGasLimit;
-    }
+    constructor() Owned(msg.sender) {}
```
saucepoint, line 33: \*happy noises\*

src/ProtocolFees.sol
```diff
@@ -17,18 +18,19 @@ abstract contract ProtocolFees is IProtocolFees, Owned {
...
 
     /// @inheritdoc IProtocolFees
     mapping(Currency currency => uint256 amount) public protocolFeesAccrued;
 
     /// @inheritdoc IProtocolFees
     IProtocolFeeController public protocolFeeController;
 
-    uint256 private immutable controllerGasLimit;
+    // a percentage of the block.gaslimit denoted in basis points, used as the gas limit for fee controller calls
+    // 100 bps is 1%, at 30M gas, the limit is 300K
```
snreynolds, line 30: Are we fine with this number? 

This is what we use for subscribers on periphery?

hensha256: yeah it seemed reasonable here too. 300k would be a lotttttttt. arguably id be happy to go with smaller lol

saucepoint (commented):
looks good to me; 300K also probably chill (running it by erin too)

is there a test that confirms the 1% is the indeed the gas limit. just to validate that the 1% math is working, and its actually being picked up correctly by ProtocolFees.sol

snreynolds:
> looks good to me; 300K also probably chill (running it by erin too)
> 
> is there a test that confirms the 1% is the indeed the gas limit. just to validate that the 1% math is working, and its actually being picked up correctly by ProtocolFees.sol

I wrote tests for this in periphery - maybe let's transfer them over @hensha256 ?

test/libraries/BipsLibrary.t.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import "forge-std/Test.sol";
+import "forge-std/StdError.sol";
```
saucepoint, line 5: unused import

saucepoint: but otherwise looks good

saucepoint (approved):

PR #866, opened 2024-09-03
SB-19: Remove revert sync

snreynolds:
minimize DoS vectors by allowing the sync currency to be overridden
https://github.com/spearbit-audits/review-uniswap-v4/issues/19


src/PoolManager.sol
```diff
@@ -278,11 +278,14 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
 
     /// @inheritdoc IPoolManager
     function sync(Currency currency) external onlyWhenUnlocked {
-        CurrencyReserves.requireNotSynced();
```
hensha256, line 281: need to delete this function from the CurrencyReserves library

src/interfaces/IPoolManager.sol
```diff
@@ -184,6 +184,8 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload, IExttload {
     /// This is used to checkpoint balances for the manager and derive deltas for the caller.
     /// @dev This MUST be called before any ERC20 tokens are sent into the contract, but can be skipped
     /// for native tokens because the amount to settle is determined by the sent value.
+    /// However, if there is an ERC20 token address set in the syncedCurrency slot and the caller instead wants to settle
+    /// native funds, this function can be called with the native currency as an input to clear the slot.
```
hensha256, line 188:
```suggestion
    /// However, if an ERC20 token has been synced and not settled, and the caller instead wants to settle
    /// native funds, this function can be called with the native currency to then be able to settle the native currency
```

hensha256: just a suggestion but i think there should be less reference to "the slot" and more of an integration explanation

test/Sync.t.sol
```diff
@@ -289,4 +294,77 @@ contract SyncTest is Test, Deployers, GasSnapshot {
...
+        actions[0] = Actions.SYNC;
+        params[0] = abi.encode(currency2);
+
+        actions[1] = Actions.ASSERT_RESERVES_EQUALS;
+        params[1] = abi.encode(value);
+
+        actions[2] = Actions.TRANSFER_FROM;
+        params[2] = abi.encode(currency2, address(this), address(manager), value);
+
+        actions[3] = Actions.SETTLE;
```
hensha256, line 354: i think not explicitly assigning the params makes it hard to read - would rather it actually assigned the native currency explicitly

hensha256: i guess thats what its doing?

snreynolds: hm no, this is cause it just uses the syncedCurrency slot

so this line is settling currency2 cause we just called sync with cyrrency2

hensha256 (approved):

src/interfaces/IPoolManager.sol
```diff
@@ -184,7 +184,8 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload, IExttload {
     /// This is used to checkpoint balances for the manager and derive deltas for the caller.
     /// @dev This MUST be called before any ERC20 tokens are sent into the contract, but can be skipped
     /// for native tokens because the amount to settle is determined by the sent value.
-    /// @param currency The currency whose balance to sync
```
HickupHH3, line 187: the natspec got accidentally removed 😅

PR #867, opened 2024-09-03
Spearbit 43 and 44

hensha256:
Adding better comments around certain hardcoded values, and how they have been approximated


src/libraries/TickMath.sol
```diff
@@ -69,6 +69,9 @@ library TickMath {
 
             if (absTick > uint256(int256(MAX_TICK))) InvalidTick.selector.revertWith(tick);
 
+            // The tick is decomposed into bits, and for each bit with index i that is set, the product of 1/sqrt(1.0001^(2^i))
```
snreynolds, line 72: either 2**128 or add comment that constants are in this notation

snreynolds (approved):
lovely!!!!! tick maths <3

PR #868, opened 2024-09-04
NoDelegateCall.sol MIT License

Alexangelj:
## Description of changes

Changes NoDelegateCall.sol from BUSL-1.1 to MIT license.


saucepoint (approved):
hope this doesnt get me fired, lets go MIT gang

PR #870, opened 2024-09-04
Spearbit 105

hensha256:
## Related Issue


test/Tick.t.sol
```diff
@@ -116,21 +116,21 @@ contract TickTest is Test, GasSnapshot {
     function testTick_tickSpacingToMaxLiquidityPerTick_returnsTheCorrectValueForLowFee() public pure {
```
snreynolds, line 116: i think we should add a test for when the min tick % tick spacing is 0 cause the number then shouldn't change from impl to impl

hensha256: there are 2 such tests already 
testTick_tickSpacingToMaxLiquidityPerTick_returnsTheCorrectValueForMinTickSpacing
testTick_tickSpacingToMaxLiquidityPerTick_returnsTheCorrectValueForMaxTickSpacing

snreynolds (approved):
🥂

PR #872, opened 2024-09-05
Remove mandatory viaIR

saucepoint:
## Related Issue

Mandatory viaIR sucks 💯

## Description of changes

- [x] Killed myself in Pool.sol


src/libraries/Pool.sol
```diff
@@ -265,13 +265,20 @@ library Pool {
     }
 
     struct SwapParams {
+        int256 amountSpecified;
```
snreynolds, line 268: does this ordering change matter?

snreynolds (approved):
i also do not like via-ir

src/libraries/Pool.sol
saucepoint: we removed the cached `exactInput = params.amountSpecified < 0`, and are now using `params.amountSpecified < 0` in place of the cached variable

making it the first element in the struct avoids the need of offset calculations

src/libraries/Pool.sol
```diff
@@ -281,22 +283,19 @@ library Pool {
         Slot0 slot0Start = self.slot0;
         bool zeroForOne = params.zeroForOne;
 
-        uint128 liquidityStart = self.liquidity;
         uint256 protocolFee =
             zeroForOne ? slot0Start.protocolFee().getZeroForOneFee() : slot0Start.protocolFee().getOneForZeroFee();
```
hensha256, line 288: `start.liquidityStart` -> `start.liquidity` ?

src/libraries/Pool.sol
```diff
@@ -308,11 +307,10 @@ library Pool {
             swapFee = protocolFee == 0 ? lpFee : uint16(protocolFee).calculateSwapFee(lpFee);
         }
 
-        bool exactInput = params.amountSpecified < 0;
-
         // a swap fee totaling MAX_SWAP_FEE (100%) makes exact output swaps impossible since the input is entirely consumed by the fee
         if (swapFee >= SwapMath.MAX_SWAP_FEE) {
-            if (!exactInput) {
+            // if not exactInput
+            if (!(params.amountSpecified < 0)) {
```
hensha256, line 313:
```suggestion
            if (params.amountSpecified >= 0) {
```
might be better?

hensha256 (approved):

PR #874, opened 2024-09-09
Flipped tick check optimization

Alexhandru:
## Related Issue

There is no issue but an optimization when checking if tick flipped.

## Description of changes

The present code for checking if tick is flipped "flipped = (liquidityGrossAfter == 0) != (liquidityGrossBefore == 0);" is basically an xor that doesn't work for liquidityGrossAfter/Before > 1, so that's why we reduce any non-zero value to 1.

I calculated 270 gas savings across all tests, though some instances of liquidity removal show slightly higher gas costs.
However, swaps seem to present slight gas savings that might be preferred.


hensha256:

PR #878, opened 2024-09-13
Change BalanceDelta and BeforeSwapDelta from int256 to bytes32

jeffishjeff:
## Description of changes
As these delta types are packed bytes of multiple variables, bytes32 would be a more "neutral" underlying type than int256. Arguably make code easier to understand and discourages direct integer manipulation.


hensha256:

PR #879, opened 2024-09-13
Simplify modifyLiquidity

jeffishjeff:
## Description of changes
- It's not necessary to initialize feeGrowthOutside (and by extension feeGrowthInside) for newly initialized ticks, as feeGrowthInside is always taken in relativity when used for calculating feeOwned
-  If that's removed, we can move fee accrue logic to the front in modifyLiquidity(), then combine previously fragmented tick related logic (i.e. clear tick) all inside updateTick()
- Net result is a bit of gas saving and making the code easier to read


hensha256:

PR #887, opened 2024-10-02
Remove pull of protocol fee

hensha256:


wjmelements (approved):

.forge-snapshots/initialize.snap
```diff
@@ -1 +1 @@
-60013
\ No newline at end of file
+52013
```
snreynolds, line 1: some nice savings

snreynolds (commented):
I think we can remove `BipsLibrary` and `BipsLibrary.t.sol` and put them in periphery again. Otherwise looks good!

kmbarry1:
You should be able to remove this event as well:
https://github.com/Uniswap/v4-core/blob/1141642f8ba4665a50660886a8a8401526677045/src/interfaces/IProtocolFees.sol#L11-L12

src/libraries/Pool.sol
```diff
@@ -94,16 +94,13 @@ library Pool {
...
-        returns (int24 tick)
-    {
+    function initialize(State storage self, uint160 sqrtPriceX96, uint24 lpFee) internal returns (int24 tick) {
         if (self.slot0.sqrtPriceX96() != 0) PoolAlreadyInitialized.selector.revertWith();
 
         tick = TickMath.getTickAtSqrtPrice(sqrtPriceX96);
 
-        self.slot0 = Slot0.wrap(bytes32(0)).setSqrtPriceX96(sqrtPriceX96).setTick(tick).setProtocolFee(protocolFee)
-            .setLpFee(lpFee);
+        // the initial protocolFee is 0 so doesnt need to be set
```
kmbarry1, line 102: typo: `doesnt` --> `doesn't`

snreynolds (approved):

PR #889, opened 2024-10-04
Update ERC-7751 revert bubbling

gretzke:


marktoda (approved):
looks good to me

src/libraries/CustomRevert.sol
```diff
@@ -75,23 +78,33 @@ library CustomRevert {
         }
     }
 
-    /// @notice bubble up the revert message returned by a call and revert with the selector provided
-    /// @dev this function should only be used with custom errors of the type `CustomError(address target, bytes revertReason)`
-    function bubbleUpAndRevertWith(bytes4 selector, address addr) internal pure {
+    /// @notice bubble up the revert message returned by a call and revert with a wrapped ERC-7751 error
+    function bubbleUpAndRevertWith(address target, bytes4 functionSelector, bytes4 additionalContext) internal pure {
```
marktoda, line 82: consider renaming these to be a bit more clear, maybe:
address revertingContract
address revertingFunctionSelector
bytes4 additionalContext

marktoda: also -- why is `additionalContext` only bytes4 here but bytes in the custom error? seems context could be a custom error w/ params so should we support that here too?

marktoda: I guess keeping it bytes4 makes the encoding more simple if we specifically dont plan on having params here

gretzke: any additional context we had in the reverts was the target address (hook, erc20 token, recipient) based on the action, so we only needed 4bytes for the selector

marktoda (approved):

marktoda (approved):

src/types/Currency.sol
```diff
@@ -30,13 +30,11 @@ function greaterThanOrEqualTo(Currency currency, Currency other) pure returns (b
 library CurrencyLibrary {
     using CustomRevert for bytes4;
```
hensha256, line 31: can this line go now?

src/libraries/CustomRevert.sol
```diff
@@ -75,23 +78,43 @@ library CustomRevert {
...
+                and(revertingFunctionSelector, 0xffffffff00000000000000000000000000000000000000000000000000000000)
+            )
+            // offset revert reason
+            mstore(add(fmp, 0x44), 0x80)
+            // offset additional context
+            mstore(add(fmp, 0x64), add(0xa0, encodedDataSize))
+            // size revert reason
+            mstore(add(fmp, 0x84), size)
+            // revert reason
+            returndatacopy(add(fmp, 0xa4), 0, size)
```
0age, line 109: It's actually cheaper to use returndatasize in place rather than assigning it to a stack variable!

0age (dismissed):

0age (approved):

hensha256 (approved):

PR #894, opened 2024-10-07
Remove initialize hookdata

saucepoint:
## Related Issue

Cantina 153, 444 - initialize hookData can be frontran with unintended data

## Description of changes

Remove `bytes memory hookData` from beforeInitialize/afterInitialize


.forge-snapshots/addLiquidity with empty hook.snap
```diff
@@ -1 +1 @@
-274206
\ No newline at end of file
+274240
```
hensha256, line 1: 😠

hensha256 (approved):

PR #902, opened 2024-10-11
Pin forge version

mr-uniswap:
Pins the version for forge

## Related Issue


.github/workflows/tests-merge.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly
+          version: 0.2.0
```
marktoda, line 23: I think this will be a long long time ago. Generally for foundry pin to a hash from here https://github.com/foundry-rs/foundry/releases

PR #903, opened 2024-10-15
chore: remove unnecessary line

jrcarlos2000:
## Related Issue

remove unnecessary line, pass tests, no functions from `Position` library are used.

### DIFF
--- a/src/PoolManager.sol
+++ b/src/PoolManager.sol
@@ -80,7 +80,6 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     using SafeCast for *;
     using Pool for *;
     using Hooks for IHooks;
-    using Position for mapping(bytes32 => Position.State);
     using CurrencyDelta for Currency;
     using LPFeeLibrary for uint24;
     using CurrencyReserves for Currency;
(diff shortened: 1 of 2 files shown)


src/PoolManager.sol
```diff
@@ -80,7 +80,6 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
     using SafeCast for *;
     using Pool for *;
     using Hooks for IHooks;
-    using Position for mapping(bytes32 => Position.State);
```
hensha256, line 83: i think youre right. Can you also remove it in the ProxyPoolManager.sol

0xquantum3labs: done ✅

jrcarlos2000: done ✅

jrcarlos2000: @hensha256

hensha256:
Sorry for the delay here! We've now added a snapshot to check bytecode hash in #924, so if CI passes here we know that you dont alter the bytecode and we can merge :)

jrcarlos2000:
thanks @hensha256

jrcarlos2000:
@hensha256 should be good now :)

PR #905, opened 2024-10-16
round swap fee up and check if protocol fee = swap fee

dianakocsis:
## Related Issue


src/libraries/Pool.sol
```diff
@@ -381,10 +381,17 @@ library Pool {
             // if the protocol fee is on, calculate how much is owed, decrement feeAmount, and increment protocolFee
             if (protocolFee > 0) {
                 unchecked {
-                    // step.amountIn does not include the swap fee, as it's already been taken from it,
-                    // so add it back to get the total amountIn and use that to calculate the amount of fees owed to the protocol
-                    // this line cannot overflow due to limits on the size of protocolFee and params.amountSpecified
-                    uint256 delta = (step.amountIn + step.feeAmount) * protocolFee / ProtocolFeeLibrary.PIPS_DENOMINATOR;
+                    uint256 delta;
+                    if (swapFee != protocolFee) {
```
hensha256, line 385: is this check equivalent to `if (lpFee != 0)` ?
I would find that more intuitive to read

src/libraries/Pool.sol
```diff
@@ -381,10 +381,17 @@ library Pool {
...
+                        // step.amountIn does not include the swap fee, as it's already been taken from it,
+                        // so add it back to get the total amountIn and use that to calculate the amount of fees owed to the protocol
+                        // this line cannot overflow due to limits on the size of protocolFee and params.amountSpecified
+                        // this rounds down to favor LPs over the protocol
+                        delta = (step.amountIn + step.feeAmount) * protocolFee / ProtocolFeeLibrary.PIPS_DENOMINATOR;
+                    } else {
+                        // if the protocol fee is the same as the swap fee, the entire fee is owed to the protocol
+                        delta = step.feeAmount;
```
hensha256, line 393: also think i would rather this if statement written as
```
uint256 delta = (lpFee == 0)
   ? step.feeAmount
   : (step.amountIn + step.feeAmount) * protocolFee / ProtocolFeeLibrary.PIPS_DENOMINATOR
```

or lpFee != 0 and the other way around, whichever is cheaper probably

src/libraries/Pool.sol
dianakocsis: we cant access lpFee because its in a scoped block :(

src/libraries/Pool.sol
dianakocsis: oh yeah nice

dianakocsis: doesnt make a difference on gas either way

src/libraries/Pool.sol
```diff
@@ -383,8 +383,11 @@ library Pool {
                 unchecked {
                     // step.amountIn does not include the swap fee, as it's already been taken from it,
                     // so add it back to get the total amountIn and use that to calculate the amount of fees owed to the protocol
-                    // this line cannot overflow due to limits on the size of protocolFee and params.amountSpecified
-                    uint256 delta = (step.amountIn + step.feeAmount) * protocolFee / ProtocolFeeLibrary.PIPS_DENOMINATOR;
+                    // cannot overflow due to limits on the size of protocolFee and params.amountSpecified
+                    // this rounds down to favor LPs over the protocol
+                    uint256 delta = (swapFee == protocolFee)
+                        ? step.feeAmount // if there is no lp fee, the entire fee is owed to the protocol instead
```
hensha256, line 389:
```suggestion
                        ? step.feeAmount // lp fee is 0, so the entire fee is owed to the protocol instead
```

test/DynamicFees.t.sol
```diff
@@ -307,7 +301,22 @@ contract TestDynamicFees is Test, Deployers, GasSnapshot {
...
 
-        uint256 expectedProtocolFee = uint256(uint128(-delta.amount0())) * protocolFee0 / 1e6;
+        uint24 swapFee = uint16(protocolFee).calculateSwapFee(lpFee);
+
+        uint256 expectedProtocolFee;
+        int128 delta0 = -delta.amount0();
+
+        if (lpFee == 0) {
+            assertEq(protocolFee0, swapFee);
+            assembly {
```
hensha256, line 311: i think in the tests i'd feel happier doing this in solidity to check our assembly in the core code is correct/
something like

```
        expectedProtocolFee = uint256(uint128(delta0)) * protocolFee0 / 1e6;
        if (lpFee == 0) {
            assertEq(protocolFee0, swapFee);
            if (((delta0 * protocolFee0) % 1e6) != 0) expectedProtocolFee++
```

test/libraries/ProtocolFeeLibrary.t.sol
```diff
@@ -71,18 +71,27 @@ contract ProtocolFeeLibraryTest is Test, GasSnapshot {
...
+        uint256 expectedSwapFee;
+        // protocolFee + lpFee(1_000_000 - protocolFee) / 1_000_000 (rounded up)
+        assembly ("memory-safe") {
+            expectedSwapFee :=
+                add(
+                    protocolFee,
+                    add(
+                        div(mul(sub(1000000, protocolFee), lpFee), 1000000),
+                        gt(mod(mul(sub(1000000, protocolFee), lpFee), 1000000), 0)
+                    )
```
hensha256, line 90: same here i think our tests shouldnt use assembly

hensha256 (approved):

PR #910, opened 2024-10-22
Update Lock.sol in Solidity 0.8.28 for transient variable support

MarcusWentz:
## Related Issue

From Lock.sol:

```solidity
/// @notice This is a temporary library that allows us to use transient storage (tstore/tload)
/// TODO: This library can be deleted when we have the transient keyword support in solidity.
```

## Description of changes

Update Lock.sol in Solidity 0.8.28 for transient variable support

We can now write Solidity transient storage variables as shown in this example:
```solidity
// SPDX-License-Identifier: BUSL-1.1
pragma solidity 0.8.28;

contract Lock {
    bool internal transient IS_UNLOCKED_SLOT;

    function unlock() internal {
        IS_UNLOCKED_SLOT = true;
    }

    function lock() internal {
        IS_UNLOCKED_SLOT = false;
    }

    function isUnlocked() internal view returns (bool unlocked) {
        unlocked = IS_UNLOCKED_SLOT;
    }
}
```

Solidity 0.8.28 documentation:

https://docs.soliditylang.org/en/v0.8.28/

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -3,7 +3,7 @@ optimizer_runs = 44444444
 via_ir = true
 ffi = true
 fs_permissions = [{ access = "read-write", path = ".forge-snapshots/"}, { access = "read", path = "./out"}, {access = "read", path = "./test/bin"}]
-solc = "0.8.26"
+solc = "0.8.28"
 evm_version = "cancun"
 gas_limit = "300000000"
 
--- a/src/PoolManager.sol
+++ b/src/PoolManager.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: BUSL-1.1
-pragma solidity 0.8.26;
+pragma solidity 0.8.28;
 
 import {Hooks} from "./libraries/Hooks.sol";
 import {Pool} from "./libraries/Pool.sol";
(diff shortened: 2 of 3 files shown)


MarcusWentz:
Related to:

https://github.com/Uniswap/v4-core/pull/751

@hensha256 was v4-core updated fully to deprecate Lock.sol with transient storage variables in Solidity 0.8.28

https://github.com/Uniswap/v4-core/pull/751#issuecomment-2227329041

?

MarcusWentz:
I see PoolManager.sol is still importing Lock.sol:

https://github.com/Uniswap/v4-core/blob/main/src/PoolManager.sol#L21

wjmelements (changes requested):
You would need to update `TransientStateLibrary.sol` because the slot is changing.

Then run `forge snapshot --isolate` and commit the `.forge-snapshots` changes.

wjmelements:
Update the solidity version in `foundry.toml`.

MarcusWentz:
@wjmelements 

1. Update the solidity version in foundry.toml.

Added this now.

2. You would need to update TransientStateLibrary.sol because the slot is changing.

What would need to be changed in TransientStateLibrary.sol? 
Would it be possible to make Lock.sol a library again?
I rewrote Lock.sol as a contract instead of a library to avoid this state error:
```
TypeError: Library cannot have non-constant state variables
```

hensha256:
Hi!! We really appreciate this contribution - and we of course wish it had been added to solidity sooner - however this code has already been through 5 audits, and so we will not be making these changes now. We will launch continuing to use `tstore` directly in assembly.

PR #915, opened 2024-10-30
store PoolKey in Pool.State

srbrbslt:
Integrators need the PoolKey to interact with a pool. Currently, this is not stored on-chain. In order to obtain this data for a specific PoolId, one needs to either:

- Search for the Initialize event, probably over a very large range of blocks.
- Find a transaction with a swap/modifyLiquidity call, trace it, find the correct frame, and decode the calldata into PoolKey.

Both methods are quite complex. This PR proposes to store the PoolKey in Pool.State during initialization and make it easily accessible for integrators and users with a new getPoolKey function in StateLibrary.

### DIFF
--- a/snapshots/PoolManagerInitializeTest.json
+++ b/snapshots/PoolManagerInitializeTest.json
@@ -1,3 +1,3 @@
 {
-  "initialize": "51532"
+  "initialize": "98285"
 }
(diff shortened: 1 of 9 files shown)


hensha256:
Hi! V4 core is now frozen :) we discussed this at length but decided that doubling the cost of deployment of new pools meant that this wasn't worth it. There will a subgraph that has this information available!

Issue #918, opened 2024-11-12
Document the Finalized Initcode


wjmelements:
### Describe the suggested feature and problem it solves.

The initcode hash is published but the initcode is not. If the initcode matching the initcode hash were to be lost, it could cause issues with the final deployment next year.

### Describe the desired implementation.

Commit the finalized output files into version control. These include the initcode and the ABI.

### Describe alternatives.

Commit separate files with the final initcode and ABI.

### Additional context.

I am having trouble generating initcode that matches the published initcode hash.

I worry that the initcode hash in the [contest contract](https://etherscan.io/address/0x48e516b34a1274f49457b9c6182097796d0498cb#code) may be incorrect because I cannot reproduce it. I calculate `0x44d2e366954aa5a7f0605ffd66ca57a5a71987845939d8d8c79c716d8dfbb3de` instead. Here is how:

I checkout `main` and do `forge update` and `forge build`.

I do `cat out/PoolManager.t.sol/PoolManagerTest.json | jq -r .bytecode.object > v4.out`

In python3 I do:
```
from web3 import Web3
with open('v4.out', 'r') as f:
    v4 = f.read().rstrip()
print(Web3.keccak(hexstr=v4))
```
The result is `0x44d2e366954aa5a7f0605ffd66ca57a5a71987845939d8d8c79c716d8dfbb3de`.

gretzke:
constructor arguments are part of the initcode so they need to be taken into account. The hash that you found (`0x2aa0ab6866fc2e2b2a0a9128e66982a049eb04516c012690426ab110f92966f8`) is the initcode encoded with the zero address as a constructor argument. If you take the Governance timelock address and encode it with the bytecode you will find that the initcode hash matches the one from the competition

PR #921, opened 2024-11-22
Add compilation restrictions

gretzke:


wjmelements:
Does this help with #918?

gretzke:
no, this PR focuses on the reduction of compilation times

gretzke:
@reviewers please double check the bytecode!

gretzke:
If we merge this, we can't update v4-periphery to point to the new main branch because I'm pretty sure `vm.getCode` will fail because the compiler artifact is different in periphery.

test/utils/Deployers.sol
```diff
@@ -83,8 +82,13 @@ contract Deployers is Test {
         }
     }
 
-    function deployFreshManager() internal virtual {
-        manager = new PoolManager(address(this));
+    function deployFreshManager() internal virtual returns (IPoolManager manager_) {
+        bytes memory bytecode =
+            abi.encodePacked(vm.getCode("out/PoolManager.sol/PoolManager.default.json"), abi.encode(address(this)));
```
hensha256, line 87: why cant you do it as it was before?

hensha256:
> If we merge this, we can't update v4-periphery to point to the new main branch because I'm pretty sure vm.getCode will fail because the compiler artifact is different in periphery.

This seems like a problem to me?

test/utils/Deployers.sol
gretzke: I agree that it's somewhat of a problem. But we can probably also migrate periphery to use briefcase instead of v4 directly?
If we deploy the pool manager using the `new` keyword instead of assembly, then the tests run against a version of the pool that is compiled without IR, so the bytecode of the pool manager will be different.

snreynolds: Why does `new` not use ir @gretzke ? Didn't know that

gretzke: because of the compilation restriction we set up (`{ paths = "**/test/**", via_ir = false }`). It means that all test files are compiled without IR. If a contract is imported to be deployed using the `new` keyword, it means that this contract is also compiled without IR for the scope of this test and thus tests run against a version compiled without IR. Which is fine in itself but if we want to run tests against the bytecode that will be deployed later on chain we need to avoid importing the contract directly and instead load the bytecode from the file that was compiled with IR.

snreynolds:
Closing since we wouldn't be able to config v4-periphery to point to the new main branch of core. Blocked by https://github.com/ethereum/solidity/issues/15582

Issue #797, opened 2024-07-16
consistency in mask constants


snreynolds:
### Component

General design optimization (improving efficiency, cleanliness, or developer experience)

### Describe the suggested feature and problem it solves.

In many places we use uint160/uint24 masks. In slot0.sol we define them as constants, but everywhere else we just inline them. I think we should just make it consistent wherever we are masking

PR #923, opened 2024-12-05
Update Slot0.t.sol so that masks are inlined (Issue #797)

floor-licker:
In response to Issue #797, the bit masks are now inlined as opposed to being defined as constants.

Slot0.t.sol previously deviated from the standard of inline bit masks, by instead defining them as constants. I've updated the code so that they now conform to the standard that the other contracts adhere to, which is masks being defined inline.

Please feel free to respond to the pull request or reach out via my socials if there are any changes that need to be made.

### DIFF
--- a/src/interfaces/IPoolManager.sol
+++ b/src/interfaces/IPoolManager.sol
@@ -224,6 +224,13 @@ interface IPoolManager is IProtocolFees, IERC6909Claims, IExtsload, IExttload {
     /// If the upper 12 bytes are not 0, they will be 0-ed out
     function burn(address from, uint256 id, uint256 amount) external;
 
+    /// @notice Emitted when a dynamic LP fee is updated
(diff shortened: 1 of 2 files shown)


floor-licker:
Hey @vimageDE just wondering if you'd be able to take a look at this PR. Thanks!

floor-licker:
Hey @hensha256 just wondering if you'd be able to take a look at this PR. Thanks!

snreynolds:
hey! I think actually issue #797 was to make all places where we do any masks a constant so that they matched Slot0.sol file, and not actually worried about the test files.

snreynolds:
Since the code is frozen now, we won't be making that change and I'll be closing the issue #797

snreynolds:
Closing as 797 is no longer an issue

Issue #920, opened 2024-11-20
Event for dynamic fee update


kaber2:
### Component

General design optimization (improving efficiency, cleanliness, or developer experience)

### Describe the suggested feature and problem it solves.

While implementing support for UniswapV4 in my software, I noticed there is no event for dynamic fee updates. I realize that hooks can override fees anyways, which will already make fully supporting UniswapV4 rather complicated. It would be appreciated if at least dynamic fees can be handled through events.

### Describe the desired implementation.

Emit event in updateDynamicLPFee().

kaber2:
> [PROTO-686 Event for dynamic fee update](https://linear.app/uniswap/issue/PROTO-686/event-for-dynamic-fee-update)

I don't know what this is supposed to be, but it leads to a registration that is followed by pretty much nothing. Please stop spamming my report.

saucepoint:
hi thank you for the suggestion but we are unable to accept changes that will change the PoolManager's deployed bytecode

PR #926, opened 2024-12-12
Define and Implement Dynamic Fee Update Event

floor-licker:
## Related Issue

#920 

## Description of changes

This PR declares a new event `DynamicLPFeeUpdated` within the `IPoolManager` contract (found in `IPoolManager.sol`) which is essentially an interface contract to be inherited by the `PoolManager` contract (found in `PoolManager.sol`). The new event is emitted whenever the function `updateDynamicLPFee` in is called, and is called on the last line of the function definition to ensure it is only called in the case that the function has executed properly.

In using 

`grep -rl "is IPoolManager" --include="*.sol" .`

We can see that the only two contracts in the `v4-core` repo which inherit from `IPoolManager` are in `./src/test/ProxyPoolManager.sol` and `./src/test/ProxyPoolManager.sol`.  I’ve implemented the event emissions appropriately in both cases.

### DIFF
--- a/src/PoolManager.sol
+++ b/src/PoolManager.sol
@@ -342,6 +342,9 @@ contract PoolManager is IPoolManager, ProtocolFees, NoDelegateCall, ERC6909Claim
         newDynamicLPFee.validate();
         PoolId id = key.toId();
         _pools[id].setLPFee(newDynamicLPFee);
+
+        // Emit the DynamicLPFeeUpdated event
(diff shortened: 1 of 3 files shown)


floor-licker:
@hensha256 Ok makes sense. Is it safe to say that all the smart contracts in v4-core are in the same boat or just specifically that directory?

kaber2:
This event is essential to get a proper view of the pool state externally. Are you saying you'd rather release it in this very difficult to work with state instead of making this trivial change?

PR #929, opened 2025-01-08
chore(infra): Setup NPM Deploy

mr-uniswap:
Sets up the npm deploy. This needs a package.json file to operate before it will work

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -39,7 +39,7 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly
+          version: stable
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,0 +1,57 @@
+name: Release
+on:
+  # manual trigger
+  workflow_dispatch:
+
+jobs:
+  deploy:
+    name: release
+    runs-on:
+      group: npm-deploy
+    environment:
+      name: release
+    permissions:
+      id-token: write
+      contents: write
+    steps:
+      - name: Load secret
+        uses: 1password/load-secrets-action@581a835fb51b8e7ec56b71cf2ffddd7e68bb25e0
+        with:
+          # Export loaded secrets as environment variables
+          export-env: true
+        env:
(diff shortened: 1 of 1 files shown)


Johnson1s (commented):
[Lint](https://github.com/Uniswap/v4-core/actions/workflows/lint.yml)

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,57 @@
...
+        with:
+          version: nightly
+
+      - name: Install dependencies
+        run: |
+          yarn install --immutable --immutable-cache
+          git submodule update --init --recursive
+
+      - name: Compile
+        run: yarn compile
```
hensha256, line 50: can this just run forge build? or do we need to add a yarn compile?

mr-uniswap: You need to build out the dependencies for the npm package

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,57 @@
...
+        with:
+          version: nightly
+
+      - name: Install dependencies
+        run: |
+          yarn install --immutable --immutable-cache
+          git submodule update --init --recursive
+
+      - name: Compile
+        run: forge install
```
hensha256, line 50: this isnt compiling, `forge build` does compilation

mr-uniswap: Ah thanks

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,57 @@
...
+        uses: actions/setup-node@1a4442cacd436585916779262731d5b162bc6ec7
+        with:
+          cache: yarn
+          node-version: 18
+          registry-url: "https://registry.npmjs.org"
+
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
```
hensha256, line 42: ohh and can you change this to `stable`, and do you mind changing `nightly` to `stable` in all the other CIs?
nightly foundry has caused us some issues recently

gretzke:
@mr-uniswap don't we want to use provenance for this package, too?

mr-uniswap:
> @mr-uniswap don't we want to use provenance for this package, too?

Yes, we do, but that gets set in the package.json file when you all have it ready in your publish config:
```
  "publishConfig": {
    "access": "public",
    "provenance": true
  },
```

PR #934, opened 2025-01-26
feat: add division by zero checks in UnsafeMath library

crStiv:
Add explicit division by zero checks in UnsafeMath library functions to prevent silent failures:
- Add DivisionByZero custom error
- Add checks in divRoundingUp and simpleMulDiv functions
- Update documentation to reflect new behavior
- Improve safety while maintaining gas efficiency with assembly operations

This change makes the library more secure by preventing silent failures on division by zero operations, while still maintaining the performance benefits of using assembly for the actual calculations.


hensha256:
Thanks for your contribution! v4-core is frozen as its already been deployed, so we are not accepting any changes

PR #941, opened 2025-02-06
npm deploy

mr-uniswap:
## Related Issue

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -30,10 +30,13 @@ jobs:
         uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
 
       - name: Setup Node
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -30,10 +30,11 @@ jobs:
         uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
 
       - name: Setup Node
-        uses: actions/setup-node@1a4442cacd436585916779262731d5b162bc6ec7
+        uses: actions/setup-node@v4.2.0
```
gretzke, line 33: why are you using a tag instead of a commit now?

mr-uniswap: the actions/setup-node is a github first party action, so we can trust it, and it never needed to be on a commit hash

gretzke (approved):

Issue #954, opened 2025-02-22
[Bug]: Amount0 and Amount1 not calculated properly in AmountHelpers.sol library


Shamik-Raja:
### Describe the bug

Contract: v4-core/test/utils/AmountHelpers.sol

Function:

function getMaxAmountInForPool(
        IPoolManager manager,
        IPoolManager.ModifyLiquidityParams memory params,
        PoolKey memory key
    ) public view returns (uint256 amount0, uint256 amount1) {
        PoolId id = key.toId();
        uint128 liquidity = StateLibrary.getLiquidity(manager, id);
        (uint160 sqrtPriceX96,,,) = StateLibrary.getSlot0(manager, id);

        uint160 sqrtPriceX96Lower = TickMath.getSqrtPriceAtTick(params.tickLower);
        uint160 sqrtPriceX96Upper = TickMath.getSqrtPriceAtTick(params.tickUpper);

        amount0 = LiquidityAmounts.getAmount0ForLiquidity(sqrtPriceX96Lower, sqrtPriceX96, liquidity);
        amount1 = LiquidityAmounts.getAmount1ForLiquidity(sqrtPriceX96Upper, sqrtPriceX96, liquidity);
    }

### Expected Behavior

The last 2 lines to calculate amount0 and amount1 should be 

amount0 = LiquidityAmounts.getAmount0ForLiquidity(sqrtPriceX96, sqrtPriceX96Upper, liquidity);
amount1 = LiquidityAmounts.getAmount1ForLiquidity(sqrtPriceX96Lower, sqrtPriceX96, liquidity);

PR #955, opened 2025-02-26
Fix amount0 and amount1 calculation with improved price range checks

nigeldelviero:
## Related Issue

#954 

## Description of changes
1. Fixed Calculation Logic in getMaxAmountInForPool:
- Ensured correct order of sqrtPriceX96 values by adding conditional checks.
- Added validation to verify that the current price is within the tick range.
- Improved handling of edge cases such as price being exactly at boundaries.
2. Enhanced Error Handling:
- Added require statements to check if the price is within the acceptable range.
- Proper error messages for out-of-range prices.

### DIFF
--- a/test/utils/AmountHelpers.sol
+++ b/test/utils/AmountHelpers.sol
@@ -18,12 +18,23 @@ library AmountHelpers {
     ) public view returns (uint256 amount0, uint256 amount1) {
         PoolId id = key.toId();
         uint128 liquidity = StateLibrary.getLiquidity(manager, id);
-        (uint160 sqrtPriceX96,,,) = StateLibrary.getSlot0(manager, id);
+        (uint160 sqrtPriceX96, , , ) = StateLibrary.getSlot0(manager, id);
 
         uint160 sqrtPriceX96Lower = TickMath.getSqrtPriceAtTick(params.tickLower);
         uint160 sqrtPriceX96Upper = TickMath.getSqrtPriceAtTick(params.tickUpper);
 
-        amount0 = LiquidityAmounts.getAmount0ForLiquidity(sqrtPriceX96Lower, sqrtPriceX96, liquidity);
-        amount1 = LiquidityAmounts.getAmount1ForLiquidity(sqrtPriceX96Upper, sqrtPriceX96, liquidity);
+        require(sqrtPriceX96 >= sqrtPriceX96Lower && sqrtPriceX96 <= sqrtPriceX96Upper, "Price out of range");
+
+        if (sqrtPriceX96 < sqrtPriceX96Lower) {
+            amount0 = LiquidityAmounts.getAmount0ForLiquidity(sqrtPriceX96, sqrtPriceX96Lower, liquidity);
+        } else {
+            amount0 = LiquidityAmounts.getAmount0ForLiquidity(sqrtPriceX96Lower, sqrtPriceX96, liquidity);
+        }
+
+        if (sqrtPriceX96 > sqrtPriceX96Upper) {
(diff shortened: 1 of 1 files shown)


test/utils/AmountHelpers.sol
```diff
@@ -18,12 +18,23 @@ library AmountHelpers {
...
         uint128 liquidity = StateLibrary.getLiquidity(manager, id);
-        (uint160 sqrtPriceX96,,,) = StateLibrary.getSlot0(manager, id);
+        (uint160 sqrtPriceX96, , , ) = StateLibrary.getSlot0(manager, id);
 
         uint160 sqrtPriceX96Lower = TickMath.getSqrtPriceAtTick(params.tickLower);
         uint160 sqrtPriceX96Upper = TickMath.getSqrtPriceAtTick(params.tickUpper);
 
-        amount0 = LiquidityAmounts.getAmount0ForLiquidity(sqrtPriceX96Lower, sqrtPriceX96, liquidity);
-        amount1 = LiquidityAmounts.getAmount1ForLiquidity(sqrtPriceX96Upper, sqrtPriceX96, liquidity);
+        require(sqrtPriceX96 >= sqrtPriceX96Lower && sqrtPriceX96 <= sqrtPriceX96Upper, "Price out of range");
```
louisbrown0212, line 26: It might be better for readability
```suggestion
        require(sqrtPriceX96Lower <= sqrtPriceX96 && sqrtPriceX96 <= sqrtPriceX96Upper, "Price out of range");
```

test/utils/AmountHelpers.sol
```diff
@@ -18,12 +18,23 @@ library AmountHelpers {
...
+
+        if (sqrtPriceX96 < sqrtPriceX96Lower) {
```
louisbrown0212, line 28: Is it possible that `sqrtPriceX96` is below than `sqrtPriceX96Lower` because of the above `require` statement?
If `sqrtPriceX96 < sqrtPriceX96Lower` is true, it will revert with `Price out of range`

test/utils/AmountHelpers.sol
nigeldelviero: Thanks for the suggestion! I agree that the suggested order improves readability. I've updated the require statement accordingly.

test/utils/AmountHelpers.sol
nigeldelviero: Good catch! Since the require statement already guarantees that sqrtPriceX96 >= sqrtPriceX96Lower, this condition was redundant. I've removed it to simplify the function. Thanks for pointing it out!

test/utils/AmountHelpers.sol
```diff
@@ -18,12 +18,23 @@ library AmountHelpers {
...
+            amount0 = LiquidityAmounts.getAmount0ForLiquidity(sqrtPriceX96, sqrtPriceX96Lower, liquidity);
+        } else {
+            amount0 = LiquidityAmounts.getAmount0ForLiquidity(sqrtPriceX96Lower, sqrtPriceX96, liquidity);
+        }
+
+        if (sqrtPriceX96 > sqrtPriceX96Upper) {
```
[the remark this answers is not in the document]
nigeldelviero: Right!  I've removed it to simplify the function.

test/utils/AmountHelpers.sol
```diff
@@ -18,12 +18,23 @@ library AmountHelpers {
     ) public view returns (uint256 amount0, uint256 amount1) {
```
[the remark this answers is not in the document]
nigeldelviero: Thanks for the suggestion! Since this function is not used internally, I've updated it to `external` to optimize gas usage.

PR #959, opened 2025-03-14
fix: StateLibrary change @param

dennnis0204:
## Related Issue
Trying to get position liquidity like in guide https://docs.uniswap.org/contracts/v4/guides/read-pool-state#getpositioninfo
Function getPositionInfo(
        IPoolManager manager,
        PoolId poolId,
        address owner,
        int24 tickLower,
        int24 tickUpper,
        bytes32 salt
    ) doesn't work with @param owner The owner of the liquidity position.
Works with @param address of the position manager.

In v4-periphery -> PositionManager.sol  -> 
bytes32 positionId = Position.calculatePositionKey(address(this), tickLower, tickUpper, bytes32(tokenId));
The same calculatePositionKey with address(this) == address(positionManager)

## Description of changes
Changed in StateLibrary.sol function getPositionInfo @param owner The owner of the liquidity position. to @param posm The position manager.

### DIFF
--- a/src/libraries/StateLibrary.sol
+++ b/src/libraries/StateLibrary.sol
@@ -219,7 +219,7 @@ library StateLibrary {
      * @notice Retrieves the position information of a pool without needing to calculate the `positionId`.
      * @dev Corresponds to pools[poolId].positions[positionId]
      * @param poolId The ID of the pool.
-     * @param owner The owner of the liquidity position.
+     * @param posm The position manager.
      * @param tickLower The lower tick of the liquidity range.
      * @param tickUpper The upper tick of the liquidity range.
      * @param salt The bytes32 randomness to further distinguish position state.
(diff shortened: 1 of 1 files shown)


src/libraries/StateLibrary.sol
```diff
@@ -219,7 +219,7 @@ library StateLibrary {
      * @notice Retrieves the position information of a pool without needing to calculate the `positionId`.
      * @dev Corresponds to pools[poolId].positions[positionId]
      * @param poolId The ID of the pool.
-     * @param owner The owner of the liquidity position.
+     * @param posm The position manager.
```
saucepoint, line 222:
```suggestion
     * @param owner The owner contract of the liquidity position. Typically a periphery contract like PositionManager
```

saucepoint (commented):
thanks for the feedback, I can see how its confusing

I would actually suggest reverting back to the `owner` naming, and just improving the natspec. Not all liquidity managers will conventionally follow a PosM-style contract

For example, a hook can manage liquidity or developers can write a contract which manages a single position; therefore, I think its misleading to call the param `posm`

src/libraries/StateLibrary.sol
dveth2: reverting back to the `owner` naming, and just improving the natspec is good enough!

PR #961, opened 2025-03-26
Safer readme example

kadenzipfel:
## Related Issue

N/A

## Description of changes

Ensuring that `unlockCallback` is only called by the `poolManager` is almost always a critically important security feature, so imo it should be included even in the most basic of examples


saucepoint (approved):
but this works against ur own job security ??

PR #964, opened 2025-04-01
fix(deps): resolve circular dependency that breaks clean builds

ccashwell:
## Related Issue

This resolves an issue that causes builds to fail in certain conditions.

## Description of changes

- Created `types/PoolOperation.sol` consisting of the structs `ModifyLiquidityParams` and `SwapParams` which are removed from `interfaces/IPoolManager.sol`
- Updated references to point to the new location of these structs.

### DIFF
--- a/src/types/PoolOperation.sol
+++ b/src/types/PoolOperation.sol
@@ -1,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import {PoolKey} from "../types/PoolKey.sol";
+import {BalanceDelta} from "../types/BalanceDelta.sol";
+
+/// @notice Common types used by both IPoolManager and IHooks
+interface PoolOperation {
+    struct ModifyLiquidityParams {
+        // the lower and upper tick of the position
+        int24 tickLower;
+        int24 tickUpper;
+        // how to modify the liquidity
+        int256 liquidityDelta;
+        // a value to set if you want unique liquidity positions at the same range
+        bytes32 salt;
+    }
+
+    struct SwapParams {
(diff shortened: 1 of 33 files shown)


src/types/PoolOperation.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import {PoolKey} from "../types/PoolKey.sol";
+import {BalanceDelta} from "../types/BalanceDelta.sol";
+
+/// @notice Common types used by both IPoolManager and IHooks
+interface PoolOperation {
```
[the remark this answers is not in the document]
ccashwell: Yeah they could be top level rather than wrapped in an interface. Will make that change.

src/types/PoolOperation.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import {PoolKey} from "../types/PoolKey.sol";
+import {BalanceDelta} from "../types/BalanceDelta.sol";
+
+/// @notice Parameter struct for `ModifyLiquidity` pool operations
+struct ModifyLiquidityParams {
```
gretzke, line 8: I see you removed the interface, can we confirm that this still solves the issue? When I implemented this without the interface, it didn't fix the circular dependency

ccashwell: Yes, this does still solve the issue. Once these structs are moved out of IPoolManager, the circular references between IHooks and IPoolManager are eliminated and the monorepo builds cleanly.

dianakocsis (approved):

PR #969, opened 2025-04-21
Add certora verification specs

ccashwell:
## Description of changes

This adds Certora's formal verification specs to the v4-core repo with no changes to any Uniswap code or the build pipeline.


snreynolds (commented):
Wonder if this should live in tests/ actually?

Happy to merge though!

ccashwell:
@snreynolds I feel like having it in `tests/` is wrong because it's not going to run without external tooling (which isn't documented anywhere in the repo and seems like a side quest), plus much of the Solidity code included here is either mocks or simplifications of existing logic which imo would increase confusion if analyzed within the same context as the actual test suite. My view is that the FV files should be segregated from the main code and tests for clarity if nothing else.

Ericak31:
Amazing

PR #989, opened 2025-10-22
chore(ci): Add trusted publishing

dgilmanuni:
## Description of changes

- Added Trusted Publishing
- Removed Semgrep workflow file
- Added and updated action version pins
- Added bullfrog
- Update npm during release


zhongeric:
We noticed that in forge `1.4.X`, `forge fmt` has a whitespace diff when applied on main, causing the `lint` action to fail, so we decided to pin the version to 1.3.6.

zhongeric (approved):

PR #1015, opened 2026-02-24
fix: correct getLiquidity docstring to reflect active liquidity

### LINKED ISSUE #984: [Bug]: `StateLibrary.getLiquidity` docstring is incorrect
### Describe the bug

Code is here: 
https://github.com/Uniswap/v4-core/blob/59d3ecf53afa9264a16bba0e38f4c5d2231f80bc/src/libraries/StateLibrary.sol#L176-L191

The docstring states that the function "[r]etrieves total the liquidity of a pool" but an examination of the code reveals that it instead gets the active liquidity, not total liquidity. Active liquidity is fine but the doscstring is misleading.

### Expected Behavior

Expected behavior is that the function would return the total liquidity.

### To Reproduce

It's a lot of overhead to show it in code (you can just look at what the source does) but it can be confirmed by doing the following:

1. Make a new, empty pool with some starting price with tick `t`.
2. Add _single-sided liquidity_ at tick `t'` where `t != t'`.
3. Check that tokens were moved into the pool manager via `getBalance`.
4. Call `StateLibrary.getLiquidity()`. It will return 0 since the current price/tick is not in the range of liquidity added.

Obviously one must take are that `t'` is not in the same tick-spacing-bin as `t`.

### Additional context

_No response_

ionodeionode:
## Summary
Closes #984

The `getLiquidity` function's natspec docstring stated it "retrieves total the liquidity" but the function actually returns the **active liquidity** at the pool's current tick (`pools[poolId].liquidity`), not the total liquidity across all positions.

Single-sided liquidity positions outside the current tick range are not included in this value. Updated the docstring to accurately describe this behavior and avoid confusion.

### DIFF
--- a/src/libraries/StateLibrary.sol
+++ b/src/libraries/StateLibrary.sol
@@ -174,11 +174,12 @@ library StateLibrary {
     }
 
     /**
-     * @notice Retrieves total the liquidity of a pool.
-     * @dev Corresponds to pools[poolId].liquidity
+     * @notice Retrieves the active liquidity of a pool (i.e. liquidity in range of the current tick).
+     * @dev Corresponds to pools[poolId].liquidity. This is NOT the total liquidity across all ticks;
+     * it only reflects liquidity that is currently active at the pool's current tick.
      * @param manager The pool manager contract.
      * @param poolId The ID of the pool.
-     * @return liquidity The liquidity of the pool.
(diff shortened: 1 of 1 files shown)


ionodeionode:
Closing to manage open PR count. Will revisit when capacity allows.

PR #1068, opened 2026-08-18
fix: add zero-address check to ERC6909 mint

angguntrie3-lgtm:
## Summary

Adds a zero-address check to `ERC6909._mint()` to prevent tokens from being minted to the zero address.

## Problem

The current `_mint` function does not check if the receiver is the zero address:

```solidity
function _mint(address receiver, uint256 id, uint256 amount) internal virtual {
    // No check for receiver == address(0)
    balanceOf[receiver][id] += amount;
}
```

If called with `address(0)`, tokens are effectively burned (balance increases at address(0) which is unrecoverable).

## Fix

```solidity
function _mint(address receiver, uint256 id, uint256 amount) internal virtual {
    require(receiver != address(0), "ERC6909: mint to zero address");
    balanceOf[receiver][id] += amount;
}
```

## Impact

- Prevents accidental token loss
- Protects users who mistakenly call `PoolManager.mint()` with address(0)
- No functional changes to existing code

## Testing

- Existing test suite should pass
- Add test for zero-address revert scenario

---

*Found during security audit by Wulansari Security Research.*

### DIFF
--- a/src/ERC6909.sol
+++ b/src/ERC6909.sol
@@ -77,6 +77,8 @@ abstract contract ERC6909 is IERC6909Claims {
     //////////////////////////////////////////////////////////////*/
 
     function _mint(address receiver, uint256 id, uint256 amount) internal virtual {
+        require(receiver != address(0), "ERC6909: mint to zero address");
+
         balanceOf[receiver][id] += amount;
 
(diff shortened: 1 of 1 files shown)


zhongeric:
Hey thanks for the suggestion, however we're not intending to implement this.

